PR #8, opened 2019-07-23
Enable Mode and Right Align option for GUI DIsp Float

robbycandra:
Enable Mode and Right Align option for GUI DIsp Float


robbycandra:
@bigtreetech , you merged it so fast? Have you test it ?

PR #17, opened 2019-08-03
User-Definable "LCD12864 Simulator" Text

thisiskeithb:
"LCD12864 Simulator" is now definable in `Configuration.h`.

### DIFF
--- a/TFT/src/User/API/UI/ST7920_Simulator.c
+++ b/TFT/src/User/API/UI/ST7920_Simulator.c
@@ -1,8 +1,13 @@
 #include "ST7920_Simulator.h"
 #include "includes.h"
 #include "GUI.h"
+#include "../../Configuration.h"
 
 #ifdef ST7920_SPI
+
+#ifndef ST7920_BANNER_TEXT
+  #define ST7920_BANNER_TEXT "LCD12864 Simulator"
+#endif
 
 ST7920_PIXEL       st7920 = {ST7920_XSTART, ST7920_YSTART, 0};
 ST7920_CTRL_STATUS status = ST7920_IDLE;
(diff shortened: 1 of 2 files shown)


Sthopeless:
Hello!
This is great! 
But my suggestion is, wouldn't it be better to have something like #define printer_name "" in Configuration.h file and use that for the text? 
Then if someone wants they can write whatever they like and have it displaying or leave it empty and have nothing

thisiskeithb:
By default, you'd set the printer name in [Marlin's `Configuration.h`](https://github.com/MarlinFirmware/Marlin/blob/52a7a8cb265d89c20e352bfaf8053e0e72fc1d0f/Marlin/Configuration.h#L138) which shows up in the status line, but in regards to this firmware, Many defines can be added to `Configuration.h`.

In fact, I'm adding 12864 background & font color defines to `Configuration.h` in my fork right now.

Sthopeless:
Sorry I didn't explain myself correctly. But look how I made the changes (I don't have a dual firmware TFT yet to test it but it compiled ok) 
1#:
https://github.com/Sthopeless/BIGTREETECH-TouchScreenFirmware/blob/d281756c4d10158932f93211aca5e572265d7784/TFT/src/User/API/UI/ST7920_Simulator.c#L129
  
2#:
https://github.com/Sthopeless/BIGTREETECH-TouchScreenFirmware/blob/d281756c4d10158932f93211aca5e572265d7784/TFT/src/User/Configuration.h#L4-L5

This way everyone could customise the msg to they're own easily.

thisiskeithb:
@Sthopeless You explained yourself fine. I updated the PR so the text is now user-definable instead of hidden.

PR #89, opened 2019-11-02
update character encode from GBK to UTF-8

Msq001:
1. update character encode from GBK to UTF-8
2. update Fatfs version from 0.12b to 0.14 for UTF-8 encode
3. improve filament runout, filter interference
4. add smart filament runout detect


thisiskeithb:
Do you have any advice on what we have to change in order to get fonts like Ubuntu (a great sans-serif font) from [this repo](https://github.com/Acenotass/BIGTREETECH-TFT35-V1.2/tree/master/firmware/Copy%20to%20SD%20Card%20root%20directory%20to%20update/TFT35/font) working with this change? It was working fine up until this PR/commit, but now incorrect letters are displayed.

SG Bitmap Font Editor -> Export screenshot:

Screen Shot 2019-11-02 at 1 07 30 AM

Ubuntu font after this change:

IMG_4397

Ubuntu font before this change (and with custom icons):
IMG_4394

bigtreetech:
hello， the ascii bitmap font should start from ‘ ’(0x20 character "space") to '~'(0x7E) now, before that it's  start from 0x00(null) to 0x7F(delete), I deleted some invisible ASCII codes and only kept the visible ones. 
now you have 2 ways:
 1. regenerate a font start from 0x20 to 0x7E and update it to TouchScreen
 2. modify firmware in here https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/381c463d2826f0bee266c632b5ff5e4b075479d1/TFT/src/User/API/Language/utf8_decode.c#L8. change '0x20' to 0,  and here https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/381c463d2826f0bee266c632b5ff5e4b075479d1/TFT/src/User/API/UI/GUI.c#L476 `uint32_t addrOffset = (ch - 0x20) * bitMapSize;` to `uint32_t addrOffset = (ch - 0x0) * bitMapSize;`
So far, I haven't found a suitable dot matrix font generator, so it's not sure about the parsing of some codes. Do you think it's necessary that to restore ASCII decode method to before，as the 2 way above，would it be more convenient to customize fonts in this way?

PS: Thank you very much for helping to maintain this project!  I'm still doing more improvements and clean up, when I finish this, I will write some documents to show how to add languages, customize fonts, customize icons, etc... And thank for again!

thisiskeithb:
> Do you think it's necessary that to restore ASCII decode method to before，as the 2 way above，would it be more convenient to customize fonts in this way?

I don't know since this is the first time I've tried modifying this type of firmware/fonts.

[This thread](https://github.com/bigtreetech/BIGTREETECH-TFT35-V1.2/issues/41) has info on creating new fonts as well as templates used, so it could be a starting point. I just don't know how to edit the fonts or export them to work with these new changes.

Issue #90, opened 2019-11-02
Latest version does not compile for TFT24


davehardy20:
I saw the changes to the report this morning and 'pulled' the changes and tried to compile for TFT24 v1.1, but got these errors

> Executing task in folder BIGTREETECH-TouchScreenFirmware: C:\Users\dave\.platformio\penv\Scripts\platformio.exe run <

Processing BIGTREE_TFT24_V1_1 (platform: ststm32; framework: stm32cube; board: STM32F105RC)
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Verbose mode can be enabled via `-v, --verbose` option
CONFIGURATION: https://docs.platformio.org/page/boards/ststm32/STM32F105RC.html     
PLATFORM: ST STM32 5.7.0 > STM32F105RC (48k RAM. 256k Flash)
HARDWARE: STM32F105RCT6 72MHz, 48KB RAM, 256KB Flash
DEBUG: Current (blackmagic) External (blackmagic, cmsis-dap, jlink, stlink)
PACKAGES: toolchain-gccarmnoneeabi 1.70201.0 (7.2.1), framework-stm32cube 2.0.181130
LDF: Library Dependency Finder -> http://bit.ly/configure-pio-ldf
LDF Modes: Finder ~ chain, Compatibility ~ soft
Found 11 compatible libraries
Scanning dependencies...
Dependency Graph
|-- <BSP-Common>
Compiling .pio\build\BIGTREE_TFT24_V1_1\src\src\User\API\Gcode\gcode.o
Compiling .pio\build\BIGTREE_TFT24_V1_1\src\src\User\API\Gcode\mygcodefs.o
Compiling .pio\build\BIGTREE_TFT24_V1_1\src\src\User\API\LCD_Encoder.o
Compiling .pio\build\BIGTREE_TFT24_V1_1\src\src\User\API\Language\Language.o
Compiling .pio\build\BIGTREE_TFT24_V1_1\src\src\User\API\Language\utf8_decode.o
Compiling .pio\build\BIGTREE_TFT24_V1_1\src\src\User\API\UI\GUI.o
Compiling .pio\build\BIGTREE_TFT24_V1_1\src\src\User\API\UI\ST7920_Simulator.o
Compiling .pio\build\BIGTREE_TFT24_V1_1\src\src\User\API\UI\touch_process.o
[...]

PR #92, opened 2019-11-02
fix for TFT35 V3

UnknownHero99:
fix for TFT35 V3, TFT24 V1.1 still unresolved


thisiskeithb:
Thanks for this fix! I'm able to compile for the TFT35 V3 now 🙂

UnknownHero99:
now completely fixed #90 issue with @spilz87 's solution included

PR #95, opened 2019-11-03
added preheat menu, temperatures are set in configuration file

UnknownHero99:
preheat menu added


guruathwal:
@UnknownHero99  In the preheat menu we have icons for ABS, PETG, PLA which already show the name for each of them.
the label can be set to show the preset temperature setting like: "T230 B100" for ABS 👍

PR #108, opened 2019-11-05
Fix Stretched 12864 Fullscreen Mode

thisiskeithb:
### Description

This PR fixes the stretched 12864 fullscreen mode and restores the ability for custom `ST7920_BANNER_TEXT` text again since the two options no longer overlap on screen.

### Benefits

Emulated pixels are now square!

### Other

Tested on a TFT35 V3.

Before PR:
Before PR

After PR:
After PR

After with `ST7920_BANNER_TEXT` enabled:
j4tuV0B - Imgur


extesy:
When the banner is enabled, shouldn't you have an empty line to separate the banner text and the main screen? You have unused empty screen space at the bottom.

thisiskeithb:
> When the banner is enabled, shouldn't you have an empty line to separate the banner text and the main screen?

Shifting the entire emulated mode down could be possible.

I wish the banner would just be removed, but some users like it. 🤷‍♂️

thisiskeithb:
Here's a version with the emulated section pushed down further:

IMG_4493

PR #116, opened 2019-11-06
Update language_de.h

arminth:
Updated German Language file. Added Colors, implemented correct German "Umlaute" (äöüß), removed 2 duplicate keys  and changed some of the media-keys (USB-Stick etc.)
Compiled it for my 2.4". Works.


thisiskeithb:
Do the letters with umlauts show up on the LCD?

arminth:
Yes indeed. It works flawless.

PR #129, opened 2019-11-08
fix extrusion in printing bugs in relative mode.

Msq001:


Msq001:
> For Syncing the fork from the Upstream Branch try this:
> 
> 1. open Git Bash in root folder of the fork on your computer.
> 2. run following commands
> 
> ```
> git checkout master
> git fetch upstream
> git reset --hard upstream/master
> git push --force
> ```
> 
> this way the PR for Sync will not get pushed to Upstream again.
> Earlier I did same as you which resulted in too much difference in commits.

thank you very much! It really helped me a lot

Issue #126, opened 2019-11-07
TFT35 V2 latest version not updating


UnknownHero99:
when trying to update to the latest version it just stucks on updating screen at 100%, prebuilt version works just fine, so possibility that problem is on my side

guruathwal:
What changes have you made to your code?

thisiskeithb:
I'm experiencing a similar issue with a TFT35 V3 with self-compiled firmware after the changes just implemented. Precompiled [from a couple days ago works fine](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/2829e446ec5cccd5fe0679ec6e17f7a1885cb4de/Copy%20to%20SD%20Card%20root%20directory%20to%20update/BIGTREE_TFT35_V3.0.24.2.bin).

guruathwal:
Try pressing reset button after of it reaches 100.
If it boots check the build number and date to confirm if the firmware updated or not.

thisiskeithb:
The binary is loaded on reset, but the LCD stays blank after the initial load like it’s dead. And resetting (without the SD card) does nothing.

Loading the older binary I linked gets it working again.

I’ll see if downgrading PIO fixes it since recent commits shouldn’t have caused this issue.

UnknownHero99:
@guruathwal  no changes at all the problem is exactly the same as @thisiskeithb is saying, removing sd card and pressing reset button only shows a black screen, prebuilt works just fine

@thisiskeithb this surely could be due to PIO update didn't even notice it updated

guruathwal:
> @guruathwal no changes at all the problem is exactly the same as @thisiskeithb is saying, removing sd card and pressing reset button only shows a black screen, prebuilt works just fine
> 
> @thisiskeithb this surely could be due to PIO update didn't even notice it updated

the current PIO version on my PC is : **Home 2.3.3·Core 4.0.3** running on VSCodium **1.36.1**, with which the firmware update is working fine.
what is your PIO version?

thisiskeithb:
4.1 was just released, so that’s what I’m running. I can’t downgrade right now, but will be able to test this weekend.

UnknownHero99:
I can confirm PIO is causing it. I just downgraded from 4.1 -> 4.0.3 and it's working fine now.
looks like there is a bug in 4.1 PIO Core so until they fix it don't upgrade to 4.1, if u already did 
[...]

PR #136, opened 2019-11-09
Override default LD Script using LINKFLAGS build scope

ivankravets:
Resolves https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/126

This is a temporary low-level solution that works with previous PIO Core versions and will work with the future. However, there is a better way to do this via `board_build.ldscript`. There is a bug in ST STM32 dev/platform which overrides user custom LD Script declared in `board_build.ldscript` way. We will fix it soon. So, you can simplify your configuration later.

Sorry for the issue.

### DIFF
--- a/buildroot/boards/STM32F103VC.json
+++ b/buildroot/boards/STM32F103VC.json
@@ -14,7 +14,6 @@
         "0x0004"
       ]
     ],
-    "ldscript": "BIGTREE_TFT35_V1.ld",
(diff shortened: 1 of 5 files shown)


thisiskeithb:
@bigtreetech This is confirmed working on at TFT35 V3.

bigtreetech:
thank you very much!

PR #145, opened 2019-11-10
Use Ubuntu Font for TFT35

thisiskeithb:
### Description

Ubuntu is now the default font for TFT35. Updates to the other models are coming. Thanks to work by @Acenotass & @guruathwal!

**Before/Times font:**
image

**After/Ubuntu font:**
image

### Benefits

Modernizes the default theme with a sans serif font.


Sthopeless:
@thisiskeithb I don't know anything about fonts or however it works(!), but awhile translating the TFT to Portuguese I noticed the fonts were missing 'special characters' such as ```ã, í, ê, á``` which we use a lot in my language.
They're also missing on this font, is there something can be done to add them? 
here's one example on  ```Início```
IMG_20191110_183054_HDR

thisiskeithb:
The characters would need to be added to the font file/mapped out. I'm new to these bitmap fonts, so I haven't figured out how to really edit these types of files efficiently yet.

PR #179, opened 2019-11-23
New feature Listview File browser with icons embeded in fonts....Update compiled binaries wiht latest updates

guruathwal:
👍 Added new feature: list-view file browser with icons embedded in fonts with possibility to change color of icons in source.

❗ Font files also need to be updated.

⚠Tested only on TFT35 V2. needs testing on TFT24 / TFT28.

Fixed missed information while status screen is not visible. 

Update compiled binaries with latest features and added old version archive.

photo_2019-11-24_23-04-02


extesy:
Why use the same version number? It's a different binary.

guruathwal:
@extesy made necessary correction.. 👍

PR #199, opened 2019-11-27
Fixed bugs in status screen, Added option to Force-Reset touch screen calibration, show X Y Z position in move menu

guruathwal:
- Fixed bug where the status screen does not update the X Y & Z axis position.
- replaced ugly icon in info-box.
- Fixed bug where info-box shows when sending 'M114" command in G-code menu.
- Added display of X Y Z position in move menu
- Option to invert Y Axis buttons.
- Added code to check for `reset` directory on boot to reset touch screen calibration and reset stored settings.
- To reset the screen you need to make a folder/directory in root with named `reset`
Touch Screen calibration will run after reset.


extesy:
> Added code to check for reset directory on boot to reset touch screen calibration and reset stored settings.

This is a great idea, but why use a directory and not a file? Usually file-based resets like this are done with files, for example on Kindle devices.

guruathwal:
@extesy nothing special about using a directory, I used it just because its quicker to create a directory.

thisiskeithb:
+1 for using a file instead of a folder since that’s more of a standard.

guruathwal:
@thisiskeithb  done 👍

guruathwal:
@bigtreetech please reduce the size of the screen photos of the display in readme.md
From this 
Annotation 2019-11-28 122900

to this:
Annotation 2019-11-28 122945

thisiskeithb:
@guruathwal: It's pretty easy to change since you have an open PR:

## Change this:
`image`
## To This:
``
## Result:

guruathwal:
@thisiskeithb not near a PC right now... will do it later..

PR #224, opened 2019-12-06
add long filename display when browsing board SD

melufre:
### Requirements

#define GCODE_LIST_MODE shall be define in configuration.h

### Description
 on gcode.c
- add M33 command
- change position of clearRequestCommandInfo(); because the free was done before the buffer hasbeen copied
-  add WaitingGcodeResponse in order to avoid colision during serial frame reception with periodic temperature report (used in extend.c and heat.c)
- mygcodefs.c do M20 and copy the buffer and after it perform M33 on each files

### Benefits
display long file name when you browse on board SD

### Related Issues
 folder name are still display with short name


melufre:
duplicate of https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/235

TFT/src/User/Menu/Heat.c
```diff
@@ -276,7 +276,7 @@ void loopCheckHeater(void)
   {  /* Send M105 query temperature continuously	*/
     if(update_waiting == true)                {nowTime=OS_GetTime();break;}
     if(OS_GetTime()<nowTime+update_time)       break;
-
+    if(RequestCommandInfoIsRunning())          break; //to avoid colision in Gcode response processing
```
guruathwal: This looks same as `if(update_waiting == true)` in `line no: 277` which is also used to avoid sending gcode commands when there are already some commands waiting for response. Please check and make sure no duplicate codes exists.

melufre: For my understanding update_waiting is a flag used to know if we are waiting for the next occurrence of temperature update  ,it is an  internal management flag  so  not duplicate with my  condition that will delay the update until we have finish with  to retrieve file list or long filename

PR #235, opened 2019-12-11
Longfilename support from onboard SD patch

melufre:
### Requirements

#define GCODE_LIST_MODE shall be define in configuration.h

### Description
 on gcode.c
- add M33 command
- change position of clearRequestCommandInfo(); because the free was done before the buffer hasbeen copied
-  add WaitingGcodeResponse in order to avoid colision during serial frame reception with periodic temperature report (used in extend.c and heat.c)
- mygcodefs.c do M20 and copy the buffer and after it perform M33 on each files

### Benefits
display long file name when you browse on board SD

### Related Issues
 folder name are still display with short name


tatusah:
Fetched and compiled this with my local branch works great. Also fetched the https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/245 to fix the up/down icons. Hopefully these get merged to the master soon so others enjoy these also. Great work!

PR #272, opened 2019-12-31
Reduce the number of flash writes and naming conventions

chzj333:
Reduce the number of flash writes and naming conventions

### DIFF
--- a/TFT/src/User/API/boot.h
+++ b/TFT/src/User/API/boot.h
@@ -143,10 +143,11 @@ enum
   ICON_LEDCOLOR,
   
 //  ICON_RESERVE
-// Preview should be in the last place
-  ICON_PREVIEW,
 // Back ground sign
   ICON_BACKGROUND,
+
+  // Preview should be in the last place
+  ICON_PREVIEW,
 };
 
 typedef union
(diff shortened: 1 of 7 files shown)


Hukuma1:
Needs TFT/src/User/Menu/menu.c from PR: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/263/files to compile:

`void menuDrawIconOnly(const ITEM *item, uint8_t positon)`
`{`
  `uint8_t *content = labelGetAddress(&item->label);`
`  const GUI_RECT *rect = rect_of_key + positon;`
`  if(item->icon != ICON_BACKGROUND)`
`    ICON_ReadDisplay(rect->x0, rect->y0, item->icon);`
`  else
 `   GUI_ClearPrect(rect);`
`}`

TFT/src/User/API/boot.h
```diff
@@ -143,10 +143,11 @@ enum
   ICON_LEDCOLOR,
   
 //  ICON_RESERVE
-// Preview should be in the last place
-  ICON_PREVIEW,
 // Back ground sign
   ICON_BACKGROUND,
+
+  // Preview should be in the last place
+  ICON_PREVIEW,
```
guruathwal: putting ICON_PREVIEW after ICON_BACKGROUND is just wastage of flash storage as the ICON_BACKGROUND is not used to write anything on the flash storage.
placing the ICON_PREVIEW after ICON_BACKGROUND just skips 1 icon count wasting the flash space at address of ICON_BACKGROUND.

PR #277, opened 2020-01-02
fixed invert axis implementation

UnknownHero99:
### Benefits
fixes invert axis implementation

### Related Issues

previously Z up button had Z down label now that is fixed, 
labels now stay the same even after inverting which is the true purpose of inverting axis
for example now if inverted Z up button will have text Z up but will go down (different types of printers)

EDIT: fixes default Z move button which was switched before (up button and down button were switched by default)

### DIFF
--- a/TFT/src/User/Menu/Move.c
+++ b/TFT/src/User/Menu/Move.c
@@ -92,24 +92,6 @@ static u32 update_time = 50; // 1 seconds is 100
 void menuMove(void)
 {
   KEY_VALUES  key_num = KEY_IDLE;
-  #ifdef MENU_LIST_MODE
-  if(infoSettings.invert_yaxis == 1){
-    moveItems.items[1].label.index = LABEL_Y_DEC;
-    moveItems.items[5].label.index = LABEL_Y_INC;
-  }
-  else{
-    moveItems.items[1].label.index = LABEL_Y_INC;
-    moveItems.items[5].label.index = LABEL_Y_DEC;
-  }
-    if(infoSettings.invert_zaxis == 1){
-    moveItems.items[0].label.index = LABEL_Z_DEC;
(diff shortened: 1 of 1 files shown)


AnHardt:
Z-axis is always upwards. Increasing Z always increases when the distance between the bed and the nozzle increases. That's always the same - regardless from if the nozzle/gantry rises or the bed is driven down!

UnknownHero99:
that is correct that's why I'm not entirely sure of purpose of this implementation in case it's just meant to switch icons (some might think of it as bed goin down so z increasing) it would be easier to switch icons rather than changing button's purpose but in any case it should at least say Z - at icon named Dec_Z and vice versa at least by default

UnknownHero99:
difference with this change is that by default this button
Dec_Z
says Z- and decreases Z 
after inverting it still says Z- and increases Z (because you inverted it)

guruathwal:
the invert Y axis option was added to let users match the buttons direction with actual Y axis for printers like prusa i3 where the bed moves when moving Y axis while in CoreXY the Extruder moves.

same goes for the invert Z axis option which was added to let users match the button direction with actual Z axis movements as some printers have the bed moving updown while some has the X gantry move updown. there was no goal to invert the actual g-code.

UnknownHero99:
ok i understand then
i'll correct it so by default it's as i said 
this button
Dec_Z
says Z- and decreases Z 
and revert other changes

currently by default it says Z+ and img name is Dec_Z and most printers u know... also wouldnt it be cleaner etc in case user selects invert ony icons would swap now icons stays in same spot but labels and what buttons do change

ipatch:
@UnknownHero99 

totally agree, the labels are updated when inverting is enabled, but the labels do not correspond to the icons their below.

Issue #275, opened 2020-01-01
[BUG] Does not compile with MENU_LIST_MODE enabled TFT35 v3


zech:
# Description

I'm trying to compile the latest firmware from the master branch myself with VSCode/PIO. As soon as I enable list mode it does not compile and throws a lot of  error messages like this:

```
TFT\src\User\Menu\Move.c:206:33: error: 'YGCODE_UP' undeclared (first use in this function); did you mean 'YGCODE_INC'?
```

It does seem like the renaming of variables was not done everywhere because it compiles fine with this option disabled.

### Steps to reproduce

1. Change `//#define MENU_LIST_MODE` to `#define MENU_LIST_MODE`
2. Compile

**Expected behavior**
Should compile without errors and warnings.

**Actual behavior**
Compiles with some errors as described above about variables not existing. And a ton of warnings about missing braces around initializers.

```
TFT\src\User\Menu\PreheatMenu.c:5:26: warning: missing braces around initializer [-Wmissing-braces]
 MENUITEMS preheatItems = {
```

### Additional Information
[Configuration_h.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/4014302/Configuration_h.zip)

guruathwal:
Fixed here #279 

you can ignore the warnings, just compile again and they will disappear.

guruathwal:
@zech please close the issue

zech:
I can do when it's merged. Why isn't it so far?

guruathwal:
BTT is slow when it comes to merging. until then you can download the source with the fix from this repository: https://github.com/guruathwal/BIGTREETECH-TouchScreenFirmware

zech:
I see. The PR builds fine. Thanks. I'll close it then.

Issue #274, opened 2020-01-01
[BUG] Custom Gcode not working in list mode


gneu42:
### Description

 On TFT35 V3.0 Unified Menu, Custom Gcode does not work in list mode but works Ok in icon mode

### Steps to reproduce

1. [Add custom gcode "M84/n"]
2. [compile]
3. [Test on printer]

**Expected behavior**
work as expected

**Actual behavior**
 Does nothing 

### TFT35 V3.0 connected to SKR 1.3

guruathwal:
Thanks for notifying the issue... Fixed here.. #279

gneu42:
Hi, 
Thank you for this new update, but the custom Gcode still does'nt work. I went back to Non list mode and they are working ok. 
What can I do to get them working ?

While in this mode I had problem with Y movments. I found a bad copy paste in move.c (see below Both Keys are sending YGCODE_UP (lines 163 & 196)
I change key 5 to YGCODE_DOWN and all is ok.

Regards
Gerard



      _case KEY_ICON_1:
        #ifdef MENU_LIST_MODE
          if(infoSettings.invert_yaxis == 1){
            storeCmd(YGCODE_DEC, item_move_len[item_move_len_i]);
          }
          else{
            storeCmd(YGCODE_INC, item_move_len[item_move_len_i]);
          }
        #else
          storeCmd(**YGCODE_UP**, item_move_len[item_move_len_i]);
        #endif
        break;
......
   
      case KEY_ICON_5:
        #ifdef MENU_LIST_MODE
          if(infoSettings.invert_yaxis == 1){
            storeCmd(YGCODE_INC, item_move_len[item_move_len_i]);
          }
          else{
            storeCmd(YGCODE_DEC, item_move_len[item_move_len_i]);
          }
        #else
          storeCmd(**YGCODE_UP**, item_move_len[item_move_len_i]);
        #endif
        break;_

guruathwal:
ok.. the Y move part is fixed now
 I unable to understand why custom g-codes are not working for you...
are you changing any part of the code ?

gneu42:
I don't know.
It works with Non list mode.

PLease find attached my configuration.h that does not work with me.

Regards
Gérard

[Configuration.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/4019845/Configuration.zip)

guruathwal:
The config looks old. Did you updated your source code?
[...]

PR #279, opened 2020-01-02
Fix compile error due to undefined references in move.c, Fix Custom gcode not working in List Mode, disable global info from showing on Status Screen and Move menu

guruathwal:
Fix compile error due to undefined references in move.c and fix move menu (#275) 
Fix Custom Gcode not working in List Mode (#274)
Disable global info from showing on Status Screen and Move menu.
add infobox icon to allow icon customization & fix parameter icon to match icon theme.
update global info icons.
add status screen infobox background in one single bmp to allow icon customization


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: '[BUG] (short description)'
-labels: ''
+labels: 'BUG'
```
extesy: Labels are usually lowercased.

guruathwal: Done👍🏼

Issue #211, opened 2019-12-03
[BUG] TFT35 v2 and SKR v1.3 M0/M1 commands cause endless Confirm prompts


kepbrown:
Screen: TFT35 V2 
Firmware: TFT35_V2.0.24.2

Board: SKR v1.3
Firmware: Marlin bugfix-2.0.x

### Description
Executing GCode from the TFT on board SD card slot that contains an M0 command causes an endless loop of screens displaying 
echo: 
busy: paused for user 

If I click the Confirm button the screen clears then pops back up. The only way to stop is to reset the screen.

I haven't attempted to run it from the board SD slot. I assume it would work the same way.

### Steps to reproduce
Print a gcode file with the following commands

;FLAVOR:Marlin
G28 ;Home
G1 Z10 ; Lift Z axis
G1 X32 Y35 ; Move to Position 1
G1 Z0
M0 Test Output ; Pause print
M106 S0 ; turn off cooling fan
M104 S0 ; turn off extruder
M140 S0 ; turn off bed
M84 ; disable motors
M82 ;absolute extrusion mode
M104 S0
;End of Gcode

**Expected behavior**
The prompt comes up but once Continue is pressed it moves on to the next step

**Actual behavior**
Endless loop

### Additional Information
The file being ran is attached. txt extension added to allow it to be uploaded.
[pause_test.gcode.txt](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/3914766/pause_test.gcode.txt)

guruathwal:
with M0 command the printer will wait for the user to press continue button on the **LCD** not the **TFT**. untill you press it the printer will send the busy command to the TFT, which will result in the same thing you are observing.

Doenermaker:
Wouldn't it be possible to send M108 when clicking 'confirm'? M108 needs to be enabled in Marlin, but to my understanding it should work as well as pressing the lcd button.
I have the same problem after M125 in the gcode. My guess is M600 would also behave the same, haven't tried tbh. My TFT35v1.2 is the only display connected, so pressing the lcd button isn't an option for me.

kepbrown:
I have the same issue with only having the TFT connected, I can't get my LCD to work with my SKR v1.3. It would be nice to be able to find a solution that doesn't require the LCD.

moag1000:
[...]

PR #297, opened 2020-01-12
Intercept M0/M1 command when printing from SD card on TFT controller

guruathwal:
enable detection of M0/M1 command when printing from SD card on TFT controller and prevent 'paused for user' prompt (issue #211)

**requires `EMERGENCY_PARSER` to be enabled in marlin firmware**


thisiskeithb:
@guruathwal: `EMERGENCY_PARSER` is not implemented for STM32 yet, but will work for LPC and AVR-based boards.

PR #311, opened 2020-01-25
Encoder control

bcmob:
### Requirements

For devices with encoder

### Description

Adds the ability to use the encoder on the touch screen.
I use on TFT24-V1.1

### Benefits

Significant Usability Improvement


AntoszHUN:
Hi! There is a SKR TFT35 V3.0 display. Chip: STM32F207VC My problem is that only firmware upgrade is possible with STM32F103VC. If I know it well this chip is 100MHz only. The other is 120MHz. Is there any way to solve that F207VC is using it?

PR #329, opened 2020-02-03
add MKS_32_V1_4 variant

darkspr1te:
Added MKS TFT 28/32" screen variant
supports bootloader/no bootloader models
uart2 - > aux1 ramps board
uart1 wifi txd/rxd
uart3 wifi tx,rx 
sd card - spi1 
eeprom - spi1
touch spi2 
i2cprom - i2c1 
usb host enabled
pb0/pb1 -> filiment runout/power
pb4/pb5 - > ps ok/unused

PR #331, opened 2020-02-05
Feature Request - UBL

ETE-Design:


guruathwal:
@ETE-Design This PR has some files changes and commits that look very similar to PR #329 by @darkspr1te like adding support for MKS displays.
can you check and confirm to avoid confusion as it might end up with too many duplicate files and merge issues.

darkspr1te:
The works by @delwinbest is similar to my PR #329 as they are collaborated patches/works , as mine is not building(remotely only, local build works fine) , I dont think this feature is related to UBL but to adding of a new variant,

guruathwal:
@darkspr1te  I tried to compile you PR on my PC but it failed for TFT35 V3

ETE-Design:
@delwinbest Have just tryed your firmware on an MKS TFT28, work great, exept offset / babystep in the level menu seems not to work... Should be enough to do this in Marlin 2.0 right:
#define BABYSTEPPING
#define BABYSTEP_ZPROBE_OFFSET
//#define MIN_SOFTWARE_ENDSTOP_Z

Scope666:
Can't use anything newer than 2.0.3 Stable ... Bugfix and 2.0.4 are broken:

https://github.com/MarlinFirmware/Marlin/issues/16889

PR #393, opened 2020-03-13
Enable fullscreen Marlin Mode by default

thisiskeithb:
### Requirements

All supported TFTs except for TFT24.

### Description

Enable fullscreen Marlin Mode by default.

### Benefits

Use full LCD screen for Marlin Mode.

### Related Issues

None.

CCing @guruathwal for comment/review.

### DIFF
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -64,10 +64,10 @@
 //#define ST7920_BANNER_TEXT "LCD12864 Simulator"
 
 /**
- * Run Marlin Mode fullscreen.
- * Not recommended for TFT24.
+ * Disable fullscreen Marlin Mode
+ * Recommended for TFT24
  */
-//#define ST7920_FULLSCREEN
+//#define ST7920_WINDOWED
 
 /**
  * CLEAN MODE SWITCHING SUPPORT
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -2,7 +2,7 @@
 #define _CONFIGRATION_H_
 
 //===========================================================================
-//=========================== Marlin Mode Settings ===========================
+//=========================== Marlin Mode Settings ==========================
 //===========================================================================
 
 /**
@@ -12,6 +12,8 @@
  *
  * Current color options from lcd.h: BLACK, BLUE, BROWN, BRRED, CYAN, GBLUE, GRAY, GREEN, MAGENTA, RED, WHITE, YELLOW
  */
+#define ST7920_BKCOLOR BLACK
+#define ST7920_FNCOLOR YELLOW
 
 /**
  * This setting determines the communication speed of the printer.
@@ -52,11 +54,6 @@
  */
 #define LCD_FEEDBACK_FREQUENCY_DURATION_L_US 11   // Default 11
 #define LCD_FEEDBACK_FREQUENCY_H_US          3    // Default 3
-
-// Marlin Mode Background & Font Color Options
-// Current color options from lcd.h: BLACK, BLUE, BROWN, BRRED, CYAN, GBLUE, GRAY, GREEN, MAGENTA, RED, WHITE, YELLOW
-#define ST7920_BKCOLOR BLACK
-#define ST7920_FNCOLOR YELLOW
 
 /**
  *  Text displayed at the top of the LCD in Marlin Mode.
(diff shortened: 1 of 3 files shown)


Hukuma1:
Any chance we could please add an option to stretch it 98%? Just slightly less than 100% so it’s not right up against the screen?

thisiskeithb:
> Any chance we could please add an option to stretch it 98%? Just slightly less than 100% so it’s not right up against the screen?

You can enable the `ST7920_WINDOWED` option if you prefer a border.

Hukuma1:
Is that the old default behavior? If so, I mean for the border to be very slight. The case I have over the screen has very precise coverage so the text in full screen is right up against the sides of it for instance. I understand that’s the full screen stretch by definition, but wondering if we could unlock the multiplier to tweak the setting possibly?

thisiskeithb:
> wondering if we could unlock the multiplier to tweak the setting possibly?

But that’s more code complexity 😉

I’m not sure how widely it’ll be used, but you can play around with the values in [`ST7920_Simulator.h`](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/TFT/src/User/API/UI/ST7920_Simulator.h) and see what works for your TFT case.

Hukuma1:
Yeah, I hear you. I think I tried a while back to change the hardcoded values and it didn't seem to do anything. Oh well, no biggie. But yeah full screen being default is great!

TFT/src/User/SanityCheck.h
```diff
@@ -25,5 +25,8 @@
   #define STARTUP_KNOB_LED_COLOR 1
 #endif
 
+#ifdef ST7920_FULLSCREEN
+  #error "Fullscreen Marlin Mode is enabled by default and ST7920_FULLSCREEN is now ST7920_WINDOWED. Please update your configuration."
+#endif
```
guruathwal:
```suggestion
#if defined(TFT24_V1_1) && !defined(ST7920_WINDOWED) 
  #error "Fullscreen Marlin Mode is not recommended for TFT24 Variants. Please Enable ST7920_WINDOWED in Configuration.h."
#endif
#ifdef ST7920_FULLSCREEN
  #error "Fullscreen Marlin Mode is enabled by default and ST7920_FULLSCREEN is now ST7920_WINDOWED. Please update your configuration."
#endif
```

thisiskeithb: `TFT24_V1_1` doesn't every get defined before the sanity check, so this won't work.

TFT/src/User/Configuration.h
```diff
@@ -53,21 +55,16 @@
...
 //#define ST7920_BANNER_TEXT "LCD12864 Simulator"
 
 /**
- * Run Marlin Mode fullscreen.
- * Not recommended for TFT24.
+ * Disable fullscreen Marlin Mode
+ * Recommended for TFT24
  */
-//#define ST7920_FULLSCREEN
+//#define ST7920_WINDOWED
```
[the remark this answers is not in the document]
thisiskeithb: What do you mean? It will now always be fullscreen unless you enabled `ST920_WINDOWED` (in the case of TFT24).

TFT/src/User/SanityCheck.h
thisiskeithb: This just checks for an old config setting.

guruathwal: I just tried to compile with the non-recommended and it works.

thisiskeithb:
I'll leave it as-is.

TFT/src/User/Configuration.h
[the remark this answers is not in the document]
guruathwal: Yes, you are right. It will allow fullscreen.
 But enabling `#define ST7920_FULLSCREEN` by default will also mean same thing.

Hukuma1:
> I’m not sure how widely it’ll be used, but you can play around with the values in [`ST7920_Simulator.h`](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/TFT/src/User/API/UI/ST7920_Simulator.h) and see what works for your TFT case.

Thanks for this! Btw I can't seem to find the native resolution of the display. Is it 480x320?

guruathwal:
@Hukuma1 for TFT35 480x320 & for TFT28/TFT24/TFT32 320x240

PR #406, opened 2020-03-14
Cleanup readme

thisiskeithb:
### Description

Cleanup/simplify readme: [Preview](https://github.com/thisiskeithb/BIGTREETECH-TouchScreenFirmware/blob/pr/cleanup_readme/README.md)

### Benefits

Easier update process for first-time users.

### Related Issues

None.


Scope666:
Nice changes!  Definitely will help the new users.

guruathwal:
Awesome.... now it looks much better...

Hukuma1:
@thisiskeithb What about if each download release had a simple .bat file for new users? So it can easily copy/overwrite to the proper locations? e.g. `Install Classic UI.bat` and `Install Unified UI.bat` (and if CUR files already on there it could delete them as well to prevent certain files from not being updated) Either way, thanks for coming back to contributing in full swing!

thisiskeithb:
> What about if each download release had a simple .bat file for new users? So it can easily copy/overwrite to the proper locations?

Batch files won’t work on macOS or Linux.

You can specify your own upload port/drive in `platformio.ini`, but that doesn’t work reliably for me. Drag and drop is probably the easiest method for most users.

thisiskeithb:
I added a bit more text to recommend an SD card <8GB and formatted as FAT32 since several users have tried to use 32/64GB cards and their icons/fonts folder can't be read so TFTs aren't fully updated.

I also explicitly called out inserting the card into the TFT and power cycling the printer to initiate the update process since it was just assumed before.

thisiskeithb:
@guruathwal What's the purpose of the "Firmware archive" folders? Those should all be available via the [Releases](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/releases) tab, no?

guruathwal:
> @guruathwal What's the purpose of the "Firmware archive" folders? Those should all be available via the [Releases](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/releases) tab, no?

The archive contains firmware for Minor-Version change in-between the releases.

thisiskeithb:
Sounds like there should be more releases 🙂

Issue #404, opened 2020-03-14
Use the beeper for its intended purpose!


Aleks-MO:
**Use the beeper for its intended purpose!**
![84460321_201735407811598_7486293520672620544_n](https://user-images.githubusercontent.com/22274265/76682766-6b58b000-6607-11ea-968e-2d88d8530bbf.jpg)
**Start of printing and end of printing I have a beeper on LSD . Can someone write it in the program?**
![212121](https://user-images.githubusercontent.com/22274265/76682861-0f425b80-6608-11ea-97bc-251b05dfc2de.jpg)

Msq001:
OK, Has been added to the my work plan and will be implemented tomorrow

Aleks-MO:
@Msq001  
Put the implementation in the program here.

Msq001:
hello, has be implemented in #415 , you can test it

Aleks-MO:
@Msq001   
Thank you very much, I will check.

PR #415, opened 2020-03-17
add M300 support for buzzer

Msq001:
### Description

support M300 for TFT Buzzer.

### Related Issues
for #404


Guilouz:
@Msq001 Just try, beep working when command is sending via Terminal on touch screen or via gcode with SD/USB on screen but not working when sending via Terminal with Octoprint, seem there is an issue with serial.

PR #422, opened 2020-03-19
Config cleanup & reorganization

thisiskeithb:
### Description

Cleanup config/reorganize settings.

### Related Issues

None.


thisiskeithb:
@guruathwal / @bigtreetech The default text color for Marlin Mode was changed to `YELLOW` at some point. Should it be reverted to `GREEN`?

`WHITE` would also be a good default color to make Marlin Mode appear more OLED-like.

guruathwal:
@thisiskeithb sure

PR #435, opened 2020-03-21
PR-#Keep knop led in marlin mode

GeminiServer:
KEEP_KNOP_LED_COLOR_MARLIN_MODE

Configuration setting - to keep the knop led on Marlin mode. Otherwise, led will be uninitialized.


thisiskeithb:
@GeminiServer Which editor are you using? It's inserting a bunch of whitespace.

PR #436, opened 2020-03-21
PR-#Resetting all settings from within the UI menu.

GeminiServer:
Able to reset the customer setting from the menu. It will bring a waring dialog to ensure to reset by mistake.  So no need to add a file to sd-card to reset the settings. Ignore lcd mode only if you reset all settings and to show the info dialog. Otherwise, it won't show if you currently in CLEAN_MODE_SWITCHING_SUPPORT.

### DIFF
--- a/TFT/src/User/Menu/FeatureSettings.c
+++ b/TFT/src/User/Menu/FeatureSettings.c
@@ -43,6 +43,9 @@ const LABEL itemMoveSpeed[ITEM_SPEED_NUM] = {
                                             };
 const  u8 item_movespeed[ITEM_SPEED_NUM] = {LABEL_NORMAL_SPEED, LABEL_SLOW_SPEED, LABEL_FAST_SPEED};
 
+#define ACTION_NUM   1
+const LABEL itemAction[ACTION_NUM]       = {LABEL_RESET};
+
 //
 //add key number index of the items
 //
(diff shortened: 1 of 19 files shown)


thisiskeithb:
> So no need to add a file to sd-card to reset the settings

This only works if you can access the UI. If the touch calibration is wrong, the SD card reset method is the only way to reset it.

TFT/src/User/Menu/FeatureSettings.c
```diff
@@ -43,6 +43,9 @@ const LABEL itemMoveSpeed[ITEM_SPEED_NUM] = {
                                             };
 const  u8 item_movespeed[ITEM_SPEED_NUM] = {LABEL_NORMAL_SPEED, LABEL_SLOW_SPEED, LABEL_FAST_SPEED};
 
+#define ACTION_NUM   1
+const LABEL itemAction[ACTION_NUM]       = {LABEL_RESET};
+
```
guruathwal: These lines are not required as the Custom Value of the `Reset settings` item does not changes.
```suggestion
```

TFT/src/User/Menu/FeatureSettings.c
```diff
@@ -208,6 +236,12 @@ void updateFeatureSettings(uint8_t key_val)
     break;
     #endif
 
+    case SKEY_RESET_SETTINGS:
+    settingPage[item_index].valueLabel = itemAction[0];
```
guruathwal:
```suggestion

```

TFT/src/User/Menu/FeatureSettings.c
```diff
@@ -280,12 +314,15 @@ void loadFeatureSettings(){
         settingPage[item_index].icon  = toggleitem[infoSettings.file_listmode];
         featureSettingsItems.items[i] = settingPage[item_index];
         break;
-
+      case SKEY_RESET_SETTINGS:
+        settingPage[item_index].valueLabel = itemAction[0];
```
guruathwal:
```suggestion

```

PR #462, opened 2020-03-26
Remove whitespace

thisiskeithb:
### Description

Remove whitespace. There's still more cleanup & formatting that can be done, but this will take care of the bulk of the cleanup effort.

### Related Issues

None.


guruathwal:
I will recommend leaving file in `TFT/src/Libraries` unchanged as those files are standard libraries files issued by STMicroelectronics

thisiskeithb:
> I will recommend leaving file in `TFT/src/Libraries` unchanged as those files are standard libraries files issued by STMicroelectronics

I've reverted those files.

thisiskeithb:
Looks like they were already modified in https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/61c8d873585acd60ed91178ef50d5f4a29c6fded.

TFT/src/User/Hal/sd.h
```diff
@@ -1,60 +1,60 @@
...
+#define CMD18   18      //����18����Multi ����
+#define CMD23   23      //����23�����ö�����д��ǰԤ�Ȳ���N��block
+#define CMD24   24      //����24��д����
+#define CMD25   25      //����25��д�������
+#define CMD41   41      //����41��Ӧ����0x00
+#define CMD55   55      //����55��Ӧ����0x01
+#define CMD58   58      //����58����OCR��Ϣ
+#define CMD59   59      //����59��ʹ��/��ֹCRC��Ӧ����0x00��
+
+/*����д���Ӧ������*/
```
guruathwal: @thisiskeithb the encoding is still being broken. I think there is some problem with how the Repo is being downloaded and uploaded by Git or with the editor.

thisiskeithb: I’ll look into it some more since there’s also a bunch of conflicts.

thisiskeithb: Encoding is now handled in [`.editorconfig`](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/.editorconfig), and everything looks OK in git. 🤷‍♂️

Translating everything to English is probably the best route for consistency, so I’ll add it to my to-do list for the weekend.

PR #467, opened 2020-03-26
Update 'Stop' button to 'Back' when print is finished

GregSKR:
### Description

When print is finished on unified menu, the screen still show the 'Stop' button (which is not logical) and when you click on it it will go to the previous screen.
Now the button decoration is changes to 'Back'.

### Benefits

* Visually showing on the screen that the print is finished
* Showing real effect of the button


AntoszHUN:
I didn't know where to write. I downloaded the TFT firmware again. But it is not connected to the motherboard. Something was stabbed. Everything works in Marlin mode but nothing works in Touch mode. It says "done" but doesn't detect anything.

PR #475, opened 2020-03-28
PR-#2-Moved marlin mode fullscreen to iu setting menu.

GeminiServer:
Since i always forget to set this in the configuration, i decided to move this settings into the menu. Removed the define from configuration.h file. Default is off.


thisiskeithb:
Looks like a lot of unrelated commits made it into this PR.

thisiskeithb:
Keeping the define would still be good for those of us that rebase configs and prefer it on 🙂

PR #486, opened 2020-03-31
PR-Marlin fullscreen mode in menu setting

GeminiServer:
Seems to that the .editorconfig touched some whole files.

Now user can switch to marlin in fullscreen mode in settings menu. The default is disabled and can be predefined with the configuration.h.


thisiskeithb:
> Seems to that the .editorconfig touched some whole files.

There's still a lot of whitespace to cleanup and I haven't submitted a follow up PR to remove it 🙂

PR #489, opened 2020-03-31
clean up lib, move cs pin from spi to application

Msq001:


thisiskeithb:
@Msq001 Have you tested these changes on a TFT35 or TFT35 E3?

I'm looking through the code now to see what changed, but it breaks the icons & fonts:
image

Msq001:
@thisiskeithb  hello, I have tested it in all version TFT,  Can you try to test the latest version after #502 ?  I compiled it and tested it, and everything was OK. 
I'm going to test https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/4e6d6f101ad5c9a34c4a6e8bcf01d18c99996408 now to see if I can reproduce the problem

thisiskeithb:
I just tested now and have another PR to put in since the backlight timeout doesn't work on the TFT35 E3 V3.

thisiskeithb:
Here's the fix: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/505

PR #510, opened 2020-04-01
move display colors to separate file

guruathwal:
- Move display colors to a separate file,
~- Remove double defining of extruder/tool count (Replace TOOL_NUM with EXTRUDER_NUM).~
  (it turns out it is need if hot-end count and extruder count differ.
- move renaming reset.txt at the end of reset.


TFT/src/User/Configuration.h
```diff
@@ -59,7 +59,7 @@
  *
  * These colors can be changed in Touch mode, but can also be set here.
  *
- * Options: BLACK, BLUE, BROWN, BRRED, CYAN, GBLUE, GRAY, GREEN, MAGENTA, RED, WHITE, YELLOW
+ * Options: WHITE, BLACK, RED, GREEN, BLUE, CYAN, MAGENTA, YELLOW, ORANGE, PURPLE, LIME, BROWN, DARKBLUE, DARKGREEN, GRAY, DARKGRAY
```
thisiskeithb, line 62: Can these be alphabetized?

guruathwal: Yes they can be

PR #514, opened 2020-04-02
improve LCD led control, add common PWM Hal library

Msq001:
### Description
* add common PWM Hal library for LCD back light control
* modify "LED_color_PIN" to "LCD_COLOR_PIN" (Macro definition all caps principle)
* add Marlin mode and backlight control macro switch in feature settings menu
* change os timer from tim4 to basic tim7, free tim4 for PWM backlight control


thisiskeithb:
Tested on a TFT35 V3 & TFT35 E3 V3 and backlight control works as expected.

PR #520, opened 2020-04-02
create probe menu to add bltouch and touchmi + update all config

RepairSyl:
### Requirements

create probe menu to add bltouch and touchmi command

### Description

create probe submenu in to ABL menu to control bltouch and touchmi command

PR #525, opened 2020-04-03
ABL menu choix Touchmi or Bltouch 

RepairSyl:
### Requirements

- Add // Touchmi probe in ABL menu #define TOUCHMI_PROBE in configuration.h

### Description
i take remark to my previous #520, so i add choix in configuartion.h if touchmi is define you see touchmi INIT ad ZTEST in ABL menu if not define is bltouch to see
touchmi_ABL


platformio.ini
```diff
@@ -23,7 +23,7 @@
 [platformio]
 src_dir      = TFT
 boards_dir   = buildroot/boards
-default_envs = BIGTREE_TFT35_V3_0
+default_envs = BIGTREE_TFT35_V2_0
```
thisiskeithb, line 26: This should be reverted back to `BIGTREE_TFT35_V3`

RepairSyl: ok i change it

digant73:
@RepairSyl ABL was largely improved since your PR. Please can you verify if you are satisfied by the current implementation so you  can close your PR?

PR #526, opened 2020-04-03
Add support for settings update through config file, switch unified & classic menu through config file & other minor changes

guruathwal:
- Add support for updating settings through a config file without the need to recompile firmware.
- Add config update progress screen on boot to show unrecognized keywords and values.
- Config keywords settings are almost similar to G-codes.
- Almost all of the settings in Configuration.h can be changed with the config.ini file.
- Move LCD colors to a separate file.
- Move Numpad to a separate file.
- Increase firmware version to 26.1
- Changeable long text strings are save to SPI flash.
- long text strings like custom G-codes and print G-codes are loaded from SPI flash only when needed.
- Add the Unified Menu / Classic Menu option to config.ini. now both menu flavors can be switched without flashing firmware.
- Fix menu title not displaying correctly in a List menu after a reminder message is cleared.
- Add option to change Status Screen XYZ display background color, list view border & button color.
- Fix show EMERGENCY_PARSER warning only on first boot.

*(Needs updating icons & Reset settings after firmware update )*


Acenotass:
My screen is TFT35 v 1.2
Rotate_ui is not working correctly: 1
Custom G-Code Commands do not support UTF encoding, if you set the parameters in Russian, it hangs during firmware

guruathwal:
@Acenotass 
Thanks for testing. I fixed the rotation bug. 
In your example, the name length is more than 20 characters. Even then also I tried to update with it. the TFT skipped that gcode and showed an error that it was not updated. No hangs noticed. however, it will pause for a few seconds when an error is displayed on the screen to allow the user to read it.

Acenotass:
custom_label_1:Отключить двигатели
custom_gcode_1:M84\n
custom_label_2:Активировать SD карту
custom_gcode_2:M21\n
custom_label_3:Деактивировать SD карту
custom_gcode_3:M22\n
custom_label_4:Сохранённая сетка стола
custom_gcode_4:M420 s1\n
custom_label_5:Сохранить EEPROM
custom_gcode_5:M500\n
custom_label_6:Восстановить EEPROM
custom_gcode_6:M501\n
custom_label_7:EEPROM по умолчанию
custom_gcode_7:M502\n


with these values ​​my firmware update freezes

If you can’t use the UTF encoding in config.ini, then at least you need to be able to use the values ​​from the firmware from Configuration.h

guruathwal:
@Acenotass I tried those also.
it just ignored them due to long lengths and the text displayed correctly in error.
Still no hangs. here is the video gif:
InShot_20200405-1586072060556

Acenotass:
Sorry it turns out I have config.ini saved in ANSI encoding

Acenotass:
So what about UTF encoding support in the config.ini file?

guruathwal:
@Acenotass the config.ini is in UTF encoding. may by your editor is changing the encoding while saving the changes.

Acenotass:
There was a problem with displaying all the digits in all fields.
IMG_20200405_175033

guruathwal:
Just checked .. I can confirm that... adding fix...

GregSKR:
A bit confused with this option:

```
 #### Show Temprature ACK in Gcode Terminal
 # Options: [enable: 1, disable: 0]
 terminal_ack:0
```
On the TFT screen, option is "**Hide** terminal ACK".
When setting terminal_ack to 0 in config.ini, meaning not showing ACK, "Hide ACK" is disabled in the firmware...
Shouldn't this option be "**Hide** Temprature ACK in Gcode Terminal"?

guruathwal:
@GregSKR why are you geting confused the meaning is in the sentence.

thisiskeithb:
> Sorry it turns out I have config.ini saved in ANSI encoding

@Acenotass (and anyone else) I'd recommend installing the [EditorConfig plugin](https://github.com/editorconfig/editorconfig-vscode). This will keep code tidy and ensure proper file encoding. I'll put in a PR shortly to add *.ini files to the UTF-8 encoding section.

.editorconfig
```diff
@@ -10,7 +10,7 @@ charset = utf-8
 [{*.c,*.cpp,*.h,Makefile}]
 trim_trailing_whitespace = true
 insert_final_newline = true
-end_of_line = lf
+end_of_line = crlf
```
thisiskeithb, line 13: Maybe `end_of_line` should be removed completely (or at least until `native` is supported) since it's not universal across macOS/Linux/Windows.

guruathwal: I agree with you

guruathwal: > @Acenotass (and anyone else) I'd recommend installing the [EditorConfig plugin](https://github.com/editorconfig/editorconfig-vscode). This will keep code tidy and ensure proper file encoding. ~I'll put in a PR shortly to add *.ini files to the UTF-8 encoding section~: [Done](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/539).

@thisiskeithb It keeps adding a new line to the code if you are typing at the end of the file and stop for 2 seconds. I think this is due to `insert_final_newline = true`. So I set those settings in VSCodium's default settings.

thisiskeithb: Odd. Which OS are you using? I don’t have that issue on macOS or Windows.

bigtreetech:
Hello, thank you very much for your work! This PR will not be merged in a short time, because there are too many changes in it. 
My main work in the near future is to fix some bugs and test them. I plan to release a new release version this week, and this PR will be merged after release version.
Thank you again!

.editorconfig
guruathwal: Windows

Acenotass:
To create beautiful themes for the TFT screen, standard color sets are not enough. I propose to make it possible to select any color in config.ini.
As a basis, you can take color coding in HTML.
https://en.wikipedia.org/wiki/Web_colors

guruathwal:
> Hello, thank you very much for your work! This PR will not be merged in a short time, because there are too many changes in it.
> My main work in the near future is to fix some bugs and test them. I plan to release a new release version this week, and this PR will be merged after release version.
> Thank you again!

@bigtreetech 
Looking forward to it 👍

Acenotass:
What kind of color coding is in the Colors.h file. I can’t match the icon color with the background color #define BLUE 0x021F.
Is my request for execution planned?
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/526#issuecomment-610201284

guruathwal:
@Acenotass  yes I will add it. It takes time.
The colors are RGB565. The lcd uses RGB565 (aka 65k colors) . Those html colors are 16bit colors.. They cannot be matched accurately.

Acenotass:
@guruathwal thank you very much!

GregSKR:
@guruathwal Thanks a lot for your work for the community, I'm using your fork for flashing my TFT35, it's awsome!

Sorry to bother you again about 'terminal_ack'...
I see you made some changes to the config.ini, thanks for that, but the confusion was not in the meaning of the sentence, but in the effect of the value.
As a picture is better than words, let me show you what I meant:

ack

So, if I want to HIDE ack, looks like I have to set the value of terminal_ack to 1 (and not 0!) to activate the hiding option?...
Sorry if I'm wrong, just want to clarify this parameter for future users of the config file,
but shounldn't the options be **[Show: 0, Hide: 1]**?

Acenotass:
@GregSKR @guruathwal IMHO, in order not to get confused in the logic of variables, the best thing is to use Boolean values ​​True and False

thisiskeithb:
I think dropping “Hide” from the menu label would help clear up confusion. 0 should be an off/disabled state.

Or what about renaming the label to “Temperature ACK”?

Acenotass:
The logic is simple, True is always 1, and the inverse logic is Hide. It follows that it makes sense to use Boolean values. Hide = True = 1

Acenotass:
I noticed an bug in displaying files when printing through the motherboard. If the files in list mode is disabled, then the files are still displayed in list mode.

guruathwal:
> @guruathwal I noticed an bug in displaying files when printing through the motherboard. If the files in list mode is disabled, then the files are still displayed in list mode.

@Acenotass It is not a bug. It is deliberate because the preview images cannot be loaded from motherboard SD card. So there is no point in showing icons for files in motherboard SD card.

Acenotass:
@guruathwal I see... BTW, how to form a preview image I did not understand. Nowhere is this written clearly.

guruathwal:
@Acenotass  this might help you: https://github.com/bigtreetech/BiQU_Tool

Acenotass:
@guruathwal 
I tried to create pictures with this application, but I didn’t see anything. The application is very raw, the problem with switching languages.

GregSKR:
Thanks!
Just check your lastest commit 'update show ack check' looks like lines have not been merged at the good place in parseACK.c:

parseACK c
After that, it's perfect!

guruathwal:
@GregSKR Thanks for pointing out... It got mixed during rebase conflicts.

GregSKR:
Did some tests, and I found I couldn't set the fan speed to 100%...

Found a typo in Settings.c, line 82:
`infoSettings.fan_max[i]           = default_max_temp[i];`

After changing the line to:
`infoSettings.fan_max[i]           = default_max_fanPWM[i];`

and resetting the parameters, I could set my fan speed to 100%!

guruathwal:
@GregSKR  thanks for checking...

Acenotass:
@guruathwal 
How are things going with the implementation of WEB color coding?

GregSKR:
I saw a few things in the code:
- **flashStore.h** line 8 and 10 are duplicates, is it intended?

- **FeatureSettings.c** curly bracket line 399 should be moved line 395 if I'm not wrong?

guruathwal:
@GregSKR  thanks for notifying

ssombra:
I am interested in updating through a config file, how does it work? any guide?

Acenotass:
@guruathwal
In the latest commit version, the fan is incorrectly regulated, maximum 7%

guruathwal:
@Acenotass resetting the settings or reloading the config file should rectify it.

roberttco:
@acenotass - what are you asking for with the colors ?  Since the colors are n RGB565 color space you could use an online tool.  There are several.

Acenotass:
@roberttco the problem is not with color conversion, but with the impossibility of choosing it, there is no way to choose any color.

roberttco:
Ahhh I see.  That makes Morse sense to me now.  Thanks.

extesy:
@bigtreetech When do you plan to merge this?

GregSKR:
**Just for the beauty of the code**: 😉 

- in Settings.c, the 3 following parameters are duplicated (lines 42-44 and 57-59):
```
infoSettings.send_start_gcode     = 0;
infoSettings.send_end_gcode       = 0;
infoSettings.send_cancel_gcode    = 1;
```
- and really not important 😄  in FeatureSettings.c lines 401-402 and 410-411:
Is it nicer to place the 'break' before or after the curly bracket?

guruathwal:
@GregSKR  done #663, and the `break` inside or outside of the curly brackets depends on how lazy the programmer is feeling is at the time of typing it. 😂

PR #562, opened 2020-04-10
Code to handle pause outside of printing

mikeshub:
This is useful for setups that need to pause for a leveling probe to be attached on ABL. My uses PAUSE_BEFORE_DEPLOY_STOW and if it doesn't send M108 the printer is hard frozen.

### Description

There is not too much to this change. Now when "paused for user" comes in and the machine is not printing a pop up comes up to let the user hit OK to continue. 

### Benefits

This makes the screen more usable. This way people don't get into a situation where the machine is waiting for a button press without a way to get one.

### Related Issues

None that I am aware of. Why does M108 need to be sent using 

`Serial_Puts(SERIAL_PORT, "M108\n");`

The same code is found in Printing.c line 317. Must store did not result in the command ever being sent. 

Notably sending M0 then M108 through the screen's little console doesn't register as an OK. That likely has to do with whatever is going on with the requirement for Serial_Puts

### DIFF
--- a/TFT/src/User/API/parseACK.c
+++ b/TFT/src/User/API/parseACK.c
@@ -10,7 +10,6 @@ const char *const ignoreEcho[] = {
   "busy: processing",
   "Now fresh file:",
   "Probe Z Offset:",
-  "paused for user",
   "Flow:",
   "echo:;",
   "echo:  G",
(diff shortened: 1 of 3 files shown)


mikeshub:
So what's up with it not handling pause while printing? I have no interest in hooking the runout sensor to screen. The screen needs to handle pauses and ideally it should properly handle runouts.

bigtreetech:
> This works for the pause that I get during the leveling process. However I need the screen to handle "paused for user" in the printing menu. This will allow me to handle the runout sensor that is onboard the machine. I do not want to use the runout sensor connected to the screen.

Handling Fiament detect from onboard printing need more programming,  I am planning to improve the function of printing onBoardSD, It will come true in the near future.

PR #586, opened 2020-04-18
Set the Z Offset with a moving nozzle

xboxpro1:
Set the nozzle test point in Configuration.h.
Set the Z Offset with a moving nozzle. Beware Software Endstop is OFF!


TFT/src/User/API/Language/language_de.h
```diff
@@ -183,6 +184,9 @@
     #define DE_120_SECONDS              "2 Min."
     #define DE_300_SECONDS              "5 Min."
     #define DE_CUSTOM_SECONDS           "Benutz."
-    #define DE_ST7920_FULLSCREEN        "Marlin mode in fullscreen"
+    #define DE_ST7920_FULLSCREEN        "Display mode in fullscreen"
+    #define DE_SOFT_ENDSTOP_OFF         "\u0881 Kein Endstop \u0881"
+    #define DE_MARLIN_MODE              "Display Mode"
+    #define DE_TOUCH_MODE               "Touch Mode"
```
guruathwal, line 190: the word 'Display' here is ambiguous here.
Marlin Mode / ST7920 Mode/ 12864 Mode are best suited here.

TFT/src/User/Configuration.h
```diff
@@ -62,12 +66,12 @@
...
  */
 #define ST7920_BKCOLOR BLACK
-#define ST7920_FNCOLOR YELLOW
+#define ST7920_FNCOLOR CYAN
 
 /**
  *  Text displayed at the top of the TFT in Marlin Mode.
  */
-//#define ST7920_BANNER_TEXT "LCD12864 Simulator"
+//#define ST7920_BANNER_TEXT "XY-2 PRO"
```
guruathwal, line 74: please remove settings specific to your printer and try to use default values in PRs.

TFT/src/User/API/Language/language_de.h
xboxpro1: yes, you are right. the main purpose is to change the term in every language as you like.
what du you think about "LCD Mode"?

TFT/src/User/Configuration.h
xboxpro1: ok, i will do it in the future. thank you for your help.

TFT/src/User/API/Language/language_de.h
thisiskeithb: “Marlin Mode” and “Touch Mode” is what’s used now to differentiate between the two modes, which works fine.

TFT/src/User/Configuration.h
thisiskeithb: Please update this PR to to remove your specific printer settings. It’ll get merged otherwise unless BigTreeTech reverts the change.

xboxpro1: done..

bigtreetech:
Hello, Can you make this PR "editable by maintainers," please? That I can resolve some file conflicts and restore the default configuration. There should be a checkbox on the right side of this page someplace,

xboxpro1:
Hello, thank you for your reply. I have deleted the fork . I want to improve the code on the weekend.
And then i make a new pull request. Thank you for your time.

Issue #584, opened 2020-04-18
[solved] SKR E3 Mini + TFT35 E3 V3. babysteps doesn't work


nicolas81:
Baby steps doesn't work : When you click + or - on the touch screen, the nozzle doesn't move.

drphil3d:
I seem to be having this issue as well

xboxpro1:
Hello, have you uncomment the line
//#define BABYSTEPPING in Configuration_adv.h / Marlin Firmware?
<https://marlinfw.org/docs/configuration/configuration.html#babystepping>

nicolas81:
Yes of course. When I use Marlin mode, babystepping works well. I think that this issue come from TFT35 firmware.

xboxpro1:
I can confirm that babystepping is working while printing. TFT35 E3 V3 + SKR V1.4.

nicolas81:
I have SKR E3 MINI V1.2.
 BabySteps are working correctly with the Marlin screen but not with BTT TFT E3 V3 screen for me...

nicolas81:
I noticed today that babysteps are working correctly within the Classic Menu but not with in the Unified Menu.

xboxpro1:
I think the baby step commands are not sent because the tft rs232 buffer is full.
try to edit BabyStep.c / TFT Firmware. Replace the command "storeCmd" with "mustStoreCmd" in BabyStep.c. The command occurs 4 times in the file. And the compile and flash TFT firmware.

nicolas81:
I tried but I get this error :

![erreur](https://user-images.githubusercontent.com/13489064/79873949-110df480-83e8-11ea-9b10-0e7ac4f4748c.png)

nicolas81:
I understood that mustStoreCmd is a void instead of storeCmd is bool function. So I will try this modification
![correction1](https://user-images.githubusercontent.com/13489064/79876098-e3767a80-83ea-11ea-99cf-6d82d5442429.png)

nicolas81:
Now, babysteps  are working  fine with + and - button. Thanks xboxpro1 !
Still remains problem with the encoder. As I am seeing in the code the encoder change the value of baby_step_value but doesn't send gcode command to the mother board. So the nozzle doesn't move with the encoder. And I don't understand the goal of the save button in this menu.

xboxpro1:
[...]

PR #599, opened 2020-04-22
Babystep fix

xboxpro1:
Babysteps are not working correct while printing over tft/rs232. Maybe the rs232 buffer is too busy.
Changed command from StoreCmd to mustStoreCmd. The Encoder did not work either. Also the Save Icon is obsolete, there is nothing to save..

This bug was found by nicolas81. Please see <https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/584>


xboxpro1:
.

PR #600, opened 2020-04-22
Babystep fix

xboxpro1:
Babysteps are not working correct while printing over tft/rs232. Maybe the rs232 buffer is too busy.
Changed command from StoreCmd to mustStoreCmd. The Encoder did not work either. Also the Save Icon is obsolete, there is nothing to save..

This bug was found by nicolas81. Please see bigtreetech#584


radek8:
Why do you want to delete the save button?

xboxpro1:
Hello, you can only save the babystep value when BABYSTEP_ZPROBE_OFFSET is enabled.
With the M500 command you change your Z Offset.
Please see <https://marlinfw.org/docs/gcode/M290.html>
BABYSTEP_ZPROBE_OFFSET disabled is recommended to avoid a side-effect.

PR #627, opened 2020-04-27
Add auto-detect On-board SD Card via M115 and change PRINT_CANCEL_GCODE behaviour to be sent only when printing from TFT's SD card

guruathwal:
- With this update, the TFT firmware will be able to auto-detect Onboard SD Card with M115 starting from Marlin Bugfix 2.0.x date: 2020-04-27.
- Enable PRINT_CANCEL_GCODE by default and change PRINT_CANCEL_GCODE behaviour to be sent only when printing from TFT's SD card and USB flash disk to avoid duplicate Cancel G-codes when printing from Onboard SD card.

### DIFF
--- a/TFT/src/User/API/parseACK.c
+++ b/TFT/src/User/API/parseACK.c
@@ -205,13 +205,12 @@ void parseACK(void)
         coordinateSetAxisActualSteps(E_AXIS, ack_value());
       }
 
-  #ifdef ONBOARD_SD_SUPPORT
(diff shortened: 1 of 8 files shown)


GregSKR:
Nice!
So we don't need anymore "**onboard_sd_support:1**" in **config.ini** (and this parameter gives now an error when flashing config.ini)...
Or maybe we could change it to a new parameter "force_sd_support" for peoples with firmware older than "Bugfix 2.0.x date: 2020-04-27"?

guruathwal:
@GregSKR  done. #663

PR #643, opened 2020-05-01
Set the z offset with a moving nozzle

xboxpro1:
Set the nozzle test point in Configuration.h.
Set the z offset with a moving nozzle. Beware Software Endstop is OFF!

Maybe someone can create a popup with the info "Software Endstop is off" when entering
the z offset menu instead of the text "No Endstop".


digant73:
@xboxpro1 A similar solution is now present. Please can you verify and eventually remove your PR if you are satisfied?

PR #670, opened 2020-05-09
Fix heater indexes

creamywhite:
### Description

Fix for wrong heater indexes issue that caused last heater to be ignored in some functions

### Benefits

M109 will work as intended when printing from TFT

### Related Issues

None that I know of.


AntoszHUN:
Does this solve the problem of starting to print without a heated extruder?

creamywhite:
@AntoszHUN, if that problem is caused by ignorance of M109 gcode command, then yes.

martinj:
Can we get this merged? This problem is really annoying.

digant73:
it is fixed by #663

PR #672, opened 2020-05-09
Fix send_cancel_gcode in abortPrinting

rebel1:
### Description
Fix misplaced send_cancel_gcode in abortPrinting

### Benefits

Enabled Cancel Gcode works now as expected, before it was not executed 

### Related Issues

Cancel Gcode doesn´t work on TFT-SD and U_DISK

### DIFF
--- a/TFT/src/User/Menu/Printing.c
+++ b/TFT/src/User/Menu/Printing.c
@@ -718,13 +718,13 @@ void abortPrinting(void)
 
     case TFT_UDISK:
     case TFT_SD:
-      if (infoSettings.send_cancel_gcode == 1)
-        mustStoreCmd(PRINT_CANCEL_GCODE);
-
-      clearCmdQueue();
+      
       break;
   }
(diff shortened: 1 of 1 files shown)


guruathwal:
this change was done to prevent sending cancel gcode when printing from onboard SDCard as @thisiskeithb reported that the cancel code was doubling up.

thisiskeithb:
Marlin has it's own cancel gcode, so the TFT's cancel gcode should be optional.

digant73:
> this change was done to prevent sending cancel gcode when printing from onboard SDCard as @thisiskeithb reported that the cancel code was doubling up.

On my side it seems the cancel gcode configured in the config.ini file is not executed even if send_cancel_gcode is enabled from the feature menu. Tested printing from SD card (so gcode should be executed when enabled on infoSettings.send_cancel_gcode)

digant73:
it was also fixed by PR #663

PR #678, opened 2020-05-11
Fixed minor formatting error in config.ini

SonBroku:
In config.ini `level_edge_distance: X20 Y20` is incorrectly set and commented by default.

The current implementation is actually expecting a single value like `level_edge_distance:20`.

Currently, if you use this setting with the X* Y* format, it will not work, and you will see the nozzle go to X0 Y0 upon requesting a lower left leveling routing. With this change, it works as it should and goes to X20 Y20.


AntoszHUN:
Does this solve the problem of going to the value I type? I played with him last night. The 4-point manual leveling didn't want to be good at all. Was that the problem then?

SonBroku:
> 
> 
> Does this solve the problem of going to the value I type? I played with him last night. The 4-point manual leveling didn't want to be good at all. Was that the problem then?

Yes it does.

PR #679, opened 2020-05-11
Fix several formatting errors in config.ini and removed duplicate entries.

SonBroku:
In config.ini the current `level_edge_distance: X20 Y20` is not correct and will cause the printer to go to X0 Y0 when you execute a lower left bed manual bed leveling command.

It seems that under the current implementation, it's actually expecting something like `level_edge_distance:20`

With this change, the nozzle moves to X20 Y20 as expected when leveling routine is run for lower left manual level. I have fixed this, as well as changed the commenting to reflect the change in formatting. 

I also removed some duplicate settings as well.

### DIFF
--- a/Copy to SD Card root directory to update - Unified Menu Material theme/config.ini
+++ b/Copy to SD Card root directory to update - Unified Menu Material theme/config.ini
@@ -192,8 +192,8 @@ pause_feedrate: X6000 Y6000 Z6000 E600
 # Leveling Z raise format        [level_z_raise: Z<distance in mm>]
 # Leveling feed Rate format      [ X<feedrate mm/min> Y<feedrate mm/min> Z<feedrate mm/min>]
 level_edge_distance:20
-level_z_pos:0.2
-level_z_raise:10
+level_z_pos:0.2f
+level_z_raise:10f
 level_feedrate:X6000 Y6000 Z6000
 
 #### Preheat Temperature
--- a/Copy to SD Card root directory to update - Unified Menu Material theme/config.ini
+++ b/Copy to SD Card root directory to update - Unified Menu Material theme/config.ini
@@ -192,8 +192,8 @@ pause_feedrate: X6000 Y6000 Z6000 E600
 # Leveling Z raise format        [level_z_raise: Z<distance in mm>]
 # Leveling feed Rate format      [ X<feedrate mm/min> Y<feedrate mm/min> Z<feedrate mm/min>]
 level_edge_distance:20
-level_z_pos:0.2f
-level_z_raise:10f
+level_z_pos:0.2
+level_z_raise:10
 level_feedrate:X6000 Y6000 Z6000
 
 #### Preheat Temperature
--- a/Copy to SD Card root directory to update - Unified Menu Material theme/config.ini
+++ b/Copy to SD Card root directory to update - Unified Menu Material theme/config.ini
@@ -198,7 +198,7 @@ pause_feedrate: X6000 Y6000 Z6000 E600
 # Leveling Z raise format        [level_z_raise: Z<distance in mm>]
 # Leveling feed Rate format      [ X<feedrate mm/min> Y<feedrate mm/min> Z<feedrate mm/min>]
 level_edge_distance:20
-level_z_pos:0.1
+level_z_pos:0.2
 level_z_raise:10
 level_feedrate:X6000 Y6000 Z6000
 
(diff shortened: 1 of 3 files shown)


Copy to SD Card root directory to update - Unified Menu Material theme/config.ini
```diff
@@ -193,14 +187,14 @@ pause_feedrate: X6000 Y6000 Z6000 E600
...
 # Leveling Z Position format     [level_z_pos: Z<position in mm>]
 # Leveling Z raise format        [level_z_raise: Z<distance in mm>]
 # Leveling feed Rate format      [ X<feedrate mm/min> Y<feedrate mm/min> Z<feedrate mm/min>]
-level_edge_distance: X20 Y20
-level_z_pos:0.2
-level_z_raise:10
-level_feedrate: X6000 Y6000 Z6000
+level_edge_distance:20
+level_z_pos:0.2f
+level_z_raise:10f
```
guruathwal, line 196: The config routine was made similar to g-codes parser. It extracts these values from text/char array with `strtod` function. it does not require `f` marker in the end. 
This is similar to parsing g-codes which also does not have the `f` marker
like `G1 X10.5 Y11.23 Z0.45`.

```suggestion
level_z_pos:0.2
level_z_raise:10
```
Many users who are not familiar with any type of programming/coding get confused with the `f` in the end.

guruathwal:
The duplicate entries were a result of bad rebase from my side.

>The way the settings are formatted has to be the same as in Configuration.h

Configuration.h and config.ini are both completely different formats.

rightawy:
I did not find any mess in config.ini apart from these changes.  config.ini looks much more better organized than configuration.h

SonBroku:
> I did not find any mess in config.ini apart from these changes.  config.ini looks much more better organized than configuration.h

Maybe I was being hyperbolic. It’s just that having entries that flat out don’t work in their default formatting as well as duplicate entries is messy to me.

digant73:
> > I did not find any mess in config.ini apart from these changes.  config.ini looks much more better organized than configuration.h
> 
> Maybe I was being hyperbolic. It’s just that having entries that flat out don’t work in their default formatting as well as duplicate entries is messy to me.
> 
> I just don’t think it’s super great when people update their firmware and their leveling menu sends the nozzle to X0 Y0 because the default option is formatted wrong with incorrect comments for formatting and duplicate entries.

there are also other errors in the config file such as:
1) ext_count:1 is repeated twice. The first one should be hotend_count:1 (according to the source code)
2) fan_max:255: it must have the format fan_max: F0:xx F1:xx etc...
3) onboard_sd_support: it is not currently supported by the source code so an error is displayed when loading the config file at boot

guruathwal:
> there are also other errors in the config file such as:
> 
> 1. ext_count:1 is repeated twice. The first one should be hotend_count:1 (according to the source code)
> 2. fan_max:255: it must have the format fan_max: F0:xx F1:xx etc...
> 3. onboard_sd_support: it is not currently supported by the source code so an error is displayed when loading the config file at boot

@digant73 these are already solved in #663

digant73:
> > there are also other errors in the config file such as:
> > 
> > 1. ext_count:1 is repeated twice. The first one should be hotend_count:1 (according to the source code)
> > 2. fan_max:255: it must have the format fan_max: F0:xx F1:xx etc...
> > 3. onboard_sd_support: it is not currently supported by the source code so an error is displayed when loading the config file at boot
> 
> @digant73 these are already solved in #663

yes sorry, I saw your pull request #663 after my comment on that pull request. I checked your last config file. Just two things you can change on the file:
1) level_edge_distance: The description is still wrong. It reports pause_pos...
2) onboard_sd_support: I should use 0 for disable, 1 for enable and 2 for auto. it's more compliant to other properties that can assume basically "true" or "false" values

radek8:
2. custom_gcode_2: G28 \ nG29 \ nG0 \ n: mělo by to být "M21 \ n"
will be fixed : https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/673

AntoszHUN:
It is not clear to me that I can configure it for my machine. I need X: 57 and Y: 28. In the old version, it was possible to set up four-point manual leveling properly. But with that being changed, I can’t do it properly. Why wasn't the old version good? The old leveling was much better. This cannot be adjusted to the four corners properly.

PR #716, opened 2020-05-25
Fix the marlin_fullscreen parameter in config.ini doesn't work.

Uz45:
### Requirements

All supported screens.

### Description
In the config.c file the code reading the marlin_fullscreen parameter from the config.ini file is missing.
### Benefits

Defining the marlin_fullscreen parameter via the config.ini file


Uz45:
> just a cosmetic review. According to the config.ini file, marlin_fullscreen parameter is placed before marlin_title parameter. So even in the code it should be put before and not after marlin_title

Yes, it makes more sense. Thanks!

PR #728, opened 2020-05-29
Restore the ability to change knob colors on TFT35 E3 v3.0, broken in release 26.1

### LINKED ISSUE #696: [BUG] Knob LED Color
I have updated my TFT35 E3 V3 to the current version (TFT_E3_V3.0.26.1)
but now the colors of the LED Knob are completely gone.
no color is correct.
OFF also no longer works.

Tygrys-1:
### Description

Changed knob LED timing from 0.4us to 0.35us to correct colors of the knob on the TFT35 E3
According to the WS2812 documentation 0.35us is the right timing to set here.

### Benefits

On some TFT35 E3 boards release 26.1 broke the knob lighting. This is caused by different tollerance of the "0" timing of different WS2812 LEDs. This PR restores the ability to set colors to the knob leds or to switch them off.
 
### Related Issues

Resolves #696

### DIFF
--- a/TFT/src/User/Menu/ledcolor.c
+++ b/TFT/src/User/Menu/ledcolor.c
@@ -59,7 +59,7 @@ void WS2812_Send_DAT(uint32_t ws2812_dat)
   uint16_t led_num;
   int8_t bit;
   uint16_t cycle = mcuClocks.PCLK1_Timer_Frequency / 800000 / 2 - 1;   // WS2812 need 800KHz (1.25us)
-  uint16_t code_0_tim_h_cnt = cycle * (0.4/1.25);  // Code 0, High level hold time, 0.4us / 1.25us
+  uint16_t code_0_tim_h_cnt = cycle * (0.3.5/1.25);  // Code 0, High level hold time, 0.35us / 1.25us; previous 0.4us was to high for some boards
   uint16_t code_1_tim_h_cnt = cycle - code_0_tim_h_cnt;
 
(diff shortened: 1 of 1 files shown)


bigtreetech:
Hell， can you test https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/728/commits/9bfedc55cc40299b2bcb6668792ffe508dcc069d ？
According to our supplier's datasheet,  its timing is as follows
image
It doesn't look like the standard ws2812 timing

Tygrys-1:
I can confirm that it is working ok. In fact, the most critical timing T0H is the same as with my solution. Other timings can be more relaxed and in this case they seems to have no effect on overall performance. [https://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/](url)

It will be nice to improve BTT Neopixel implementation for SKR mini E3 v1.2. It is currently based on 'nop' cycle counting and it is not interrupt aware. This causes strange effects on Neopixel leds when heating the bed/nozzle or moving the motors. The good news is that it uses the same PC7 pin as the TFT35 and TIM6 is not used in Marlin F1 implementation. :-)

yousssert:
Doesn't work for me
Is there another way to fix it

yousssert:
A video 

https://www.reddit.com/user/YousefAlHassan/comments/jd0jid/my_tft35_e3_v3_led_knob_dont_work/?utm_medium=android_app&utm_source=share

Issue #758, opened 2020-06-04
[BUG] No resume after change filament (M600)


VijitCoder:
### Description

I used Cura 4.x. 

Sliced a model, set M600 command for changing filament. But the printing is never resume. On the screen constantly shows _"Printer Paused. OK to continue"_. By pressing Ok this message shows again.

Also I noticed that the nozzle become colder, which is strange. Then the temperature upped to the specified value but the message toggle and toggle again.

What else. Before begin to print my model, I made a test with this feature. I took a cylinder d20, sliced it and added Change Filament command. I reduced temperature of the nozzle and the bed cause I made this print without filament. And it works as expected. Paused and resumed.

The problem model from [here](https://www.thingiverse.com/thing:4247272). _ImpossiTableTop.stl_ Change filament at layer 4. Layer height - 0.25.

Both gcodes will be attached.

### Steps to reproduce

Well, you can use my gcode, printing an hour and half. Then see what will happen on the filament change.

**Expected behavior**

Printing must resume by pressing "Ok" in the popup.

**Actual behavior**

No printing resume.

### Hardware Variant
Bigtreetech TFT35-E3 V3.0
### TFT Firmware Version & Main Board Firmware details
Screen - 26.x
Marlin - 2.x

### Additional Information

[files.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/4729203/files.zip)

marsolek84:
I too have this issue.  There are two band-aid fixes I have found that work.  I have printed it in the "Marlin emulator" and that does bring up the pop-up to resume.  The other is for touch mode.  When it gets to the filament change, you click ok, and change filament.  If you push the encoder knob in, it will purge more filament and keep doing so until you resume.  After you have purged and gotten clean filament, rotate the encoder knob clockwise one notch (bump) and push the encoder button in.

No pop-ups appear on the screen, but it will resume print.  Sure seems like a bug to me, but this works.  It isn't ideal by any means and would like to know a fix too.

Give it a try and let me know if it works for you.

HappyCodingRobot:
Marlin 2.0.x requires an LCD connected for the advanced pause features. (see Configuration_adv.h)
Afaik this cannot work with the serial connected TFT and actual Marlin versions.

VijitCoder:
[...]

PR #764, opened 2020-06-05
Adding emulated M601

HappyCodingRobot:
### Description
This is an improvement / new feature.

Prusa firmware and Prusaslicer support an advanced pause M601.
Prusaslicer can add this to the gcode to pause at certain heights e.g. insert sthg or change filament.

Marlin does not implement M601 at the moment and 
Marlins advanced pause feature (M600) only works with LCD / TFT in Marlin Mode.

### Benefits
* support this advanced pause gcode for prints from sdcard & usb
* enable advanced pause  for serial-only connected displays

### Related Issues

maybe #758

### DIFF
--- a/TFT/src/User/API/interfaceCmd.c
+++ b/TFT/src/User/API/interfaceCmd.c
@@ -668,6 +668,19 @@ void sendQueueCmd(void)
               speedSetSendWaiting(0, false);
             }
             break;
+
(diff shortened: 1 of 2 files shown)


ETE-Design:
So you only need to activate advanced pause in Marlin that's all?

ETE-Design:
@HappyCodingRobot Also If you have the runout sensor connected to your mainboard?

PR #791, opened 2020-06-12
Neopixel sequential in accurate 800Khz

Msq001:
### Requirements

### Description
1. The signal captured by the logic analyzer shows that this configuration cycle is about 1.35us, run in ~800Khz


Tygrys-1:
According to BTT the timing is accurate, as LED's are not standard WS2812. Please, take a look at #728.

traffic-light:
So... This request van je closed?
@bigtreetech

Issue #826, opened 2020-07-07
[BUG] Encoder glitch when adjusting fan speed, speed and flow


kind3r:
### Description

When adjusting fan speed, speed and flow using the encoder in TFT mode the respective value glitches into a loop going up and down in value until reboot.

### Steps to reproduce

Rotate the encoder knob fast enough when in fan, speed and flow menu.  
Made a short video about the issue: https://youtu.be/u0Gz-jvzVVs

**Expected behavior**

The value should change up and down depending on the rotation.

**Actual behavior**

The value is stuck in a glitch automatically changing up and down by a few increments until reboot

### Hardware Variant

TFT 35 V3.0

### TFT Firmware Version & Main Board Firmware details

Current firmware built from source (https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/868d61549cc480a06c6c2526b828efab5807539a).

### Additional Information

The issue is not configuration dependent and was introduced in the https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/41a11faf2384957114bb23421895fa861b6458d4 commit.

kind3r:
Finally discovered it's caused by the duplicate updating of values by the encoder and by the command queue.
Will put in a PR.

oldman4U:
Would be great if this could be fixed. Happens also when using the touchscreen and not the encoder and stops the print

kind3r:
> Would be great if this could be fixed. Happens also when using the touchscreen and not the encoder and stops the print

I had a feeling that would be the case but I could not reproduce it by only using the screen. Anyway, you can try my PR and check. For me it's working so far.

digant73:
> Would be great if this could be fixed. Happens also when using the touchscreen and not the encoder and stops the print

yes, I confirm that it happens also to me with the touchscreen. I also see that changes on speed and flow are not reset to 100% after a print. They will remain also for other prints. A reset is needed to restore default values.

kind3r:
> I also see that changes on speed and flow are not reset to 100% after a print. They will remain also for other prints. A reset is needed to restore default values.

This is not the case for everyone. You can reset them in your start or end g-code of your slicer.

gordo3di:
We had the same issue with speed and flow when just adjusting values using the + - buttons. Crashed the print. This happened on the TFT35 E3 V3 and the TFT35 V3

radek8:
[...]

PR #827, opened 2020-07-07
Fix duplicate update of speeds and flow #826

kind3r:
### Requirements

Rotary encoder BTT TFT

### Description

When using the rotary encoder after https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/41a11faf2384957114bb23421895fa861b6458d4 the values for fan speed, speed and flow were getting stuck in a loop as they were updated from both the respective menu and the command queue via g-code parsing.

### Benefits

Fix loop updating values via menu and command queue.

### Related Issues

Fixes https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/826

### DIFF
--- a/TFT/src/User/API/interfaceCmd.c
+++ b/TFT/src/User/API/interfaceCmd.c
@@ -469,20 +469,24 @@ void sendQueueCmd(void)
 
         case 106: //M106
         {
-          u8 i = 0;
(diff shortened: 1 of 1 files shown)


Xtrem69:
_I have the same problem with the encoder or the + / - buttons and it freezes my machine_

PR #863, opened 2020-07-23
UBL Specific features & menu

bryan065:
### Description

Adding support for Unified Bed Leveling, along with auto-detection for it as well.

### Benefits

- Added Activate / deactivate button to movement menu (`M420 Z` command) which also shows the current state.
- Added `M420` parameters
- Added popup prompt when UBL leveling is complete along with automatic smart fill
- Added menu to save/load mesh to slots
- Added Z Fade menu
- UBL menu is defaulting to auto-detect but can manually be overridden in `Configuration.h`.

### To-do

- Currently, auto-detection is experimental and relies on using `M503 S1` to detect `echo:; Unified Bed Leveling`. However, this only works with English. I'm not sure if other languages will be localized. Waiting on MarlinFirmware/Marlin#18734  so we can add ~bed~ probe type to M115.
- New strings were added but not localized yet


bryan065:
Whoops, thanks @digant73. Fixed.

bryan065:
@digant73 

No problem making those changes as it makes sense.

1. Fixed the missing .bmp files.
2. Can you double check if the ABL ON/OFF is working for you? I have `RESTORE_LEVEL_AFTER..` defined but that function doesn't seem to be working for me. I checked out https://github.com/MarlinFirmware/Marlin/issues/16306#issuecomment-568614948, and it doesn't seem to be working even on stock LCD for me. Although any `G29 A`, `G29 D`, or `M420 Z` command should update the button.
3. I haven't touched the stock ABL menu yet as I didn't know if we should roll up all the changes together in 1 PR or leave that for another one.
4. Makes sense, I just followed the layout of the probe offset menu as that menu had a save button too. A popup if changes are detected is more friendly. Although there isn't an easily accessible `M500` button anywhere, I was planning on adding it to an easier to reach place. 

As for the ProbeOffset menu, those improvements are greatly welcome. That was also on my todo list! (along with filament change, load/unload menu).

bryan065:
Updated UnifiedMove.c to poll for ABL status and update button accordingly. As pointed out by @digant73, sometimes leveling state can be changed by commands outside of G29 or M420 so it's better to poll for status after all.

digant73:
> @digant73
> 
> No problem making those changes as it makes sense.
> 
> 1. Fixed the missing .bmp files.
> 2. * Can you double check if the ABL ON/OFF is working for you? ~I have `RESTORE_LEVEL_AFTER..` defined but that function doesn't seem to be working for me. I checked out [MarlinFirmware/Marlin#16306 (comment)](https://github.com/MarlinFirmware/Marlin/issues/16306#issuecomment-568614948), and it doesn't seem to be working even on stock LCD for me.~
>    * I misread how the feature works. It's working as expected for me. The ABL ON/OFF button is correctly reflected when`RESTORE_LEVEL_AFTER..` is enabled. Need to test with it disabled.
>    * Although any `G29 A`, `G29 D`, or `M420 Z` command should update the button.
> 3. I haven't touched the stock ABL menu yet as I didn't know if we should roll up all the changes together in 1 PR or leave that for another one.
> 4. Makes sense, I just followed the layout of the probe offset menu as that menu had a save button too. A popup if changes are detected is more friendly. Although there isn't an easily accessible `M500` button anywhere, I was planning on adding it to an easier to reach place.
> 
> As for the ProbeOffset menu, those improvements are greatly welcome. That was also on my todo list! (along with filament change, load/unload menu, and UBL mesh validation, and mesh editing).
> 
> Edit: Also, please let me know if you have any suggestions/improvements on handling the ABL ON/OFF button
> 
> ~Edit2: Right now for the ProbeOffset.c changes, `probeStartZOffset()` will move X, Y, to the correct starting position. However, Z will move to the absolute 0 point but the displayed Z offset amount on the TFT is whatever has been last set by the user. If we start the Z probe point at the user's last value, it might cause issues. If the last value was bad, the nozzle might crash into the bed. I'm not sure how you'd want to handle this. Personally, I'd have a popup prompt to see if the user wants to start from 0 or their last used value or we can have a `#define` flag in configuration.h.~ Fixed.

2) yes, it works on my printer
3) yes, you can. The change is minimal. You simply aggregate the existing functions related to BLTouch test in a better organized menu + the new Z fade menu
4) restore my ProbeOffset.c file because the change you made is wrong. the move to Z0 guarante that the nozzle is moved to the current Z0 value associated to the current set Z Offset
5) For your ABL buttons I will define two new icons (so new icon name and text). I will consider something like "on" and "off" icons. The current eeprom_save icon in particular is totally wrong. It is more like a "on" or "checked" flag. In my improvements I replaced the eeprom_save icon with a more correct image (see my improvements attached below). Eventually, if you don't want to create new icons, you can simply duplicate the current "eeprom_save" and "stop" icons (you are using now) naming these as "on.bmp" and "off.bmp" and replace the eeprom_save icon with the one provided by me

Attached a zip file with some improvements I made in my FW. Here a description:
1) added 5th leveling point on manual leveling menu
2) added a "Connection" menu on Settings menu where I moved all typical connection features such as baudrate, disconnect button etc...)
3) added eeprom save and eeprom reset buttons on Machine menu.

For these new menus I created the icons for the TFT28 Unified Menu Material theme. They must be provided also for all the other TFTxx and also for the Classic Menu theme
For the new icons I defined those info in the following files:

Language.inc:
X_WORD (EEPROM_RESET)
X_WORD (EEPROM_RESET_INFO)
X_WORD (CONNECTION_SETTINGS)

Language_XX.h (below the value for Language_en.h):
    #define EN_EEPROM_RESET           "Reset"
    #define EN_EEPROM_RESET_INFO      "Reset EEPROM to stock printer settings?"
    #define EN_CONNECTION_SETTINGS    "Connection"

icon_list.inc:
X_ICON (EEPROM_RESET)
X_ICON (POINT_5)
X_ICON (CONNECTION_SETTINGS)

I'm also working on the filament change menu, so you can work on other features

[digant_PR.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/4977453/digant_PR.zip)

bryan065:
You're right about the ProbeOffset.c changes I made. My issue was actually fixed with 4816c76. Reverted

digant73:
> You're right about the ProbeOffset.c changes I made. My issue was actually fixed with [4816c76](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/4816c767343def25d51cfa0de1c60dc490132dbe). Reverted.
> 
> Also, I think I'll leave the connection menu changes for another PR as it's sort of outside the scope of this one.

no problem. Probably I will open a PR with all of my changes, so it'is better that you don't provide my files in your PR. Just to avoid conflicts (I'm changing something also in the ProbeOffset.c)

bryan065:
Ah, I'm finishing off the polishing and updating the icons right now.

### To-Do
- Update all the new icons for the classic theme
- Update the ABL ON/OFF icons
- Update the nozzle dec/inc to nozzle down/up to be more user friendly as per #869 
- Update the bilinear menu to be more consistent with the new UBL menu.

After that, I think it's okay for now and wait for this to get merged

labotecno:
great !!! you make it good !

ETE-Design:
@bryan065 Can you pls. Merge this Up with newest branch? Would really like to try it out on my MKS TFT28 :-)

ETE-Design:
@digant73 The filament change feature "M600" your working on, can you make it so you can choose if it is connected to the screen or the board?

radek8:
> @ bryan065 Dokážete pls. Sloučit to s nejnovější pobočkou? Rád bych to vyzkoušel na MKS TFT28 :-)

Of the MKS TFT’s, only the TFT32 1.4 is supported:

Issue #854, opened 2020-07-21
[BUG] E0 never re-selected after chaning to another extruder.


VE5BNC:
I running a printer with two extruder.   When manually extruding,  

### Steps to reproduce

1. Choose the Extrude menu
2. Pres Extrude. E0 will extrude filament
3. Choose the second nozzle
4. Press Extrude.  E1 will extrude filament.
5. Choose the first nozzle.
6. Press Extrude.  E1 will extrude instead of E0

**Expected behavior**
The screen should send a T# GCODE each time it sends a manual extrude command.

**Actual behavior**
The screen sends a T# for all extruder except when E0 is selected.

### Hardware Variant
TFT35 V3 and TFT35 V2

### TFT Firmware Version & Main Board Firmware details
Freshly compiled version as of July 20 2020. 3.0.26 and 2.0.26

### Additional Information
I verified that it does not send a T0 by monitoring the serial communication between the board and the scren.

I was able to correct the behaviour by editing BIGTREETECH-TouchScreenFirmware-master\TFT\src\User\Menu\Extrude.c and commenting out line 133 as shown below.

129	    if(extrudeCoordinate != eTemp)
130	    {
131	      extrudeCoordinate = eTemp;
132	      extrudeCoordinateReDraw(true);
133	//      if(item_extruder_i != heatGetCurrentToolNozzle() - NOZZLE0)
134         storeCmd("%s\n", tool_change[item_extruder_i]);
135       storeCmd("G0 E%.5f F%d\n", extrudeCoordinate, infoSettings.ext_speed[item_speed_i]);
136     }

bigtreetech:
Confirmed it's a bug of SINGLENOZZLE.  This happens when there are multiple extruders with only one hotend.  will fixed later.

VE5BNC:
Thanks

Issue #515, opened 2020-04-02
[Q] Add Chamber Temperature at Unified Menu


tpipits:
I would like to know how can i modify the code so the chamber temperature can appear to main/print screen at unified menu and at the top right corner at the status bar. I use TFT24 v1.1.

I have trigoriila v1.1 board (based on ramps 14) with marlin 2.0.5.2 firmware and i have added a chamber temperature sensor. At marlin simulation screen the chamber temp appear fine and i want to add it and to touch mode. 

I have no problem to be combined with other icon (for example the bed temp icon to change between bed temp and chamber temp.

I am waiting for your help.
Thank you

tpipits:
After 4 days that have been past i couldn't find any solution on how to display chamber temp and fan to 2.5.3 firmware for tft 24v1.1 screen.

@Developers could you change the issue's label to enhancement and add this function as feature update?

Some printers have a heated or not heated chamber with a fan and a temperature thermistor and marlin can support these functions out of the box (at marlin simulator screen champer's temp and fan appear and function very well but when i use touch mode i can't see the chambers temp and i can't see status and make use of the fan).

Thank you.

bigtreetech:
chamber feature is merged in #875， so close here, if you have some else issue, please reopen here and contact me

PR #875, opened 2020-07-28
clean up & improve heater

Msq001:
### Description
* fix M109/190 index when gcode with in "N *" check bytes
* clean up heater, fix Multiple extruder one hotend bug #854
* add chamber heater support #515 
* improve top temperature status display
* delete the built-in end gcode in the program. After printing, only the end gcode configured in the config.ini file will be sent if enabled.
* Auto-detect the number of hotend and whether bed/chamber enabled  by the response of M105 from Marlin,


Msq001:
> in the Heat menu you removed limitValue when assigning a new value.
> Is that a desired behavior?
hello.
limitValue has be called in `void heatSetTargetTemp(uint8_t index, int16_t temp)
`

digant73:
> > in the Heat menu you removed limitValue when assigning a new value.
> > Is that a desired behavior?
> > hello.
> > limitValue has be called in `void heatSetTargetTemp(uint8_t index, int16_t temp) `

ok.
You forgot to include in config.ini the last added slovenian language and also to align the file in the 3 directory where it has to be copied

Attached you can find your better ini file with the addition of the slovenian language (with the correct indentation in the description section)

[config.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/4989017/config.zip)

PR #877, opened 2020-07-28
Fix point 2 of leveling

Guilouz:
Fix machine size min for second point.


radek8:
Why did you close it?

Guilouz:
> Why did you close it?

Bad handling, sorry...

carl1961:
seems this sure should have been merged

Issue #822, opened 2020-07-04
[FR] (feature request title)100mm move on the buttons


MontyFlange:
Any chance we could have movement of 100mm . Unloading a bowden at 10mm a go is boring. Also on a 400mm Z getting it to the top takes ages. 

Thanks

NicoletoAerials:
Hi, I am new here and do not know how to post the enhancement. I made some changes in a couple of files, and I have it working for the “move” and “extrude” menus.

MontyFlange:
Ohh nice can you send it to me ?

MontagueFlange@gmail.com

oldman4U:
Hi.

This is implemented in the latest version of the firmware. Have you been able to install it? If not please let us know how we can help. If yes, please close this ticket.

Thank you

oldman4U:
@MontyFlange 

Close this ticket

MontyFlange:
Thanks

PR #879, opened 2020-07-29
Add 100MM movement also 100MM and 200MM extrusion.

NicoletoAerials:
### Requirements

Add. the option to move up to 100mm with one touch
Add the option to extrude 100mm and 200mm of filament (Upload and Unload)

### Description

Added the option to select 100mm and 200mm of filament extrusion (Load and unload)
Added the option to perform 100m increment movements.
Included the corresponding icons for the unified menus

### Benefits

Helps to perform fast filament changes when trying to do 2 color prints. Also helps to perform longer carriage movements (100m increments)

### Related Issues

Will solve enhancement #822


NicoletoAerials:
> The default max extrusion length in Marlin firmware is 200mm.

You are right. In such case I'll change the 400mm extrusion for 200mm extrusion in order to avoid conflicts with the default max extrusion in Marlin firmware.

Issue #920, opened 2020-08-06
[BUG] ? Stuck on Black screen when switched to Marlin mode


charakaSamawry:
Description
Latest firmware from last night (master) and when switching from touchscreen mode to marlin mode it as usual goes black screen but the encoder whee doesnt 'bring it to life' now and restarting the printer, it stays stuck on the black screen, though i thought it should reboot into touchscreen mode as default.. (touchscreen mode works fine, but no mesh edit like on marlin mode) 

Steps to reproduce
1. Turn on touchscreen with SD card update
2. Change to marlin mode
3. Move encoder wheel

**Expected behavior**
Switches to marlin mode

**Actual behavior**
stuck on Black screen.

Hardware Variant
TFT35 V3 E3 with SKR mini v2 E3

TFT Firmware Version & Main Board Firmware details
Both flashed last night, master and bugfix.

Additional Information
https://github.com/charakaSamawry/BIGTREETECH-TouchScreenFirmware/tree/sam1

-https://github.com/charakaSamawry/BIGTREETECH-TouchScreenFirmware/blob/sam1/Copy%20to%20SD%20Card%20root%20directory%20to%20update%20-%20Unified%20Menu%20Material%20theme/config.ini

https://github.com/charakaSamawry/Marlin/tree/sam1

-https://github.com/charakaSamawry/Marlin/blob/sam1/Marlin/Configuration.h
-https://github.com/charakaSamawry/Marlin/blob/sam1/Marlin/Configuration_adv.h

traffic-light:
It could be a problem with my code in that latest commit.
I'm searching for a solution.

traffic-light:
That's why I din't noticed it.. 
It doesn't work in combination with #910 OR #912

traffic-light:
Could it be that you have activated the wrong version? instead of 12864 mode the 2004 mode?

charakaSamawry:
Sorry dont think you had commited your change when i found the bug, but have tried your fork anyway to no avail though sadly. Tried with default config as well incase id changed something weird... (and without custom theme just in case)
Problem still happening, switch to marlin mode black screen doesnt go away with move of the encoder wheel and stays black but lit even when restarted. Only fixed by a reflash. Reset button seems to make the back light flicker but nothing else..

I wasnt sure where you changed the settings you mentioned regarding 12864 and 2004 mode..?

traffic-light:
I think you've got an older config.ini.
[...]

Issue #808, opened 2020-06-26
[FR] (knob_led_color idle delay)


devildant:
**Is your feature request related to a problem? Please describe.**
when the idle delay of the screen activates the knob led stays on.

**Describe the solution you'd like**
I would like there to be an option so that when the screen goes to idle, the knob leds goes off, and it turns back on with the same color when the screen comes out of idle

**Describe alternatives you've considered**

**Additional context**

traffic-light:
Oeh. That sounds nice. Currently I'm working on it. When I'm ready (or something else maybe) with it I'll post an update.

traffic-light:
At the moment i'm planning to make it work together with the LCD screen. When the LCD go's idle the knob's LED will go out (or will dim if possible. (it should be possible because it's a neopixel)

### Updates

- I've added the menu item and settings management. ![photo_2020-08-05_13-45-44](https://user-images.githubusercontent.com/37051137/89409260-120e2c00-d722-11ea-9d25-4e91e5108d87.jpg)

To do:

- [x] Add menu item to the featuremenu (toggle idle mode)
- [ ] Make the button go off when the screen go's in idle
- [ ] Make it possible to set a idle precentage

devildant:
nice, it's progressing well :)

traffic-light:
It's almost done. The only thing that I can't get working is the settings.ini
It's telling me over and over again that ''''knob_led_idle'''' keyword is not found...
So.. I'll have to dig a little deeper for that one.

Maby someone around here knows where I have to add new keywords? 
I've placed it in *flashstorage.c*
````

  wordToByte(infoSettings.knob_led_idle,              data + (index += 4))
````

traffic-light:
I've made a pull request. Unfortunatly it's not compiling on **BIGTREE_TFT35_V3_0**

If you wan't to test it, see #914 
(or my fork https://github.com/traffic-light/BIGTREETECH-TouchScreenFirmware)

devildant:
Thx @traffic-light I will test it as soon as possible (I saw that it was merged, nice)

devildant:
hello, i test, 
this crashes the LCD emulation :(
![image](https://user-images.githubusercontent.com/8700106/89559042-60502780-d815-11ea-8b93-75975c7d582f.png)

traffic-light:
Oh... So my code do's that? I'm sorry. I'll try to fix it

traffic-light:
[...]

PR #924, opened 2020-08-06
Added brightness level 0% and fixed the bug in knob_led_idle

### LINKED ISSUE #918: [BUG] Update font and icons fails
Updating font's and icons fails. Font's stops around image 43 and font's stops around 37% with the word_unicode.fon

### LINKED ISSUE #922: Dimmer
Please add the possibility to dim the screen down to 1% and or even 0%. People like me that have the printer in the bedroom would love the possibility to turn it off during the print.

Of course to implement also a filament load and unload button like said in a previous post would be also great. 

best regards

traffic-light:
### brightness level 0%
**With this update the brightness level can be set to 0%. This resolves #922** 
You can set the brightness_idle to 0%. When you change the normal brightness it skips the 0%.
When someone sets 0% for the normal brightness in the config.ini it will be set to 5% instead (at startup).

### knob_led_idle bug in marlin
With this update the scrambled screen when the knob_led_idle starts to kick in is fixed.
The IRQ is replaced to solve this. (you want to switch of the IRQ as short as posible)
(there is no issue mentioned for this. It was mentioned in an older isue (#808 ))

### Other things
This pull request also resolves #918 
It could be possible that #920 will be resolved too. But I'm not sure about that.

### Note:
Can someone please verify for me that the problem/ bug mentioned above is gone?
On my screen it's working flawlessly


charakaSamawry:
#920 not fixed with your fork. Tried a default config.ini in tft35 firmware and default skr mini e3 marlin configuration.adh but problem persists. Will try going back through commits till it fixes itself tomorrow...

Msq001:
Hello, #918 should be fixed in #925

traffic-light:
@Msq001 I did notice it later on. The strange thing is that it did work after my changes one time. When I did it another time it didn't work.

Thanks for the update

PR #925, opened 2020-08-07
fix hang when updating

Msq001:
### Requirements

### Description
the `XPT2046_Init()` and `knob_LED_Init()` function is moved before` OS_TimerInitMs()`
Because touch and knob LED will be called in the timer interrupt, so the initialization must be completed before calling

### Related Issues
#918


traffic-light:
With this adjustment (knob_led_init) the led of the knob is not on at startup..

PR #926, opened 2020-08-07
Add auto FW retract parameter (M209) & more

bryan065:
### Description

- Added parameter option for "Set Auto Retract" (`M209`) feature in marlin.
- Fixed `M420 Z` parameter as it only takes INT and not FLOAT (easier to punch in the value on the numpad)
- Added `M503 S0` checks when navigating parameter menu in case any parameters are changed via terminal/USB/ESP3D
- Added save EEPROM settings button in case any settings were changed via terminal/USB/ESP3D
- Added configuration option for the new EEPROM save/load/reset buttons in the machine settings menu

### To-Do

- [ ] Add config.ini option for `FRIENDLY_PROBE_OFFSET_LANGUAGE`
- [ ] Add config.ini option for new `QUICK_EEPROM_BUTTON`


TFT/src/User/Configuration.h
```diff
@@ -230,14 +230,25 @@
...
+/**
+ * Enable quick EEPROM save/load/reset button
+ * 
+ * This will enable a handy button in the (settings > machine) menu
+ * 
+ * Note: if disabled, EEPROM operations can also be accessed in the (settings > machine > parameters) menu
+ * 
+ * Options:  0: Disabled    1: Enabled
+ */
+#define QUICK_EEPROM_BUTTON 0
```
guruathwal, line 250: This should be left default ON. Although I will recommend using the EEPROM flag from M115 command to determine the visibility of this option.

bryan065: Whoops, this was supposed to be left on but I guess I left it at 0 during trsting

PR #929, opened 2020-08-07
update language_du and some more

### LINKED ISSUE #928: [BUG] Marlin scrambled when knob_led_idle is active
When you're in marlin the screen get scrambled when knob_led_idle is activated.

I solved this with the next adjustment:
In the file ledcolor.c:

Moving    __disable_irq();
from line **69** to **76**

And __enable_irq();
from line **89** to **85**

This adjustment works great. For both, the neopixel led and the marlin mode.
(the color of the neopixel do not change and the pixels are controlled the rightway)

@bigtreetech did remove this change.

(want to know why and if it's possible to bring back.)

Problem:
image

(this is solved by moving the disable and enable _irq(); to the new locations)

traffic-light:
+ i've updated the language_du.h file (don't know why someone in the Netherlands would prefer it but oh well..
+ I changed steps per MM to steps per mm in all languages. MM is NOT milimeter
+ Changed MM to mm on some other places
+ changed led to LED (my fault with knob_led_idle...)
+ Changed DEFAULT_LCD- BRIGHTNESS & IDLE_BRIGHTNESS. With 0% as new percentage this value was too low
+ fixed the issue that the knob led doesn't turn on after a change in #925


bigtreetech:
Hello, About the the problem of Neopixel Led and the scrambled of Marlin mode problem, I think #931 will handle it better. Can you test it? If ok, please synchronize your Pull request

traffic-light:
I will do that m thank you.

Currently my printer is printing. When that's done I will test.
If it works i'll sync it

traffic-light:
@bigtreetech thanks. That works good.
I'll change the name of this pull request.

I removed the part with my adjustments

GregSKR:
> MM is NOT milimeter

I agree.
I suggest to change it also in PrintingMenu.c line 271:
`my_sprintf(tempstr, "%.2fmm",coordinateGetAxisTarget(Z_AXIS));`

traffic-light:
@GregSKR  Done.

Issue #921, opened 2020-08-06
G29 and other Gcodes in top of Marlin mode


RicardoEmePe:
### Description

When I switch to Marlin mode, at the windows  top some Gcodes are displayed. I had before firm. TFT35_V3.0.25.1 and  it works well !!!
![IMG_7382](https://user-images.githubusercontent.com/37264449/89563955-904ef900-d81c-11ea-9f6d-b4d17335d787.jpg)

### Steps to reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior**
No Gcodes in top. 

**Actual behavior**
Read description bug

### Hardware Variant
BIGTREETECH_TFT35_V3.0
### TFT Firmware Version & Main Board Firmware details
TFT35_V3.0.26.x jul 13 2020  // SKR V1.3

### Additional Information
* Include a ZIP file containing your `Configuration.h` or use [Pastebin](https://pastebin.com/) and paste a link in this issue.
* Provide pictures or links to videos that clearly demonstrate the issue.

guruathwal:
you need to reflash the config.ini file

RicardoEmePe:
Ok, now it's Ok. Yesterday eveningh I did it, but no changes. Now, Ok and "Invalid Keyword" message while reading config.ini don't show " Lcd12864_title: LCD12864 simulator" , and "lcd2004_title: LCD2004 simulator".
Anyway, while config.ini reading, the message in "Invalid Keyword" shows: 
ack_popup_type:1
marlin_type:1
bed_en:1
chamber_en:0
pl_recovery_en:1
knob_led_idle:1
lcd_brightness:11

I don't know if this is important. 

Thanks, guruathwal, for your quick reply !!!

guruathwal:
that's because you are using a new config file with old firmware

traffic-light:
@RicardoEmePe you can close this one when it's working again

RicardoEmePe:
I come back home next Sunday, I’ll try and close if ok. Many thanks 

Enviado desde mi iPhone

RicardoEmePe:
New firmware and new config file. Same problem. Enclosed you have a film with the  issues

[https://drive.google.com/file/d/1ProUmeQCrPmC9IbuyJh4kZ1BtTsY6WxV/view?usp=sharing]

traffic-light:
We don't have permission to see the file

PR #899, opened 2020-08-02
improved menus + new menus + new popup type + custom ACK message handling + cleanup and fixes

digant73:
Improved existing menus such as:
* ProbeOffset: added new functions (save, XY axis unlock...) via a new submenu + cleanup
* ManualLeveling: Added new functions (XY axis unlock, bed edge distance menu...) via a new submenu + cleanup
* MachineSetting: Added new functions (save, restore, reset) via a new submenu
* ZFade: added save button (saving feature is now the same as in ProbeOffset and Babystep menus)
* UnifiedMenu: improved ABL ON/OFF update + cleanup

New menus:
* ConnectionSetting: Any connection settings such as baudrate, disconnect button, shutdown button were migrated AS IS in this new menu located under Settings menu
* Tuning: it contains features such as PID autotune menu and typical 100mm extrusion button (usefull for calculating proper steps-per-unit). The menu is located under Settings menu
* PID autotune: menu allowing to handle the entire PID process for bed + nozzles. A dialog box will popup the result of the process (success, fail, timeout)
* LevelingEdgeDistance: it allows to set the "level_edge" parameter. Thus to set the bed edge distance of the four leveling points

New features:
* added new "notification" popup type. A notification popup is displayed for few seconds (by default set to 1.5 seconds, configurable in Configuration.h). No user interaction is needed
* added a configurable popup type and buzzer type for the "known" echo messages (registered in parseACK.c).
    
    Available popup types are:
      none:         no popup. The echo message is silently discarded
      reminder:     reminder popup. A reminder popup is displayed. A user interaction is needed
      notification: notification popup. A notification popup is displayed for few seconds. No user interaction is needed
    
    Furthermore, the use of buzzer (on or off) is also selectable

    Functions signature to read and set the parameters for any known echo message are available in parseACK.h.
    E.g. I used that in Z Fade menu to temporary disable some echo messages

* added a configurable popup type and buzzer type for the error ACK and echo ACK messages.
    
    Available settings, provided in config. ini and FeatureSetting menu, are:
      OFF:   no popup. The message is silently discarded
      ON:    A reminder popup is displayed. A user interaction is needed
      SMART: A notification popup is displayed for few seconds. No user interaction is needed

    That feature is useful for example to temporary suppress any kind of error or echo ACK message

Fixes:
* general cleanup (converted tabs to spaces + correct indentation) on different files
* restored slovenian language in config.ini file (it was removed by mistake in the last BTT merge)
[...]

PR #939, opened 2020-08-07
Clean up and remove duplicate codes

guruathwal:
- Rename **bed_en** to **heated_bed**, rename **chamber_en** to **heated_chamber**.
- Remove duplicate code & unused icons.
- Prevent error messages from closing in smart notifications to allow user interaction.
- Restore main menu icon layout


guruathwal:
@digant73  Hi, the best way to do that is to detect the expected printer response and display it and ignore the extra responses at the time of processing them in **parseACK.c**. Blocking new messages will also block many other messages which may trigger while the earlier message is open.
Also, the EEPROM buttons are already available in parameter settings & custom menu, This need to be sorted out as well because we have very limited program memory and flash storage.

digant73:
@guruathwal yes, I know neither the first nor the last message approach is the best solution. At the end, based on all the tests I performed, for the notification I preferred to display the first message (it provided the most important info). Maybe a configurable (at compilation time) parameter in configuration.h to choose the first or last message approch could be the best solution. Anyone will use the his preferred approch.
For the EEPROM I preferred to put that in the menu because is more intuitive, self explaining and easy to use. The restore and reset function on parameter menu could be misinterpreted as applied only to parameters and not to all system. I never used the custom menu.

Msq001:
@digant73 
Hello, Can you confirm that your freeze problem can be solved in here https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/950 ?

guruathwal:
the freeze on preheat menu must be due to #899 . as new settings variables were added but the  **PARA_SIGN** in flashStore.c was not changed. reflashing the config.ini should resolve the freeze.

guruathwal:
#921  is a similar issue with custom gcodes showing in marlin mode

digant73:
@Msq001 it seems it solved the problem. Just another feedback I received from some people. They said that there were (before your fix) freezes in the statusScreen menu when text displayed on the status bar needed scrolling. Could it be an issue to investigate?

@guruathwal I always install from scratch, so probably the problem was somewhere else. However, with the last main branch I'm not having any freeze

Issue #933, opened 2020-08-07
[BUG] Marlin mode not working on stm32f10x due to wrong queue usage


digant73:
Marlin mode is no more working (black screen) for stm32f10x.
The problem is caused by the use of the wrong queue in MarlinMode.c.
On stm32f10x/spi_slave.c file the used queue is SPISlave.data while the used queue in MarlinMode.c was marlinQueue.data

I temporary fixed the bug with the following code in MarlinMode.c but a bugfix must be applied at the root of the problem

/*    while(marlinQueue.index_r != marlinQueue.index_w)
    {
      marlinParse(marlinQueue.data[marlinQueue.index_r]);
      marlinQueue.index_r = (marlinQueue.index_r + 1) % QUEUE_MAX_BYTE;
    }*/

    while(SPISlave.rIndex != SPISlave.wIndex)
    {
      marlinParse(SPISlave.data[SPISlave.rIndex]);
      SPISlave.rIndex = (SPISlave.rIndex + 1) % SPI_SLAVE_MAX;
    }

As I suggestion it's better all the API and its data structures (such the queue) for SPI slave remains confined in the spi_slave.c file.
Maybe a new function to retrive the pointer to the queue can be added in each spi_slave.c/h files and used on MarlinMode.c

### Steps to reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior**

**Actual behavior**

### Additional Information
* Include a ZIP file containing your `Configuration.h` or use [Pastebin](https://pastebin.com/) and paste a link in this issue.
* Provide pictures or links to videos that clearly demonstrate the issue.

digant73:
correlated bug #920

PR #942, opened 2020-08-08
fixed Marlin mode for STM32F10x (bug #933) + cleanup

digant73:
BUG FIXES:
* Fixed bug #933. Marlin mode was not working on STM32F10x due to wrong queue usage in MarlinMode.c
   Now the queue for Marlin mode and new HD44780 simulator are fully handled on SPI_slave.c (as it was prior the the last merge) and HD44780.c for both STM32F2_F4xx and STM3210x. MarlinMode.c simply invokes a function on SPI_slave or HD44780.c to retrieve the data from the proper queue

IMPROVEMENTS:
* BabyStep.c, MachineSettings and ZFade menu: "save icon" and related save function are not displayed/available in case EEPROM support is not provided by the underlying Marlin FW

CLEANUP:
 * FeatureSettings.c, popup.c and Selectmode.c: converted tabs and adjusted indentation


bigtreetech:
Hello, I have tested the screen of BTT TFT series. The LCD 2004 / 12864 is in normal working state.
Since 2004 and 12864 can only exist one at a time, I hope to use the same memory (queue) to process 2004 / 12864 in order to keep the program concise.
Does the problem of #933 only exist in MKS TFT?

Msq001:
Hello, #951 should fixed Marlin mode problem,  please confirm. If it's ok, please synchronize your PR.
Thank you very much.

digant73:
hi @bigtreetech, @Msq001
Yes, I agree to use only one queue, but it should be better to perfecly decouple MarlinMode from SPI_slave and HD44780.
In this moment we have HD44780 and SPI_slave depending eacho other and also depending on MarlinMode (for the queue). Furthermore HD44780 depends on SPI_slave for the interrupt handler. Plus all the hd44780 code present in the interrupt handler (in spi_svalde.c) should be moved totally on HD44780.c as I did in my code).
Also the queue handling can be managed in spi_slave and hd44780 (as I did providing the xx_getData function.
spi_slave and HD44780 are API, so in my opinion all the implementation must be confined on there.
spi_slave  is a general API that can be used not only for Marlin mode. With the current implementation spi_slave can be used only for Marlin (the queue is instanciated in MarlinMode.c).
If you agree, I can simply define a new file e.g. called CircularQueue.h under HAL directory and spi_slave.c and HD44780.c will simply include that file (instead of MarlinMode.h). On spi_slave.c and HD44780.c I will provide a function to set the queue that will be internally used. It will be responsibility of the utilizer (e.g. MarlinMode) to pass the queue pointer to spi_slave or HD44780 and to deallocate the queue (so the current implementation on MarlinMode remains the same).
With that approach spi_slave, HD44780 API and MarlinMode are perfectly decoupled. I can continue to use spi_slave even for other scopes.
If you agree I can proceed with that adjustement. Please let me know

bigtreetech:
I agree with you, Thank you for your work!

digant73:
> I agree with you, Thank you for your work!

Thanks. I'm going to make the changes

digant73:
> I agree with you, Thank you for your work!

Ok, I merged my improvements on spi_slave, HD44780 and MarlinMode.
Now a pointer to circular queue must be provided in the xxInit (for spi_slave) or xxConfig (for HD44780) functions. The circular queue is defined under /HAL/CircularQueue.h. Just a little bit change was made in the buffer type. "data" field is now an array of uint8_t and not a pointer to uint8_t. So now in MarlinMode.c , it is enough to simply instanciate the circular queue object and passing its reference to spi_slave or HD44780. In spi_slave and HD44780, I always intialize the queue indexes (index_r, index_w and count) before using the queue. I suspect that in the old version the queue indexes could be used with dirty values.
I also made cleanup on spi_slave, HD44780 and MarlinMode, the usual tabs removal, comments adjustment and aligned functions (it allows a better comparison with tools such as "beyond compare" etc...).
For me the PR could be merged (after your testing).
thanks

PR #961, opened 2020-08-11
Add non-blocking toast notifications, Add a separate sound control menu, Add SPI flash usage detail in info screen

guruathwal:
- Add non-blocking toast notifications
  - Show toast notification in the title bar area.
  - Multiple toast messages  (up to 5) are enqueued to make sure all notifications are displayed.
- Add SPI flash usage detail in info screen.
- Clean flash addresses for fonts and icons.
- Add a separate sound control menu.
- Use Numpad to edit edge distance and fade height
- Reduce UBL code.


digant73:
@guruathwal Each "\n" character on a label value (e.g. "Save printer settings\nto EEPROM?") is translated to " " (space character) in my TFT, while it should force a new line. That could be also something to fix.
You also need to align ZFade.c with last merged PR (where I check  EEPROM is supported by Marlin)

GregSKR:
@guruathwal: I see you removed the space characters before the question marks in the french section.
For your information (I'm french) the question marks (and the exclamation marks) needs a space before in french language...
https://fr.wikipedia.org/wiki/Point_d%27interrogation

guruathwal:
@digant73  if the TFT displays the text in windows in a single line if the text width fits window width so that the line is centred in the window. For multi-line text which can not fit in one line, this starts from the top-left corner. 
This PR and all the files are rebased to the latest commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/c75f9f316ec14e26b44e862be34fa290d2a6f947
I will add back the ignore code. 
About the `QUICK_EEPROM_BUTTON`, I have no idea about it as I did not find any reference in previous PRs. Although the EEPROM settings are already available at two places which are already been used by the users. 
I would like to ask if the `100 mm` extrude button is needed when it can also be done through the Extrude menu, where the required length can be controlled by the user. 

@GregSKR  Thanks. My fault, I did not know about that. I will revert it in next commit.

digant73:
@guruathwal
Ok thanks, so the space character is not a bug but it's the correct implementation.
The QUICK_EEPROM_BUTTON was added by this PR https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/926#event-3641198832 just merged in main branch today (the same you are working now).
But also my PR was merged today and my changes on MachineSettings (not including the changes made by bryan065) are now the current implementation. I attached above my changes including the support to QUICK_EEPROM_BUTTON made by bryan065 in his PR. You can simply add my version attached above to your PR. Also it should be better you set QUICK_EEPROM_BUTTON to 1 instead of 0 in configuration.h (that was also your suggestion to bryan065 in his PR).

digant73:
@guruathwal I also find a bug in the print menu. Although a file name can be 255 chars long, the buffer used  in the dialog box asking if you want to proceed with the print is only 75 chars long. Many people referred that they always have a freeze when selecting a file name even 60 chars long.
The problem is on line 310 on print.c where the following buffer is allocated:

            char temp_info[75];

Issue #966, opened 2020-08-12
[BUG] Pause does not park TFT50


Rebel3D-CZ:
The pause is not parked, or only sometimes at defined XY coordinates. Printing stops and retracts. After selecting "Resume", deretration is performed and printing continues. I use it for a filament sensor connected to a TFT50.

I'm using the current compiled version from github.
Board FYSETC S6
FW Marlin 2.0.6

The ideal solution would be to implement M600 (ADVANCED PAUSE FEATURE), or add similar / same options to FW TFT https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/965

sorry for english, i use translator :)

bigtreetech:
Hello, It has supported M601 for gcode pause after #764 
And after #971 The default Settings included M600/M601
They are consistent with the action of manual pause(click pause button in touch screen)
Can this solve your needs?

Rebel3D-CZ:
582/5000
I don't understand it or it's working badly.

1. I compiled the latest version from github
2. in Marlin FW I have M600 enabled = verified from 1284 LCD. It stops, parks at the positions specified in the Marlin FW, ejects the thread and waits for insertion, and continues by pressing. Then it returns to the print position and continues. It's alright.
3. Press manual pause on the TFT50. Printing stops and nothing happens. After pressing the "Continue" button, printing continues. It is bad.
4. When I enter and send the M600 on the TFT50 LCD terminal, printing stops and nothing happens. This is also wrong, so the M600 does not send to the Marlin FW?

bigtreetech:
If the touch screen send M600 to Marlin for pause, Marlin must Enabled `EMERGENCY_PARSER` , `HOST_ACTION_COMMANDS` and `HOST_PROMPT_SUPPORT`, and the touch screen needs to send M876 to Marlin for resume, otherwise the touch screen will lose communication with Marlin, Because Marlin stops responding the normal gcode only responds M876 (when enabled `EMERGENCY_PARSER` )  in the M600 Park state.
And the analysis of the interaction of HOST_ACTION_COMMANDS has not been implemented now, So the current touch screen dont send M600 to Marlin, instead, touch screen will send manual pause command to marlin for pause, if the config.ini is seted in the figure below
![image](https://user-images.githubusercontent.com/38851044/90203305-e7ebf680-de12-11ea-935f-c542679c0d57.png)
the pause gcode is
```
<Set absolute coordinate mode>
[...]

Issue #965, opened 2020-08-12
[FR] M600 support in Unified menus


jariivanoff:
**Is your feature request related to a problem? Please describe.**
I like to print with different colors, and change filament at a certain layerheight. I use to make that in slicer, and it sends an M600 command. Now that I have installed an TFT35 v2.0 I have no way to continue the print since there is no support for that in the menus and the TFT2.0 have no Marlin mode which I could switch over to and make the print continue after filament change.

**Describe the solution you'd like**
A button that I can press to make the print continue after a M600

**Describe alternatives you've considered**
Getting an other display with Marlin support or go back to my LCD12864

**Additional context**

bigtreetech:
Hello, It has supported M601 for gcode pause after #764 
And after #971 The default Settings included M600/M601
They are consistent with the action of manual pause(click pause button in touch screen)
Can this solve your needs?

ETE-Design:
@bigtreetech Is it possible to use this feature if my sensor is connected to the mainboard and not the screen? End where in firmware do I define the pin so it know It's connected to the mainboard?

jariivanoff:
Well.. the thing is that if you use a slicers feature of M600 the board won't know that the printer has paused.. if it  comes with the gcode so to say... and besides that.. I haven't found the pausfunction yet.. but that is perhaps since I always print from Octoprint and never use the display for pausing...

bigtreetech:
> Well.. the thing is that if you use a slicers feature of M600 the board won't know that the printer has paused.. if it comes with the gcode so to say... and besides that.. I haven't found the pausfunction yet.. but that is perhaps since I always print from Octoprint and never use the display for pausing...

@jariivanoff Hello, Please come #966  to discuss it

bigtreetech:
> @bigtreetech Is it possible to use this feature if my sensor is connected to the mainboard and not the screen? End where in firmware do I define the pin so it know It's connected to the mainboard?

Hello, it also need support for HOST_ACTION_COMMANDS and HOST_PROMPT_SUPPORT in Marlin and Touch Screen, TouchScreen has not  implement it now.
 Please come #966 to discuss it

PR #971, opened 2020-08-13
default support M601 gcode pause, add M600 same as M601

Msq001:
### Requirements

### Description
* default support M601 gcode pause, add M600 same as M601 
* restore font color when leave info menu, otherwise font color is Gray

### Related Issues
#965 
#966


ETE-Design:
@Msq001 How does it Work? Can er use filament runout connected to out mainboard? And what features need to be enabled in Marlin other that Filament Runout?

PR #972, opened 2020-08-13
Correction in language and icons.

AntoszHUN:
The 2 newly added options have been translated into the language_hu.h file.

I deleted the unnecessary icons from the Hybrid Red Menu ZIP directory for each display.

Hello, AntoszHUN


AntoszHUN:
I didn't know that was how the spelling happened. Thanks for the fix!

PR #973, opened 2020-08-13
Fix extrude label

Guilouz:
Fix label to use 'Extrude' translation for filament extrusion.


thisiskeithb:
I'm not sure this is a bug since you want it to say "Load" instead of "Extrude" in this case.

radek8:
This is not a good idea

Guilouz:
Ok no problem.

PR #976, opened 2020-08-14
LCD Encoder configurable delay

digant73:
IMPROVEMENTS:
* LCD_Encoder.c:  On different main boards, the hard coded delay (8 us) used in LCD_Encoder.c is not enough. The result is no scrolling in Marlin menu or a backward sliding. Now a configurable delay named LCD_ENCODER_DELAY is present in Configuration.h.


traffic-light:
Good work cleaning some files. It's something that's not really noticable for the end user, but it is also verry important.
It will be a mess without doing that.

> It's a little off topic. But oh well:
> I don't want to push you or look like an unsure looking for attention guy or something. But I've got the feeling that you know the structure of the btt firmware pretty good. Can you take a look at my new PR and give me some feedback? It's my first big PR. So, if you got some time:  #982 ;-))

PR #977, opened 2020-08-14
move tuning menu to machine setting and simplify eeprom buttons, Fix fan speed display issue

### LINKED ISSUE #883: F0 value set in Marlin mode is not updated in Touch mode.
The part fan (F0) value set in Marlin mode does not get sent back to Touch mode when switching.
(However, F0 value set in Touch mode does get successfully transferred to Marlin mode.)

guruathwal:
- Move the tuning menu to machine settings
- Simplify EEPROM buttons.
- Fix fan speed display issue (#883)

Fixes #883

### DIFF
--- a/TFT/src/User/API/FanControl.c
+++ b/TFT/src/User/API/FanControl.c
@@ -13,6 +13,7 @@ static bool    fan_send_waiting[MAX_FAN_COUNT] = {false};
 void fanSetSpeed(uint8_t i, uint8_t speed)
 {
   fanSpeed[i] = speed;
+  fanSpeedPercent[i] = (fanSpeed[i] * 100) / infoSettings.fan_max[i];
 }
 
 uint8_t fanGetSpeed(uint8_t i)
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -265,7 +265,7 @@
  *
  * Options:  0: Disabled    1: Enabled
  */
-#define QUICK_EEPROM_BUTTON 0
+#define QUICK_EEPROM_BUTTON 1
 
 /**
  * Manual Leveling
--- a/TFT/src/User/Menu/Fan.c
+++ b/TFT/src/User/Menu/Fan.c
@@ -65,94 +65,92 @@ void menuFan(void)
     KEY_VALUES key_num = menuKeyGetValue();
     uint8_t actFan = fanGetSpeed(curIndex);
     uint8_t actFanPercent = fanGetSpeedPercent(curIndex);
-    switch(key_num)
+    switch (key_num)
     {
-      case KEY_ICON_0:
-        if (actFan > 0)
+    case KEY_ICON_0:
+      if (actFan > 0)
+      {
+        if (infoSettings.fan_percentage == 1)
         {
-          if (infoSettings.fan_percentage == 1)
-          {
-            fanSetSpeedPercent(curIndex, --actFanPercent);
-          }
-          else
-          {
-            fanSetSpeed(curIndex, --actFan);
-          }
-        }
-        break;
-
-      case KEY_ICON_3:
-        if (actFan < infoSettings.fan_max[curIndex])
-        {
-          if (infoSettings.fan_percentage ==  1)
-          {
-            fanSetSpeedPercent(curIndex, ++actFanPercent); //adding 2 to current speed % increases speed % by 1.
-          }
-          else
-          {
-            fanSetSpeed(curIndex, ++actFan);
-          }
-        }
-        break;
-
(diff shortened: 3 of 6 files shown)


GregSKR:
Totally agree with you, @discip :+1:

BTW, I'd like to share my opinion here about the gestion of fan speed in percents.
I think only raw values should be changes whith the + / - buttons and then percents be calculated.
Actually, "minus" button is fast (values 2 by 2) and "plus" is very slow (you must click twice for 1%)
And when speed is 1%, "Minus" button will go to 100%!...

I did some tests, and completly removed:
[FanControl.c] fanSetSpeedPercent()
[Fan.c] actFanPercent
and changed code in [Fan.c] like this (from line 63) and it looks like mutch better (maybe needs some cleaning):
```
  while(infoMenu.menu[infoMenu.cur] == menuFan)
  {
    KEY_VALUES key_num = menuKeyGetValue();
    uint8_t actFan = fanGetSpeed(curIndex);
    switch (key_num)
    {
    case KEY_ICON_0:
      if (actFan > 0)
      {
        if (infoSettings.fan_percentage == 1)
        {
          if (actFan > 3)
		  {
			  actFan = actFan - 3;
		  }
		  else
		  {
			  actFan = 0;
		  }
		  fanSetSpeed(curIndex, actFan);
        }
        else
        {
          fanSetSpeed(curIndex, --actFan);
        }
      }
      break;

    case KEY_ICON_3:
      if (actFan < infoSettings.fan_max[curIndex])
      {
        if (infoSettings.fan_percentage == 1)
        {
          if (actFan < infoSettings.fan_max[curIndex] - 3)
		  {
			  actFan = actFan + 3;
		  }
		  else
		  {
			  actFan = infoSettings.fan_max[curIndex];
		  }
		  fanSetSpeed(curIndex, actFan);
        }
        else
          fanSetSpeed(curIndex, ++actFan);
      }
      break;

    case KEY_ICON_4:
      if (infoSettings.fan_count > 1)
      {
        curIndex = (curIndex + 1) % infoSettings.fan_count;
        fanSpeedReDraw(false);
      }
      else
      {
        fanSetSpeed(curIndex, (infoSettings.fan_max[curIndex] / 2) + 1);
      }
      break;

    case KEY_ICON_5:
      fanSetSpeed(curIndex, infoSettings.fan_max[curIndex]);
      break;

    case KEY_ICON_6:
      fanSetSpeed(curIndex, 0);
      break;

    case KEY_ICON_7:
      infoMenu.cur--;
      break;

    default:
      #if LCD_ENCODER_SUPPORT
        if(encoderPosition)
        {
          if (actFan < infoSettings.fan_max[curIndex] && encoderPosition > 0)
          {
            fanSetSpeed(curIndex, ++actFan);
          }

          if (actFan > 0 && encoderPosition < 0)
          {
            fanSetSpeed(curIndex, --actFan);
          }
          encoderPosition = 0;
        }
      #endif
      break;
    }

    if (lastFan != fanGetSpeed(curIndex))
    {
      lastFan = fanGetSpeed(curIndex);
      fanSpeedReDraw(true);
    }

    loopProcess();
  }
}
```

digant73:
@guruathwal I see a conflict on MachineSettings.
IMHO it should be better to change the meaning of QUICK_EEPROM_BUTTON to allow the use of compact menu (without the need to sacrifice other menus as the terminal).
I would use QUICK_EEPROM_BUTTON to choose between the current implementation based on compact menu and your preferred version that need to sacrifice other menus. Everyone has different tastes but we should avoid to change so frequently menu layout. Both are correct, but honestly I prefer by far my implementation

guruathwal:
Hi @digant73, No menu is being sacrificed. it is already available in the main menu. IMHO the use of the next button is to toggle a different icon is not the best way. The use of the next button that way is ambiguous for the user as it does not give any indication of what it will do when pressed. Isn't it better to show the normal way when possible? It would have been more easily accessible if the EEPROM icons were in a different menu, which would give equal access to all 3 buttons.

digant73:
@guruatwhal yes, I agree that it is always better to use icons when there is room enough in the menu. However the use of the compact menu leave space for Terminal and also for Tuning. It's not a drama in any of the two scenarios (I don't see any problem on pressing on an icon to show the next correlated EEPROM features), but when possible I prefer to not remove features that were in the menu (with the exception of all the connection features that were previously spread in different menus due to the absence of a specific Connection menu).
However, I will rollback my MachineSettings. In any case it's better than having nothing as in the past.
I think the feedback from bigtreetech should be important for such kind of eesthetic aspects.

About the flash usage, is there an indication of the weight of the different kind of info stored in it (e.g icons)?

PR #990, opened 2020-08-18
Updated firmware binaries

traffic-light:
I made a branch for updating the binaries.
The latest commit message will show the date and time

Example: **Updated bin 18-08-2020@10:07-UTC0**

I'll try to make these so now an then inside this branch.

And so now and then I will make a pull request for it ;-)
(ONLY when there is a bug fix/ big improvement. (Especially when people talk about it ;-) )


traffic-light:
Wi

traffic-light:
Will reopen at the end of the day (Maybe more commits/ merges will come this day)

PR #996, opened 2020-08-19
 cleanup extruder tuning, Chinese Translation

Msq001:
### Requirements

### Description
* cleanup extruder tuning, The hotend temperature will be too cold  that can't extrude if set target temp to 0 when extrude 100mm
* Chinese Translation


traffic-light:
Thanks. Forgot to remove that. Before there was a question if it was done with extruding. Thanks!
This will avoid damage to the filament ;-)

PR #998, opened 2020-08-19
fix & cleanup onboard sd print

Msq001:
### Requirements

### Description
* cleanup `requestCommand` code, Flash use reduced 100 bytes, Global RAM use reduced 470 + bytes
* M21 start Magic “SD” modify to "SD Card ", because error Magic "No SD Card" will be parsed to start Magic

### DIFF
--- a/TFT/src/User/API/Gcode/gcode.c
+++ b/TFT/src/User/API/Gcode/gcode.c
@@ -1,23 +1,47 @@
 #include "gcode.h"
 #include "includes.h"
 
-REQUEST_COMMAND_INFO requestCommandInfo;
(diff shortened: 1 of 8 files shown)


digant73:
as far as I remember the babystep menu on More menu was restored because the babystep menu provided in the Printing menu can be used (replaced) by the Preview menu. I will eventually avoid to have different menus in the Printing menu. E.g I will provide Preview menu in Printing menu and babystep menu in More menu.

ETE-Design:
@digant73 Your righton that, when using Preview the Babystep function is not possible, thats why it was moved to the more menu :-)

PR #1004, opened 2020-08-19
Change Extrude icon in Tuning menu

Guilouz:
I think it's better to use extrude icon (without 100mm inscription) in Tuning menu.

IMG_1666

### DIFF
--- a/TFT/src/User/Menu/Tuning.c
+++ b/TFT/src/User/Menu/Tuning.c
@@ -9,7 +9,7 @@ void menuTuning(void)
     LABEL_TUNING,
     // icon                         label
     {{ICON_PID,                     LABEL_PID},
-     {ICON_EXTRUDE_100,             LABEL_TUNE_EXTRUDER},
+     {ICON_EXTRUDE,             LABEL_TUNE_EXTRUDER},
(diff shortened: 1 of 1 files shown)


traffic-light:
This doesn't say it's tuning the extruder. And not the right image.

@bigtreetech please ignore this request

bigtreetech:
@traffic-light Because the external flash(8Mbytes W25Q64) used to store the icons is almost used up, I plan to clean up some icons with similar functions. The same icon is used for similar functions, and the differences are reflected by `LABEL`.
What do you think

radek8:
> @traffic-light Because the external flash(8Mbytes W25Q64) used to store the icons is almost used up, I plan to clean up some icons with similar functions. The same icon is used for similar functions, and the differences are reflected by `LABEL`.
> What do you think

good idea

traffic-light:
@bigtreetech I think about changing the ext-100 can be changed to a more logic one telling you it's for tuning extruder. I'm not so good in designing and have no idea yet how it should look like.
Them change the icon in the part where you set the temperatures for the hotend into a regular extruder. The label underneath it already tells what it is supposed to do. Extruding 100mm

PR #1005, opened 2020-08-19
Add new E-steps stored to EEPROM

Guilouz:
Add new E-steps stored to EEPROM

### DIFF
--- a/TFT/src/User/Menu/TuneExtruder.c
+++ b/TFT/src/User/Menu/TuneExtruder.c
@@ -239,6 +239,8 @@ void menuNewExtruderESteps(void)
       {
         char tmpBuf[120];
         storeCmd("M92 T0 E%0.2f\n", new_esteps);
+        if(infoMachineSettings.EEPROM == 1)
+        mustStoreCmd("M500\n");
 
         sprintf(tmpBuf, (char*)textSelect(LABEL_TUNE_EXT_ESTEPS_SAVED), new_esteps);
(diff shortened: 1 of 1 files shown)


traffic-light:
There is a reason that's not saving in eepprom. You should do it yourself. Just like in marlin.  It's bad to save too many times to eeprom

@bigtreetech: please ignore this pull request...

Guilouz:
Why too many ? Once you have tune your e-steps, you save it once and that's all...

traffic-light:
Every save cycle makes the eeprom live shorter. It's a max of 100000 cycles. When you press the button 2 times the eeproms live is shorten with 2 possible saves

Guilouz:
Yes I know, but what is the difference between saving it directly or exiting the menu and going to save it in another menu?
Or maybe add an optional button to save it in eeprom for people want to save it directly ? 
Generally we do not save in the eeprom many times once the settings have been made and saved.

No the third commit is for updating translation file...

traffic-light:
Saving it directly saves it on the ram of the microcontroller. Save it with m500 saves it in eeprom.
Ram memory gets 0 on reboot/ powerloss. Eeprom stays the same.

Eeprom memory is some sort of burn memory. When writing to it it's life cycle goes down.

So she. Writing to eeprom ALL the parameters in the ram of your microcontroller will be written to eeprom.


There is already a button in the machine settings. It will be a mess when placing everywhere a eeprom save button.

I know it's updating the language. But it could be in one pull request.

Title could've been: update language_xx and tune extruder menu

PR #1007, opened 2020-08-19
Updates for fw binairies

traffic-light:
I made a branch for updating the binaries.
The latest commit message will show the date and time

Example: Updated bin 18-08-2020@10:07-UTC0

I'll try to make these so now an then inside this branch.

And so now and then I will make a pull request for it ;-)
(ONLY when there is a bug fix/ big improvement. (Especially when people talk about it ;-) )


redgod:
the last compilation does not work in tft24 v1.1, after a few minutes the tft freezes and stops working and I cannot find the error

traffic-light:
Oh. What a shame.  I can't test it at the moment. My screen is broken.

When does it freez?

redgod:
leaving the tft only after 1 minute on the main screen the touchscreen does not respond and the button is not frozen completely

traffic-light:
I've asked someone with a TFT35-E3. There is no problem on that screen. maybe it's only for your screen type. Did you already make a issue?

redgod:
There is another user with the tft24 v1.1 that also has the same problem, to see it is with the tft24 the freezing

PR #1009, opened 2020-08-20
Updated bin 20-08-2020@14:452 UTC0 (5d2c70a)

traffic-light:
I created this pull request because there is a fix for the onboard sd-card.
So important enough to update the binaries

(I noticed there is enough audience that likes it pre compiled. Isn't it possible that the build test (when something is merged) generate a .bin file for every board and automatically places this in the "copy to sd-card folder"?)


traffic-light:
Don't understand why it tells it conflicts. But it shouldn't have any conflict. It's just replacing the old binaries.

PR #1013, opened 2020-08-22
clean status screen code, update icons, separate notification library for toast notification

guruathwal:
clean status screen code
- remove duplicate axis location storage.
- update some new icons.
- move toast notification to separate files.
- move menu.h & menu.c to API directory.
- remove duplicate LCD dimmer code.


AntoszHUN:
Hi! It's all beautiful and good. But again, why did you have to vary the icons? When will the icons finally be finalized? I don't understand that if something is good once, why doesn't it stay that way?

Sincerely! AntoszHUN

bigtreetech:
@AntoszHUN Hello, Because the flash space for storing icons is running out of space, we need to clean up some useful but not very important icons

AntoszHUN:
I was thinking about the look of the icons. Couple icon. Exactly numerically 9 pieces have been changed. I wondered why? Did anyone like it? Obviously, everyone customizes the display for themselves. I have my own icons because it doesn’t change every week. I did it once and it was done. The firmware itself would not take up much space in flash memory if it only contained useful enhancements. Obviously, not everything can be put into it. But with so much power, a microSD slot on the display motherboard could be solved to have the firmware on it like on BTT motherboards. A 128MB card would be enough.

bigtreetech:
@AntoszHUN 
  This PR only removes an `unlock xy` icon (because the function of unlockxy is the same as disable motor, but the difference is only one unlock xy and one unlock xyze) and add missing icons of other resolutions LCD.
  

>  a microSD slot on the display motherboard could be solved to have the firmware on it like on BTT motherboards. A 128MB card would be enough.


 Does this mean that the icon files should be stored in the SD card instead of SPI Flash(W25Qxx) in TFT?

AntoszHUN:
> 
> 
> @AntoszHUN
> This PR only removes an `unlock xy` icon (because the function of unlockxy is the same as disable motor, but the difference is only one unlock xy and one unlock xyze) and add missing icons of other resolutions LCD.
> 
> > a microSD slot on the display motherboard could be solved to have the firmware on it like on BTT motherboards. A 128MB card would be enough.
> 
> Does this mean that the icon files should be stored in the SD card instead of SPI Flash(W25Qxx) in TFT?

It could be solved. Or not? You just need an SD slot for the displays. There are less important files. Although it could be solved with the current SD slot. Only if the user removes the SD card will the connection to the icon files be lost.

guruathwal:
Hi @AntoszHUN the icons were updated to match visually with the rest of the icon set and give a more general meaning so they can be easily reused for other purposes. Also some of the icons are changed to use symbols instead of the English words so they can be language neutral and some to avoid copyright infringement as the source/license of the icons were unknown.
At this point of time it is necessary to reducing the icons count because some of the SPI flash storage is needed to offload the data from the main program memory to give space for adding more features to the firmware.
The Master branch in this repository is being used as development branch and currently the only way for the devs. to add/update/fix anything is through PRs. All the changes are be released as single update in the release branch which is generally recommend for use.

The micoSD on the main controller boards is only used to flash the firmware onto the program memory of the boards. Using the microSD to store the icons will result in firmware being slow as reading files from microSD is slower than reading data from the SPI flash. It will also result in hangs/freeze when the microSD is removed to transfer the gcode files.

AntoszHUN:
> 
> 
> Hi @AntoszHUN the icons were updated to match visually with the rest of the icon set and give a more general meaning so they can be easily reused for other purposes. Also some of the icons are changed to use symbols instead of the English words so they can be language neutral and some to avoid copyright infringement as the source/license of the icons were unknown.
> At this point of time it is necessary to reducing the icons count because some of the SPI flash storage is needed to offload the data from the main program memory to give space for adding more features to the firmware.
> The Master branch in this repository is being used as development branch and currently the only way for the devs. to add/update/fix anything is through PRs. All the changes are be released as single update in the release branch which is generally recommend for use.
> 
> The micoSD on the main controller boards is only used to flash the firmware onto the program memory of the boards. Using the microSD to store the icons will result in firmware being slow as reading files from microSD is slower than reading data from the SPI flash. It will also result in hangs/freeze when the microSD is removed to transfer the gcode files.

I understand. So then it cannot be solved. It's okay. there will be some solution to it one day. Maybe icons converted to .bin file? They take up less space, don't they? The firmware could be rewritten to read bin files.

digant73:
@guruathwal but how the icons are converted and stored in the flash? In my TFT all the icons are about 2MB (in bmp format) so the total weight in the flash (8MB) should be about 25%.

bigtreetech:
@AntoszHUN 
In fact, the icon stored in W25Qxx is already a bin file,. The firmware of TouchScreen will read BMP file and parse it into .bin and store it in W25Qxx. The color value of each pixel is 16bit (2bytes).  For tft35, the memory required for full screen (480 * 320) is 480 * 320 * 2bytes = 300KBytes.
If you put the icon size into the SD card, the interface refresh speed will also become very slow. At present, the way of icon refresh is the fastest way in this hardware. Using DMA to directly write the color value to TFTLCD from W25Qxx.
I have tested that for TFT35 V3.0, It can refresh the full screen 480 * 320 pictures up to 10 frames per second in this way . If you read from the SD card, it may not reach 1 frame per second.

bigtreetech:
> @guruathwal, @bigtreetech but how the icons are converted and stored in the flash? In my TFT all the icons are about 2MB (in bmp format) so the total weight in the flash (8MB) should be about 25%. I would expect that the icons were also optimized (compressed) when stored in the flash reducing a lot the weight of the icons

In addition to icons, there are font files stored in flash, which share all W25Q64 8MB space.
As mentioned above, if the icon is compressed for stored, it needs to be decompressed before being displayed, which will lead to a very slow refresh rate

AntoszHUN:
Now I understand everything. Thank you everyone!

digant73:
> > @guruathwal, @bigtreetech but how the icons are converted and stored in the flash? In my TFT all the icons are about 2MB (in bmp format) so the total weight in the flash (8MB) should be about 25%. I would expect that the icons were also optimized (compressed) when stored in the flash reducing a lot the weight of the icons
> 
> In addition to icons, there are font files stored in flash, which share all W25Q64 8MB space.
> As mentioned above, if the icon is compressed for stored, it needs to be decompressed before being displayed, which will lead to a very slow refresh rate

Yes, I know also the font files are stored in the flash (I was referring to icons only because it seems the only thing that could be reduced to free some space in the flash). But even with the font files (they are about 2.33MB) I should reach about 50% of flash usage while from the Info menu I see it's about 94%. Just to understand why I see an extra 40% of usage

PR #1014, opened 2020-08-22
improvements, cleanup, fixed bug #1015

digant73:
IMPROVEMENTS/BUG FIXING:
* LCD_Encoder,c: improved detection of button when pressed. A delay was added between the pulse 0 and pulse 1. This improves a lot the detection of the button when pressed.
* interfaceCmd.c: a toast dialog is used for notifying M117 messages. This replaced the previous use of dialog box requiring an annoying user interaction
* MachineSettings.c: as already done for other featurues (e.g. Connection), EEPROM buttons were grouped in a menu called EEPROM
* Pid.c: minor improvements, plus cleanup
* TuneExtruder.c: improved message displaying and message handling, plus cleanup
* Leveling.c: fixed a bug on Z Fade editor reporting always 0 as min and max values
* Fixed bug #1015. M150 gcode is now used to control LED


traffic-light:
@digant73 is your version somewhere online? It sounds interesting to have!

digant73:
@traffic-light it's on https://www.thingiverse.com/thing:4294049. It's the binary only. I just made some changes in some menus according the feedback provided by the guys. Most of the time I collect a list of the bugs to be solved

traffic-light:
Ah. Okay.

digant73:
@traffic-light I always try to merge my changes in the official BTT main branch. Currently only some menu changes are present in the binary. As you know anyone has different preferences about the GUI and it's impossible to reach an agreement

traffic-light:
> @traffic-light I always try to merge my changes in the official BTT main branch. Currently only some menu changes are present in the binary. As you know anyone has different preferences about the GUI and it's impossible to reach an agreement

Jup. That's indeed impossible. Thank you for the link. I'll use some parts of your ideas there.

oldman4U:
May i ask a question.

I can see, that for most (if not for all) values which are saved to EEPROM I am now asked if I would like to save them to EEPROM once i changed the setting. This - for me - is the best way to handle this. The ability to delete the EEPROM and to restore it is only important at very specific times, for example after loading a new firmware. Because this does not happen very often, the little more hidden functions in the old version worked very well for me and the new more prominent place is nice but not really important to me. So I ask myself if i did not get the point, that this functionality needs a more prominent place. Please let me know.

Thank you

digant73:
@oldman4U
Which version did you prefer? I didn't understand if as "little more hidden functions" you refer to the compact EEPROM sub menu (with toggle button in the MachineSettings menu) or something else. The EEPROM menu was added simply because there were no EEPROM menu present (there was just an item in the custom gcodes or Parameter settings)

oldman4U:
Hi digant73.

It is a version from approximately 3 weeks ago. And i am not saying that this is bad, i just wonder if i miss the point.

Thank you
IMG_7056

IMG_7055

digant73:
@oldman4U Ah ok understood. These were present since many time. But custom gcodes are typically modified according to our needs. E.g. I replaced all the default gcodes to provide other featues that are really custom and depending on my printer (e.g. present bed, extrude etc... The provided default custom gcodes are basically mandatory features that must be provided in their specific categories and not provided as generic custom features. That was the main reason I also collected typical connection features on the Connection menu. Of course menus must also be contextual, so once a feature is implementd it could be available in other menu just to improve the use case (e.g. save button is present in babystep menu etc...)

oldman4U:
There are other settings like invert axis and so on and for me this would
be the right place to have all the EEPROM specific functions.

But i agree when you say that it is impossible to make it right for
everyone ( incl. me😁)

On the other hand resources seem to be very limited and bottons are changed
frequently because of this.  So requiring additional recourses for
something which was there before, without adding additional benefits sounds
like a luxury to me.

Just my 2 cents

Thank you

digant73 <notifications@github.com> schrieb am Mi. 26. Aug. 2020 um 17:51:

digant73:
@oldman4U yes I know, the problem of the resources is arising now but everything can be improved. Resources could be optimized in the flash. Also the weight of the icons is less than 25% in the flash. It's not an icon more (in a total of about 140) the main cause of flash usage. The fact that something can be reached in different ways is the reason of many different point of views and priorities. There are features such as Power off or even Emergency stop that basically are not used by anyone but are there while they could be provided as custom gcode. I normally collect feedback from users on thingiverse, were I provide a bundle for my printer, and the main critics are related to the lack of a clear idea or guide lines on how menus and features must be provided (one was related to a missing access to EEPROM features organized in a unique main menu as it is normally found in any kind of device. I just try to follow standardized menu organization provided in many other type of devices. There were also a lot of critics regarding the pretty much useless Screen menu that does not allow to access the most used features such as preheat, terminal etc...). In my distribution I rearranged some menus according to the feedback received. But obviously I will never provide them in one of my PR

oldman4U:
Good to hear that there is a way to improve resources. Now we have to find someone who can do the job;-)

Otherwise, that's my favorite when it comes to features and making decisions because someone somewhere asked for it;-)
https://www.youtube.com/watch?v=H8eP99neOVs

PR #1065, opened 2020-09-06
Fixed update of print speed value in PrintingMenu

### LINKED ISSUE #1048: [BUG] Update the print speed value when printing
If I change the print speed while printing (for example, with an encoder directly in Marlin), the speed will not change on the PrintingMenu screen until I open the Speed Control menu. Only then will the Speed value be updated.
If it does not print, the StatusScreen screen works fine

https://youtu.be/kRYwBqbGrRU

radek8:
Fixed update of print speed value in PrintingMenu

Resolves #1048 

PR #1051 only worked when heating the pad and hotend. The print speed value was no longer updated after the print head started moving.
PR #1065 works all the time

### DIFF
--- a/TFT/src/User/Menu/PrintingMenu.c
+++ b/TFT/src/User/Menu/PrintingMenu.c
@@ -261,6 +261,7 @@ void toggleinfo(void)
       reDrawFan(FAN_ICON_POS);
     }
 
+    storeCmd("M220\nM221\n");
     c_speedID = (c_speedID + 1) % 2;
     nextTime = OS_GetTimeMs() + toggle_time;
     rapid_serial_loop();   //perform backend printing loop before drawing to avoid printer idling
--- a/TFT/src/User/API/SpeedControl.c
+++ b/TFT/src/User/API/SpeedControl.c
@@ -60,7 +60,7 @@ void loopSpeed(void)
 
 void speedQuery(void)
 {
-  if (infoHost.connected == true && infoHost.wait == false)
+  if (infoHost.connected == true)
   {
     if (!queryWait)
     {


TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -261,6 +261,7 @@ void toggleinfo(void)
       reDrawFan(FAN_ICON_POS);
     }
 
+    storeCmd("M220\nM221\n");
```
guruathwal, line 264:
```suggestion
    speedQuery();
```
Use speedQuery(); to ensure the next query is not added to buffer before the first command is sent, and also reduce code size.

radek8: I've already tried that in request # 1051
It only works if the hotend and bed are being heated.
As soon as G-codes start to be sent to move the head, the speed icon stops updating.
There is probably a problem with the condition of the speedQuery ();

radek8: @guruathwal  you have a suggestion on how to modify speedQuery (); to work in this case as well?

radek8: @guruathwal your change proposal only works if I also adjust speedQuery this way. Do you agree?

image

guruathwal: I will check and revert to you.

guruathwal: @radek8 the implementation of infohost.wait needs work

radek8: I don't understand your answers.
under the condition (infoHost.wait==false), updating the print speed value does not work correctly.
Do you know another way to solve it?

Issue #1000, opened 2020-08-19
[Q] Bootloader TFT50


Rebel3D-CZ:
Hi, I have a TFT50 with a faulty CPU. I replaced the CPU with a new one. Is it possible to get a bootloader for TFT50 with the procedure how to upload it?
I have ST-Link. I will connect to the processor.

radek8:
maybe someone who has the same display will help and copy the chip's memory :-)
viz
https://github.com/bigtreetech/BIGTREETECH-TFT35-V3.0/issues/20

radek8:
instructions try this one
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/TFT35_V2%20Bootloader%20fix/BTT-TFT35-V2.0-Bootloader-flashing-guide-CH340-debugger.pdf

Rebel3D-CZ:
I believe the manufacturer's support is solid and the bootloader will provide its customers :)

Rebel3D-CZ:
Is the support really unable to provide customers with a bootloader? Very sad :(

av71966:
> Hi, I have a TFT50 with a faulty CPU. I replaced the CPU with a new one. Is it possible to get a bootloader for TFT50 with the procedure how to upload it?
> I have ST-Link. I will connect to the processor.

Hi, i have simmilar problem and have second working tft50 and want to repair broken one. If you help me with copying eeprom i can send it to You

Rebel3D-CZ:
Hi, I have another TFT50 in about 2 days. I'll read the bootloader and save it here. Chinese ST-LINK V2 is enough to read. Then I will try to help or I will proceed here, but now I am 2 days on vacation :)

Rebel3D-CZ:
Everything you need here :)

https://1drv.ms/u/s!AvBL3Ps7S-g3ktIRlwZE_Jj3X-WJCQ?e=6MwVQf

radek8:
Great.
Don't you want to post it on the github page so that it is available to others who might need it?

Rebel3D-CZ:
Unfortunately, I don't have much experience with github. I don't know where it can or cannot be given and what the rules are. If you can, put the data where it belongs.

[...]

PR #1101, opened 2020-09-17
Bootloader for display TFT50 V3.0

radek8:
Bootloader for display TFT50 V3.0

See #1000


oldman4U:
I reply so it does not get abandoned in 2 weeks.

;-)

Issue #1015, opened 2020-08-22
[Bug] Protected Pin error when trying to PID tune bed / Hotend


Hyrules:
Hi everyone, 

I'm new to the TFT3.5 V3 / SKR 1.4 Turbo community and i'm struggling with a few issue on the TFT3.5 V3.

Whenever I try to do an PID autotune on the TFT screen, (in touch screen mode) it gives me an protected pin error and get stuck in the pid autotune in progress screen. At some point I get a message to save my pid settings but the screen never disappear. Any one has idea what is happening ? Thanks. I'm also using marlin 2.0.6 bug-fix firmware.

bigtreetech:
It's because `M42` command in PID menu,  I'll deleted these non generic gcode to avoid similar problems

digant73:
> It's because `M42` command in PID menu, I'll deleted these non generic gcode to avoid similar problems

it should be replaced with a standard gcode, if any

Hyrules:
Thanks and keep up the good work ! Let me know when If I can test anything.

digant73:
Fixed by PR #1014

digant73:
> Thanks and keep up the good work ! Let me know when If I can test anything.

the fix is now merged in the BTT main branch. Please, can you verify if now the problem is solved and eventually close this bug request?

Hyrules:
I don't have the PID Protected Pin error anymore but I have a mcode error :  M150 R255 U0 B0 when I try to pid tune (tuning does itself anyway)  I don't have a led. The only led I see is in the corner of the screen sometimes. This seems to be ignoring the fact that I put knob_led_color: 0 in the config.ini. Might be related.

digant73:
ok the error you got it's due to the standard gcode to change the LED color. However you can ignore it (because you don't have any LED) the PID process should run correctly. The Knob led is another kind of LED

Hyrules:
There should be a way to not have this error message at all. maybe silence the error if there is no led. Normally should not produce an error if there is no led.

digant73:
I have to see. You should not receive any error if you don't have any LED

Hyrules:
Something else has also changed. Now when I want to print, the printer does a bed level before each print which should not happen if I have already a bed leveled. I have to print something else again to confirm that. I will finish my current print in a few hours so I will start something else after.

digant73:
I didn't yet installed the new release with the new PRs. I will have a look

[...]

Issue #288, opened 2020-01-08
[FR] Ability to edit bed-leveling mesh grid in touch mode


Hukuma1:
**Is your feature request related to a problem? Please describe.**
![ubl](https://user-images.githubusercontent.com/54511555/72014393-9f34e500-3214-11ea-8cfe-4c0c64f6a920.jpg)
Taken from: https://youtu.be/RudXmCgobso?t=224

Would love for BTT to incorporate a way to edit the mesh data that ABL creates in Marlin. UBL has this feature for Marlin mode where you can switch between the points using the knob, which definitely can be improved with touch mode I think. Ideally this would work for bilinear bed leveling, not just UBL. Would love to quickly adjust a specific mesh point with a visual guide (e.g. my left quadrant of the bed needs to be slightly lower Z offset than the rest of the bed).

**Describe the solution you'd like**
ABL/UBL manual editing of mesh z-data points after probing. But I guess if you add this code, you can also add the ability to just add the manual bed leveling mesh creation too?

guruathwal:
that looks like a 5 inch screen...

Hukuma1:
Harder to implement, for sure.

Marlin does have ULB with a similar grid layout though. That would fit:
https://youtu.be/ONpKxkil16Q?t=914

traffic-light:
I decided to make this. I hope that bigtreetech will awnsers my email about a replacement unit soon. And that I would receive one as soon as possible.

Then I can test it when building ;-)

When there is a version that's, in my opinion, good enough to be checked by others I will create a draft.

oldman4U:
Hello Hukuma1.

Your requested feature is available through PR #819 

Please help the community and close this ticket.

Thank you

Hukuma1:
Great work! Thank you all! Closing. :)

PR #1111, opened 2020-09-20
Added Mesh Bed Leveling + optimized some code defining new API, improvements and bug fixes

digant73:
### Description

Improvements:
* defined new APIs:
 1) BabystepControl
 2) ProbeOffsetControl
 reusable for Babystep, Probe Offset, MBL and the future Mesh editor
* Added BL menu hosting all common BL features (Auto and Manual) such as:
  * ABL, UBL and MBL menu
  * ZFade
  * BL on/off 
  * a future Mesh Editor
* Added Mesh Bed Leveling menu
* PID: minor improvements
* other improvements

Fixes:
* Fixed bug #1015

**PR Status:**
**FINAL**: the PR is ready to be merged on BTT main branch

**NOTE:**
Previous FLASH space problems were fixed. Now it's about 87% of FLASH usage on all the TFT variants


digant73:
@bigtreetech The failed checks are related to no flash space availbale on TFT35 E3 V3 and above. Is there any solution applicable to solve the problem (e.g. make configurable the languages to include) or we can say THE END :-)?

szepnorbee (approved):

bigtreetech:
Yes.  At present, all language characters occupy about 70-80KBytes of flash memory. we can add the macro definition to select the language to use in the source code to reduce the memory usage.
And, The current source code also needs to be cleaned up to reduce some duplicate code.
And we will do it in next work.
Thanks for your hard work!

digant73:
@bigtreetech Good thanks. I will mark my PR as draft version (not ready to be merged), merging it with your next updates
br.

traffic-light:
@digant73
So you're about to make the #288 ?

Then I will quit working on it...

Hukuma1:
Combine your efforts @traffic-light and @digant73! :)

digant73:
@traffic-light 
yes, I''m starting to work on it during some free time.
First I will work on the GUI layout and then to the parser

traffic-light:
@bigtreetech maybe it's handy to change the names, for example en_cancel, without the language subscript. So for the example given: *en_cancel* into *cancel*.

This way it's easier to implement new words in all languages files.

digant73:
@bigtreetech 
Previous FLASH space problems were fixed. Now it's about 87% of FLASH usage on all TFT the variants.
Optimized some code creating new APIs to handle babystep and probe offset.
Those APIs are used also by the new MBL menu and the future Mesh editor.
My PR is in final version and could be merged on your main branch

ETE-Design:
@digant73 Could you mabye look at the Runout Sensor, seems like it is not working, at least when you have it connected direct to the mainboard :-( The Runout trigger, but the TFT does nothing...

radek8:
> Yes. At present, all language characters occupy about 70-80KBytes of flash memory. we can add the macro definition to select the language to use in the source code to reduce the memory usage.
> And, The current source code also needs to be cleaned up to reduce some duplicate code.
> And we will do it in next work.
> Thanks for your hard work!

@bigtreetech
What if there were only two languages in the code. Primary - English and secondary also English.
In the display menu I would be able to select only the primary and secondary language, but the secondary jakzyk could be overwritten, for example, by using the language.ini file at boot time?

freez12:
Manual mesh is not getting saved and when i do a Manual mesh leveling i always get m851 unknowen commad.

radek8:
> Manual mesh is not getting saved and when i do a Manual mesh leveling i always get m851 unknowen commad.


You do not have Baby steping allowed in Marlin

freez12:
> > Manual mesh is not getting saved and when i do a Manual mesh leveling i always get m851 unknowen commad.
> 
> You do not have Baby steping allowed in Marlin

i hve Babystepping enabled in marlin when i use lcd mode everything is working mmbl also.

PR #1129, opened 2020-09-26
Update language_fr.h

Guilouz:
Update french translation


Guilouz:
Why compilation doesn't pass for TFT35 E3 v3.0 ?

PR #1130, opened 2020-09-26
Update language_fr.h

Guilouz:
Update french translation


Guilouz:
@bigtreetech Why TFT35 E3 V3.0 doesn't pass the compilation ?

Guilouz (commented):
small fix

traffic-light:
Because you made the flash overflow ;-)

Bigtreetech is working on it
 You don't need all this languages. So it would be selected when building.

Issue #1050, opened 2020-09-01
[Q] How to display M117 messages in touch mode info area on TFT35 E3 V3.0?


StarNamer:
**Your question or details you need. Please describe.**
When in Marlin (12864 simulator) mode, M117 messages regarding progress etc are displayed on the last line of the screen. How do I get these to display on the Info Area when in touch mode. With the current firmware, the text "Ender-3 Ready" is always displayed.

**specify the device variant related to your question.**
TFT35 E3 V3.0.

**Additional context**
So far, I've only tested this while printing GCode sent via USB from OctoPrint, either as a print job or as terminal commands. The test displays in Marlin mode, but not in touch mode.
.

traffic-light:
I've got the same issue.

I think it's some sort of lost with an update....

StarNamer:
If I type an M117 command in the touch mode 'terminal' on the TFT then if gets displayed in to info area (though there's not must you can type). If I type an M117 command in OctoPrint's 'terminal' then it gets displayed in Marlin mode, but the info area in touch mode doesn't change.

traffic-light:
A couple of day's ago I did take a look at it. Couldn't find anything..

I'll take a look again. maybe I'm lucky today :-P

radek8:
In the parseACK.c file you define messages that will be suppressed and will not be displayed. Try to allow what you need.

StarNamer:
What I need is for it to display the text reported by the M117 command in GCode. How do I allow it to do that?

radek8:
Comment out these lines (//) in the file parceACK.c

![image](https://user-images.githubusercontent.com/46979052/91933868-29711280-ecea-11ea-9ba5-88ce0f5443ed.png)

traffic-light:
@radek8 already did that. Also added notify toast ;-)

Doesn't work

radek8:
Then Marlin can filter it.
Swap ports in Marlin configuration.
Set the touch screen to serial_port,
on serial_ port2 set Octoprint
I don't know if it will help but I would try it.
Another option is to connect the Octoprint to the serial port of the display, but you probably won't want that ...

StarNamer:
@radek8 I don't understand your comment. Are you suggesting to change the Marlin firmware to swap the definitions of SERIAL_PORT and SERIAL_PORT_2? However, you "don't know if it would help"?

Currently:

`/**
 * Select the serial port on the board to use for communication with the host.
[...]

PR #1138, opened 2020-09-28
Update icons, Fix ABL G-code sending on mode change, Fix text overlap in update screen, Move language translations to SPI flash to reduce main memory usage, Fix other Issues/Bugs

### LINKED ISSUE #365: [FR] IDEA: move language.c into language.ini
@guruathwal

please consider moving language support form code into configuration,
in other words, from `mcu flash` into `language.ini` and then into `spi flash`

similarly to how it is done for use-managed configuration support with `config.ini`
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/360

major benefits:
* for users: can fix and customize their language themselves
* for developers: no need to maintain languages they do not understand :-)

### LINKED ISSUE #1058: Bug when printing from Marlin mode.
### Description

When a print is launched from the Marlin mode, the display changes and automatically activates the TFT mode

### Steps to reproduce

1. [Update to latest release]
2. [launch print from Marlin mode]

### Hardware Variant
TFT35 V3
Latest Firmware for TFT35 V3 with SKR 1.4 Marlin Bugfix 2.0.6.1

### Additional Information

### LINKED ISSUE #1176: bug in code
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/dd12c39659f3ca54f82f5da4f2f6e20476f24010/TFT/src/User/API/Printing.c#L184

E_axis indeed is Z_axis
correct with 
Line 184 in dd12c39

 mustStoreCmd("G1 Z%.3f F%d\n", tmp.axis[Z_AXIS] + infoSettings.pause_z_raise, infoSettings.pause_feedrate[Z_AXIS]);

guruathwal:
- Update icons.
- Fix ABL G-code sending on Mode change.
- Fix text overlap in update screen.
- Fix value step button not displaying the correct value in heat menu.
- Move language translations to SPI flash to reduce main memory usage and allow switching/updating language translations without recompiling the firmware. (#365)
  - To update/ add a second language, the user needs to copy the required `language_xx.ini` file from `Language Packs` folder to the SD card root ~and rename it to `language.ini`~ and press reset button.
- Fix click blank list item in the list widget menu. 
- Prevent switching to printing screen when printing from marlin mode when SERIAL_ALWAYS_ON is enabled. (#1058)
- ~Fix wrong extruder count in case of mixing extruder.~
- Add Encoder enable signal to enable/disable encoder on marlin firmware to prevent unexpected changes on the printer in touch mode. (https://github.com/MarlinFirmware/Marlin/pull/19796) 

Resolves #365, resolves #1058, resolves #1176


Guilouz:
Why do you change some icons like eeprom_reset, eeprom_restore etc ? They are already very well like that...

digant73:
@guruathwal I'm finishing the Mesh Editor but I see some overlaps of menus. The problem is that the Mesh Editor GUI is not the standard menu (with 8 icons). In the main loop of the mesh editor I use the loopBackEnd() function instead of the loopProcess(). This is due to avoid the notification bar/temperature bar on top of the display. However some information are still displayed on the top (such as the bullet on the right or the message "busy....).
Is there the possibility to fix that problems? I was thinking a possible solution is to register a callback to redraw a custom menu.
So the loopBackEnd() could display some info (such as notification popup or error or reminder popup) but then the loopBackEnd() will finally call the registered redraw callback to reprint the custom menu.

In your PR you mentioned you fixed text overlap. Is that fix a solution also for my problem with the custom menu?
I don't see any change in your PR aboaut text overlap and ABL. Maybe you haven't committed all files yet

traffic-light:
@digant73 he did fix the overlap in the message that tells the user he needs to update the fonts, icons and or config.ini

It's fixed in the code. It's now placing the fonts according to the font height ;-)

digant73:
Below some pictures. In one you can see the problem. On top you can see the busy dor on the right and a notification

IMG_20200929_234851
IMG_20200927_210339

traffic-light:
@digant73 

Looks good!
Can't you turn off those notification temporary to complete off?
That you set notification type to none?

ETE-Design:
@digant73 Looks Great, when will you make a PR for it? Seems like you'r pretty good at this stuff... Could you mabye look at a GUI for Tramming Assistant also? :-D

guruathwal:
@discip 
Made some changes to ABL/UBL detection.
for Autopower, it looks like someone did not change the setting sign values after adding a new setting variable. resetting the settings should fix it.

@digant73 I would not recommend disabling the 'busy' icon and other similar system notifications. 
As a rule of thumb, it is better to keep the title bar visible for a menu which required more than 10-15 seconds of user interaction. 
In the pictures you posted, the grid size can be reduced as it is only for visual feedback, which will give more space for other controls. why are you using too small buttons condensed in a small area for a tiny display? You were not in favour of small buttons earlier? simplifying the UI by rearranging the elements and removing unnecessary ones will help solve the problem.

discip:
@guruathwal
Thank you. 👍
I am going to test that later.

digant73:
> @digant73 Looks Great, when will you make a PR for it? Seems like you'r pretty good at this stuff... Could you mabye look at a GUI for Tramming Assistant also? :-D

it's under testing, Just making some optimizations and cleaning here and there. Some days and it should be available. However, the main problem is that the FLASH space for some TFT variants (all the STM32F2xxx) is reached (it's 256K - 32K). For that variants the compilation fails. The workaround is to change the limt to 256K - 28K (as it is on all the other variants).
Obviuosly this is only a way to compile the code, waiting for BTT fix for the reduction of the languages included in the binary.

digant73:
> @discip
> Made some changes to ABL/UBL detection.
> for Autopower, it looks like someone did not change the setting sign values after adding a new setting variable. resetting the settings should fix it.

@guruathwal 
Some problems could be related to initMachineSetting() and setupMachine() in Settings.c, The first is never invoked while the second is invoked in parseAck.c just in some cases (e.g. if FW is not Marlin) . Both of them should be invoked to initialize all the parameters to default values

discip:
@guruathwal
After having updated to your last commit, the **ABL menu (icon) is *only visible***, if the **bed is not yet calibrated**, as soon as the calibration is done, the menu disappears. (in ***Configuration.h*** this **`#define ENABLE_BL_VALUE 1`** is set)

thanks
regards

guruathwal:
@digant73 The changes you are suggesting are already available in the source https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/4f2c665cf585ac73f3896bd826ec5f165702620f/TFT/src/User/API/menu.c#L278-L286.

>Even reducing the grid size (leaving the top part of the screen blank), the problem will not be fixed because it seems that only standard menus (using menuDrawPage()) are properly handled. 

You need to study the `SendGcode.c` and 'Numpad.c' file for implementing your custom menu and even the `Printing.c` for a mix of both.
why will you leave the title bar blank. You will have to show the title of the menu if it is the **grid preview**  menu or **point edit** menu showing the index of the current point.

digant73:
@guruathwal 
The main loop in Numpad is:
while (1)

while in the Mesh Editor I used the standard loop:
while (infoMenu.menu[infoMenu.cur] == menuMeshEditor)

giving the possibility to an external entity (e.g. loopProcess, loopBackEnd etc...)  to put on top of the menu queue another menu forcing the Mesh Editor to temporary exit and giving the control to the new menu. The Mesh Editor menu should be resumed (reloaded) by the menu handler once the menu put on top of the queue is closed (or expired in case of a toast, busy indicator etc...) but this is not happening. It seems that not all info displayed in the top bar are based on the use of menu handler. As far as I remember, popups are based on menu queue but the info displayed on the title bar area does not rely on that. The info in the title bar should be also rely on menu queue. They should be put in the menu queue and then the control should be acquired again by the previous menu in the queue.

guruathwal:
@digant73 
The loop in numpad.c is used to avoid changing menu level. Did you not read the sendgcode.c file. It also does the same thing you want to acheive.
The full redraw of a regular menu is done with 'menuDrawPage()', if this is not called by your menu and you set the menu type to custom then it should not redraw  at all. 
The title bar area is used to display all sorts of information  which are independent of the menu level and should not be used as regular menu area. If it is also done with the menu queue then the menu queue wiill keep switching between the titlebar and the actual menu triggering unnecessary screen redraws.

digant73:
@guruathwal 
yes, I understand the logic used in numpad (numpad is not a menu but a function requiring input and output attributes so it cannot be even managed by the queue handler. My menu is compliant to sendGcode (it is managed by the queue handler and use the loopBackEnd() to avoid the title bar info because they don't trigger a redrawn of the underlying menu.
But when the Mesh Editor menu is loaded by the queue manager, the artefacts in the pictures are present. They should not be drawn at all because I used loopBackEnd() in the main loop (just to avoid that kind of info). After the menu is drawn (with that artefacts), no other info are displayed on the title bar (as I was expecting also at the menu startup because I use loopBackEnd() only in my menu main loop).
The artefacts info in my pictures are due to info related to gcodes sent in the main loop just before calling the loopBackEnd(). The loopBackEnd() sends the gcode and parse also the result provided by Marlin but it also draws in the title bar notification and busy indicator etc... Why did the loopBackEnd() draw that info if they should be handled only by loopFrontEnd() (that is not present in my menu)?

traffic-light:
@guruathwal Good idea using a ini file for languages.

I don't know if anyone would agree, but I think it's easier to put every language.ini in a folder for the language instead of rename the language_XX.ini.

So example:
languages packs
.                 |______AM
.                 |            |_ language.ini
.                 |______CA
.                 |            |_ language.ini
etc.

digant73 (commented):
In main.c, shouldn't be better to call infoMachineSettings() to initialize all the machine settings to default values? isMarlinFirmware = -1 should be included in infoMachineSettings(). You currently set only isMarlinFirmware = -1 in main.c

Also setupMachine() is never executed if Marlin is detected. It is executed if FW is not Marlin or "Cap:CHAMBER_TEMPERATURE:" is detected (in parseAck()). setupMachine() enables some features such as forcing a specific BL type (if ENABLE_BL_VALUE is > 1 in Configuration.h).
shouldn't be better setupMachine is executed also if Marlin is detected?

in parseAck() you check ABL using

     else if(ack_seen("Auto Bed Leveling") && if(ENABLE_BL_VALUE==1)){     // if Auto-detect
          infoMachineSettings.blType = BL_ABL;
      }

shouldn't be better to maintain the previuos code?

    else if(ack_seen("Auto Bed Leveling")){
        if(ENABLE_BL_VALUE==1)                             // if Auto-detect
          infoMachineSettings.blType = BL_ABL;
      }

In the second scenario you avoid to continue with the other "if then else" if there is a match with "Auto Bed Leveling" but not with ENABLE_BL_VALUE == 1

ETE-Design:
Love the idea with ini files for Languages... Is it possible to do so the user just copy the ini files to the SD Card when flashing the firmware, and then the firmware just use flash that file as the language?

guruathwal:
@traffic-light  now there is no need to rename file just copy and flash.

@digant73 
the `setupMachine()` logic in `parseACK()` is now changed.

```C
      else if (ack_seen("Auto Bed Leveling") && ENABLE_BL_VALUE == 1)
      {
        infoMachineSettings.blType = BL_ABL;
      }
```
it will exit the **if loop** if the first condition is not satisfied and will not continue to the second condition just like the second one you mentioned. this way it looks cleaner and easy to understand the logic.

digant73:
@guruathwal 
yes I understand the logic you changed for ABL, MBL.. detectcion, but IMHO it creates different paths according to the value the hardcoded ENABLE_BL_VALUE. As you said, if ENABLE_BL_VALUE is different than 1 you always continue on the following if-then-else even if the target information "Auto Bed Leveling" string matched. Why should we continue to apply other if-then-else if we know no other target will match?
Putting the check ENABLE_BL_VALUE == 1 inside the check on the target string will guarantee that the same path is applied (it will ends always on target "Auto Bed Leveling". So there will be no different (unexpected) behaviour if someone configure a different ENABLE_BL_VALUE on Configuration.h.

On main.c I will also call initMachineSetting() just to always initialize all the machine parameters. Some parameters then will be updated in parseAck() by M503 and M115 but some parameters (as the software endstops state I will add in my PR) are not updated by M503 and M115 and needs a default value at startup.
In my current code I initialize isMarlineFirmware to -1 inside the function initMachineSetting() and then in main.c I call initMachineSettings() just before the infoMenuSelect() (that will send the M503...) as reported below:

initMachineSetting();      // init to default values. It must be called before infoMenuSelect()

  #ifdef LCD_LED_PWM_CHANNEL
    Set_LCD_Brightness(LCD_BRIGHTNESS[infoSettings.lcd_brightness]);
  #endif
  GUI_RestoreColorDefault();
  infoMenuSelect();
  fanControlInit();

this guarantees the initialization is perfermod before parseAck() (invocked in infoMeneSelect()) will update some machine parameters.

In the current code, initMachinSettings() is never executed (so dirty values could be accessed if not updated in parseAck()).

guruathwal:
@digant73 
> but IMHO it creates different paths according to the hardcoded ENABLE_BL_VALUE. As you said, if ENABLE_BL_VALUE is different than 1 you always continue on the following if-then-else even if the target information "Auto Bed Leveling" string matched. Why should we continue to apply other if-then-else if we know no other target will match?

Both ways will produce the same result. 
First, it will check for `Auto Bed Leveling` string, if it is not found then it will exit the if statement just like the other one. And if the string is found then only it will check the `ENABLE_BL_VALUE`. All I can do is provide an example: https://repl.it/@guruathwal/if-else-test

the initMachineSetting() should have been called after checkflashSign() in main.c

digant73:
@guruathwal 
yes, the main result is the same (the code infoMachineSettings.blType = BL_ABL; is executed or not executed with both variants).
But with your variant if you configure ENABLE_BL_VALUE to 2 in Configuration.h you should always got a dialog box reporting an error (e.g. with the text "Auto Bed Leveling"). This does not happens with the previous variant. It's just to avoid a useless error message. We know that we forced BL type to the specified value in ENABLE_BL_VALUE.

yes, I also called initMachineSettings after the flash checks in main.c (as you made in your last commits)

digant73:
@discip 
can you send the result of M503 before (when you see the ABL menu) and after (when it disappears) calibration? from a laptop with pronterface etc... send the M503 command and send the otuput

guruathwal:
@discip  now check

digant73:
@guruathwal the check on autoLevel is not enough to detect all the BL type. E.g. for MBL, autolevel is not provided by M503 so you will not see the MBL menu.

guruathwal:
@digant73  the auto-level flag is provided by M115

digant73:
> @digant73 the auto-level flag is provided by M115

yes sorry, but it's not important, if you have MBL on Marlin autolevel is not set to 1 so you will not have any MBL menu.
the M503 provide the datail about what BL type is present (ABL, MBL, UBL). blType is set according to that

discip:
@guruathwal
## I do not know which one of the latest commits did the trick, but now **all the crucial issue are ironed** out. ❤
# Thank you! 👍

In fact there is one thing, which made me wonder, what the intent of it is.

1. After having flashed your last commit, the ***ABL menu* is up and running**, but the **BL switch** is set to ***off***, nevertheless *calibrating* the bed *works as intended*.
The switch is set to ***on*** after the calibration, but then again on the next startup it is ***off*** again, so I am wondering: What the purpose of this switch might be?

2. And, if I am allowed to remind, the ***red*** switch is still ***red*** in the **`Auto Power`** entry, where it should be ***green***, because it is actually on. 😊

kind regrads

digant73:
@discip good. but unfortunately now other types of BL are not working :-)

discip:
@digant73 
Do you need the results of the latest commit, or the one not working for me?

discip:
@digant73
**Before calibration:**
echo:  G21    ; Units in mm (mm)
echo:; Filament settings: Disabled
echo:  M200 S0 D1.75
echo:; Steps per unit:
echo: M92 X20.03 Y20.03 Z99.94 E102.25
echo:; Maximum feedrates (units/s):
echo:  M203 X300.00 Y300.00 Z50.00 E150.00
echo:; Maximum Acceleration (units/s2):
echo:  M201 X2000.00 Y2000.00 Z100.00 E10000.00
echo:; Acceleration (units/s2): P<print_accel> R<retract_accel> T<travel_accel>
echo:  M204 P1500.00 R2000.00 T2000.00
echo:; Advanced: B<min_segment_time_us> S<min_feedrate> T<min_travel_feedrate> X<max_x_jerk> Y<max_y_jerk> Z<max_z_jerk> E<max_e_jerk>
echo:  M205 B20000.00 S0.00 T0.00 X8.00 Y8.00 Z0.30 E5.00
echo:; Home offset:
echo:  M206 X0.00 Y0.00 Z0.00
echo:; Auto Bed Leveling:
echo:  M420 S0 Z0.00
echo:; PID settings:
echo:  M301 P65.03 I11.95 D88.45
echo:  M304 P39.30 I7.71 D133.50
echo:; Z-Probe Offset (mm):
echo:  M851 X5.40 Y-34.20 Z-0.76
echo:; Stepper driver current:
echo:  M906 X700 Y800 Z800
echo:  M906 I1 Z800
echo:  M906 T0 E600
echo:; Driver stepping mode:
echo:  M569 S1 X Y Z
echo:  M569 S1 I1 Z
echo:  M569 S1 T0 E

**and after:**
echo:  G21    ; Units in mm (mm)
echo:; Filament settings: Disabled
echo:  M200 S0 D1.75
echo:; Steps per unit:
echo: M92 X20.03 Y20.03 Z99.94 E102.25
echo:; Maximum feedrates (units/s):
echo:  M203 X300.00 Y300.00 Z50.00 E150.00
echo:; Maximum Acceleration (units/s2):
echo:  M201 X2000.00 Y2000.00 Z100.00 E10000.00
echo:; Acceleration (units/s2): P<print_accel> R<retract_accel> T<travel_accel>
echo:  M204 P1500.00 R2000.00 T2000.00
echo:; Advanced: B<min_segment_time_us> S<min_feedrate> T<min_travel_feedrate> X<max_x_jerk> Y<max_y_jerk> Z<max_z_jerk> E<max_e_jerk>
echo:  M205 B20000.00 S0.00 T0.00 X8.00 Y8.00 Z0.30 E5.00
echo:; Home offset:
echo:  M206 X0.00 Y0.00 Z0.00
echo:; Auto Bed Leveling:
echo:  M420 S0 Z0.00
echo:  G29 W I0 J0 Z0.02043
echo:  G29 W I1 J0 Z0.25056
echo:  G29 W I2 J0 Z0.19053
echo:  G29 W I3 J0 Z0.45067
echo:  G29 W I4 J0 Z0.40064
echo:  G29 W I5 J0 Z0.68080
echo:  G29 W I6 J0 Z0.60076
echo:  G29 W I0 J1 Z-0.07962
echo:  G29 W I1 J1 Z0.15051
echo:  G29 W I2 J1 Z0.07046
echo:  G29 W I3 J1 Z0.35062
echo:  G29 W I4 J1 Z0.28058
echo:  G29 W I5 J1 Z0.57074
echo:  G29 W I6 J1 Z0.47068
echo:  G29 W I0 J2 Z-0.15967
echo:  G29 W I1 J2 Z0.07046
echo:  G29 W I2 J2 Z-0.00958
echo:  G29 W I3 J2 Z0.26057
echo:  G29 W I4 J2 Z0.20053
echo:  G29 W I5 J2 Z0.46068
echo:  G29 W I6 J2 Z0.38063
echo:  G29 W I0 J3 Z-0.22971
echo:  G29 W I1 J3 Z0.00042
echo:  G29 W I2 J3 Z-0.06962
echo:  G29 W I3 J3 Z0.20053
echo:  G29 W I4 J3 Z0.10048
echo:  G29 W I5 J3 Z0.39064
echo:  G29 W I6 J3 Z0.29058
echo:  G29 W I0 J4 Z-0.13966
echo:  G29 W I1 J4 Z0.03044
echo:  G29 W I2 J4 Z-0.08963
echo:  G29 W I3 J4 Z0.13049
echo:  G29 W I4 J4 Z0.03044
echo:  G29 W I5 J4 Z0.25056
echo:  G29 W I6 J4 Z0.14050
echo:  G29 W I0 J5 Z-0.14966
echo:  G29 W I1 J5 Z0.00042
echo:  G29 W I2 J5 Z-0.14966
echo:  G29 W I3 J5 Z0.03044
echo:  G29 W I4 J5 Z-0.13966
echo:  G29 W I5 J5 Z0.06046
echo:  G29 W I6 J5 Z-0.09963
echo:  G29 W I0 J6 Z-0.17968
echo:  G29 W I1 J6 Z-0.05961
echo:  G29 W I2 J6 Z-0.23971
echo:  G29 W I3 J6 Z-0.07962
echo:  G29 W I4 J6 Z-0.25972
echo:  G29 W I5 J6 Z-0.07962
echo:  G29 W I6 J6 Z-0.23971
echo:; PID settings:
echo:  M301 P65.03 I11.95 D88.45
echo:  M304 P39.30 I7.71 D133.50
echo:; Z-Probe Offset (mm):
echo:  M851 X5.40 Y-34.20 Z-0.76
echo:; Stepper driver current:
echo:  M906 X700 Y800 Z800
echo:  M906 I1 Z800
echo:  M906 T0 E600
echo:; Driver stepping mode:
echo:  M569 S1 X Y Z
echo:  M569 S1 I1 Z
echo:  M569 S1 T0 E

digant73:
@discip thanks!

discip:
@digant73 
These values are based on the latest commit!
### You are welcome!

digant73:
> @digant73
> These values are based on the latest commit!
> 
> ### You are welcome!

yes it's ok. tomorrow please try the not working version. Move on the BL menu->ABL menu but do not start any calibration. then move back to main menu and try to move again in the BL->ABL menu. Let me know if you see the BL->ABL menu or it disappeared

discip:
@digant73 
Sorry, but could you please be *more specific* on the steps you would like me to make?
Are you talking about the ***non working*** commit?
What exactly do I have to do?

digant73:
@discip 
thanks. You have the same printer I mine. Can you send me the source code of the marlin version you are using? I will try to replicate the problem
Also can you confirm that you had the same problem (after calibration the BL menu was not present) whit the current BTT main branch (not this PR)?

@guruathwal 
1) I was trying to test your PR but the installation stucks during config loading. The progress bar reaches 100% but then it freezes
2) In the code I would define the new macro using capital letters (speaking about init_label, labelChar etc...)
3) In the code there are textSelect(....). Should all of them be removed or handled with the new macros?

discip:
@digant73
Here you go.
1. [Marlin-bugfix-2.0.x.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5346467/Marlin-bugfix-2.0.x.zip)
2. loading ...

digant73:
> @digant73
> Here you go:
> 
> 1. [Marlin-bugfix-2.0.x.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5346467/Marlin-bugfix-2.0.x.zip)
>    1.1. Maybe this helps [config.ini.txt](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5346857/config.ini.txt). man_shrugging
> 2. Yes, I confirm the behavior you described!

if you use the MKS TFT28 in your SWX1 try that TFT FW and let me know please if you have the same issues.

[tft.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5349621/tft.zip)

discip:
@digant73 
Good evening,
1. I have flashed the **TFT FW**, which you provided and I had the same issue, regarding the *disappearing of the ABL menu* after calibration.
2. One more thing appeared right after the successful flash:
The bootlogo froze with a *toast* stating: **`echo:Soft endstops: ON Min: X0.0`**, but after removing the SD and resetting, it started as expected and on some occasions (I think whenever I saved to EEPROM.) the same *toast* appeared.

By the way: Nice *'mesh inspector'* you coded there. 👍
And at some point, you will have to tell me, how to relocate certain icons. 😁

digant73:
@discip 
thanks for the check. The soft endstop toast is now suppressed in my next release. However. If you could repeat the test getting all the output collected by pronterface it could help to finally solve the problem.
I mean, connect pronterfare (with verbose flag enabled) to the printer and then make the ABL process, you should see all the outupt exchanged with Marlin

discip:
@digant73
Sorry, I missed your reply, was dealing with Tasmota.

1. I do not know where to enable verbose.
2. Do you mean repeating all the steps I did [here](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1138#issuecomment-705191076), with *'ABL process'*?
3. I am not sure what you meant with the sentence starting with *ENABLE_BL_VALUE*.
4. I thought you did rearrange the icons in the ***move menu***. I knew about replacing ***\*.bmp**s*.

digant73:
@discip tomorrow I will send the last release with minor changes. I simply need you repeat ABL with pronterface/octoprint connected to the printer and monitoring the process.I will analize the log you will provide from pronterface with the log on my printer

discip:
@digant73 
Good afternoon sir.
1. I still do not know how to activate *verbose* for pronterface, like you asked me to do.
2. What do you mean with repeat ABL?
***=>*** Send ***M503*** and ***M115*** before and after calibration?

digant73:
> @digant73
> Good afternoon sir.
> 
> 1. I still do not know how to activate _verbose_ for pronterface, like you asked me to do.
> 2. What do you mean with repeat ABL?
>    _**=>**_ Send _**M503**_ and _**M115**_ _before_ and _after_ calibration?

forget the verbose flag. Simply connect with pronterface (as you did when you sent M503 and M115) and leave it connected.
Then start ABL process. On the pronterface terminal you will see the dialog between TFT and Marlin. 
At the end of the ABL process simply collect the output produced on pronterface terminal and send it to me. I will compare it with the output produced in my printer.

discip:
@digant73 
I flashed your version, connected the printer to pronterface and started the ABL process, but as **always**, *in my case*, there was ***no dialog between pronterface and the printer***, so no data to provide.

What do I need to do, to make them chat?

digant73:
> @digant73
> I flashed your version, connected the printer to pronterface and started the ABL process, but as **always**, _in my case_, there was _**no dialog between pronterface and the printer**_, so no data to provide.
> 
> What do I need to do, to make them chat?

if you have SImplify3D, use its pronterface. Close to "baud rate" settings, a verbose checkbox  is present. Just enable it.

discip:
No unfortunately, I do not have Simplify3D.
What could I do else?

digant73:
hmm, nothing more. however no problem. we will find the root cause

discip:
Sorry, I would really like to help. 😩

digant73:
> Sorry, I would really like to help.

no problem :-)

Guilouz:
Possible to fix also M177 message in parceACK like in this post ?

https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/1050#issuecomment-687325623

digant73:
just some bugs and improvements.

**bugs:**
1) as I already explained, you ow unified the level state with bl type with the unique variable infoMachineSettings.leveling.
This simply doesn't work. At startup M115 is called after M503 and the parseAck() will always set infoMachineSettings.leveling to the value of Cap:AUTOLEVEL (that can assume only 0 and 1 as value). Also in case of MBL, Cap:AUTOLEVEL is set to 0. This means that in the best case you will see only the ABL menu (even if M503 provided UBL) or no BL menu at all (if MBL is enabled in Marlin). You should at least avoid to set infoMachineSettings.leveling inside Cap:AUTOLEVEL. But in that case you will simply restore the previous existing logic that was logically correct but (most probably for some bugs elsewhere) causing for some users the missing BL menu. It should be better to discover that bug.
2) At the end of the "Updating configuration" procedure (with the config.ini file), when the end of the process is reached (progress bar is at 100%) the system freezes. I had a look to the code and (in my case) it freezes on config.c at line 65. That means when "storePara();" is called.

**improvements (to avoid possible causes of random bugs):**
1) in settings.c some parameters were not initialized (such as rotate_ui whos deafult value is compared with the value from config.ini file (line 570 on config.c). This means a random handling calibration can be performed (on line 60 of config.c)
2) on config.ini file, on "Power Supply Settings" section, the parameter "ps_on_active_high" is defined but in config.c the value read form file is stored into parameter infoSettings.powerloss_invert.powerloss_invert that doesn't have an equivalent setting in config.ini (it should be on "Power Loss Recovery" section). In config.c, it should be used "infoSettings.ps_active_high" instead of "infoSettings.powerloss_invert.powerloss_invert". This last parameter should be removed from settings.h and settings.c.
3) also, the list of paramters in settings.h and settings.c should be provided according to the sequence in config.ini (that is already aligned with config.c). This will make more easy a cross check in the project. The same should be made with flashStore.c (that is currently not aligned with config.c and config.ini. Attached I provided a reviewed version of settings.c and .h (including your last update on laguage paramter). It is aligned with the sequence in config.ini and config.c with the exception of all the array that are initialized at the end.

[settings.c_h.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5359351/settings.c_h.zip)

guruathwal:
> 1. as I already explained, you now unified the level state with bl type using the unique variable infoMachineSettings.leveling.
>    This simply doesn't work. At startup M115 is called after M503 and the parseAck() will always set infoMachineSettings.leveling to the value of Cap:AUTOLEVEL (that can assume only 0 and 1 as value). Also in case of MBL, Cap:AUTOLEVEL is set to 0. This means that in the best case you will see only the ABL menu (even if M503 provided UBL) or no BL menu at all (if MBL is enabled in Marlin). You should at least avoid to set infoMachineSettings.leveling inside Cap:AUTOLEVEL. But in that case you will simply restore the previous existing logic that was logically correct but (most probably for some bugs elsewhere) causing, for some users, the missing BL menu. It should be better to discover that bug.

This can be fixed.

> 2. At the end of the "Updating configuration" procedure (with the config.ini file), when the end of the process is reached (progress bar is at 100%) the system freezes. I had a look to the code and (in my case) it freezes on config.c at line 65. That means when "storePara();" is called.

I am unable to reproduce this as it is working as expected in two different devices I have.

> 1. in settings.c some parameters were not initialized (such as rotate_ui whos deafult value is compared with the value from config.ini file (line 570 on config.c). This means a random handling calibration can be performed (on line 60 of config.c)

It is already initialized in flashStore.c and should never be initialized anywhere else. The config is read after loading the settings.

> 2. on config.ini file, on "Power Supply Settings" section, the parameter "ps_on_active_high" is defined but in config.c the value read form file is stored into the parameter infoSettings.powerloss_invert that doesn't have an equivalent setting in config.ini (it should be on "Power Loss Recovery" section). In config.c, it should be used "infoSettings.ps_active_high" instead of "infoSettings.powerloss_invert.powerloss_invert". This last parameter should be removed from settings.h and settings.c.

This feature is related to switching on/off the power supply/printer. It is independent of the `Power Loss Recovery` feature and should not be grouped with it. In the source the word `invert` make more sense to invert the logic of the said variable. Tha name in the config file is so to use easy to understand words as there are many users who does not know proper english.

digant73:
> > 1. as I already explained, you now unified the level state with bl type using the unique variable infoMachineSettings.leveling.
> >    This simply doesn't work. At startup M115 is called after M503 and the parseAck() will always set infoMachineSettings.leveling to the value of Cap:AUTOLEVEL (that can assume only 0 and 1 as value). Also in case of MBL, Cap:AUTOLEVEL is set to 0. This means that in the best case you will see only the ABL menu (even if M503 provided UBL) or no BL menu at all (if MBL is enabled in Marlin). You should at least avoid to set infoMachineSettings.leveling inside Cap:AUTOLEVEL. But in that case you will simply restore the previous existing logic that was logically correct but (most probably for some bugs elsewhere) causing, for some users, the missing BL menu. It should be better to discover that bug.
> 
> This can be fixed.
yes. I saw the easy fix (I commented out the entire row). However, the same solution was not working for someone although the correct strings were provided by M503. For someone the BL menu was never displayed. For other (e.g. discip) the BL menu disappeared after calibration. That should never happens, because in parseAck() we can set only to ABL, UBL and MBL (not to DISABLED or the previous UNKNOWN value)
> 
> > 1. At the end of the "Updating configuration" procedure (with the config.ini file), when the end of the process is reached (progress bar is at 100%) the system freezes. I had a look to the code and (in my case) it freezes on config.c at line 65. That means when "storePara();" is called.
> 
> I am unable to reproduce this as it is working as expected in two different devices I have.
Ok, unfortunately in my case it is sistematic. Maybe a temporization problem with other HW initialized or that needs to be initialized
> 
> > 1. in settings.c some parameters were not initialized (such as rotate_ui whos deafult value is compared with the value from config.ini file (line 570 on config.c). This means a random handling calibration can be performed (on line 60 of config.c)
> 
> It is already initialized in flashStore.c and should never be initialized anywhere else. The config is read after loading the settings.
Ok. I was thinking to the case were it was not yet present on the flash (a new parameter in the config.ini) and also not initialized in settings.c
> 
> > 1. on config.ini file, on "Power Supply Settings" section, the parameter "ps_on_active_high" is defined but in config.c the value read form file is stored into the parameter infoSettings.powerloss_invert that doesn't have an equivalent setting in config.ini (it should be on "Power Loss Recovery" section). In config.c, it should be used "infoSettings.ps_active_high" instead of "infoSettings.powerloss_invert.powerloss_invert". This last parameter should be removed from settings.h and settings.c.
> 
> This feature is related to switching on/off the power supply/printer. It is independent of the `Power Loss Recovery` feature and should not be grouped with it. In the source the word `invert` make more sense to invert the logic of the said variable. Tha name in the config file is so to use easy to understand words as there are many users who does not know proper english.

I see you merged my settings.c, .h. Just to align everything in the project, I would also align flashStore.c to Config and settings. Just a cleanup, not a bug.
Finally, if possible I would avoid to have so much languages resources in the project (currently there are 3 different language resources. 2  identical language packs, plus the built in language pack). I would provide only one language pack in the project
(that based on the new .ini files)

radek8:
@guruathwal If you are already redesigning the system like this, take a look at setting the number of extruders. In the Configuration.h and config.ini file, I can set the number of extruders, but the parameter is overwritten by the detected value found from Marlin.
If I have a mixing hotend, Marlin reports only one extruder, and the value 2 I entered is overwritten to 1. In the current setting, manual setting of the number of extruders is not applicable.
I tentatively suggested patch #1123.
It would be better to correct the autodetection logic if 0 was entered in the number of extruders, then autodetection would be used, and if it was incorrectly detected, the user would be able to set his own value, which would take precedence over autodetection.

radek8:
> fix wrong extruder count for mixing hotend

This adjustment is not good.
If I have a two-color printer with one hotend and 2 extruders, with filament retraction, Marlin reports 2 extruders correctly, but after this adjustment, the TFT will incorrectly detect 2 hotends.

guruathwal:
> > fix wrong extruder count for mixing hotend
> 
> This adjustment is not good.
> If I have a two-color printer with one hotend and 2 extruders, with filament retraction, Marlin reports 2 extruders correctly, but after this adjustment, the TFT will incorrectly detect 2 hotends.

@radek8 Iooked into the marlin firmware and the extruder count is actual extruder count setup up by the user and is not calculated/derived from anything. I do not have this type of setup, so I need more time to resolve it.

radek8:
OK, leave it as it is. It's functional.
Only printers with a mixing hotend will enable the option in PR #1123

radek8:
Compilation is an issue added by this PR


TFT\src\User\API\LCD_Encoder.c: In function 'HW_EncActiveSignalInit':
TFT\src\User\API\LCD_Encoder.c:28:5: warning: implicit declaration of function 'setEncActiveSignal'; did you mean 'HW_EncActiveSignalInit'? [-Wimpli
     setEncActiveSignal(1);
     ^~~~~~~~~~~~~~~~~~
     HW_EncActiveSignalInit
TFT\src\User\API\LCD_Encoder.c: At top level:
TFT\src\User\API\LCD_Encoder.c:31:8: warning: conflicting types for 'setEncActiveSignal'
   void setEncActiveSignal(uint8_t status)
        ^~~~~~~~~~~~~~~~~~
TFT\src\User\API\LCD_Encoder.c:28:5: note: previous implicit declaration of 'setEncActiveSignal' was here
     setEncActiveSignal(1);
     ^~~~~~~~~~~~~~~~~~

image

Guilouz:
I have alert too but compiling is ok

Capture d’écran 2020-10-19 à 23 48 34

GregSKR:
Just wondering (maybe this is not important)...
In User/API/menu.c, lines 272, 486 and 511:
`rapid_serial_comm()//perform backend printing loop between drawing icons to avoid printer idling`
Isn't a semicolon missing after rapid_serial_comm()?
`rapid_serial_comm(); //perform backend printing loop between drawing icons to avoid printer idling`

PR #1149, opened 2020-10-03
make language set at compile time save ~72kb flash

BlackDragon2020:
### Description
The changes makes the user have to set the language in the setlanguage.h. removes the menu for changing language. The code only #defines one language and it's assets, omitting all others. This frees up 28% Of flash or 72.2kb.

Spi flash has about 500kb left of 7Mb, maybe can move there, but do we need all the bloated language files?

I read btt was thinking about doing this, here is a start, could free up some more read the setlanguage.h

### Benefit

Flash was basically full. All the languages used 28% of the flash or 72.2kb. this gives you that flash space back, but makes language set at compile time.

Compile should be faster based on the way #elif works.

I have not coded in years, this was basically me familiarizing myself with the code and knocking the dust of my boots
 Feel free to criticize. Thanks for your time.

### DIFF
--- a/TFT/src/User/API/Language/Language.c
+++ b/TFT/src/User/API/Language/Language.c
@@ -1,173 +1,315 @@
+#define LANGUAGE_CHECK
 #include "Language.h"
 #include "includes.h"
+
+#if LANGUAGE_CHECK==0
 #include "language_en.h"
+#elif LANGUAGE_CHECK==1
 #include "language_cn.h"
+#elif LANGUAGE_CHECK==2
 #include "language_ru.h"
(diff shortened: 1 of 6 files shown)


ETE-Design:
Looks like what's already merged in to #1138

blueeagle69:
Just tested this.
It does make a big difference to the binary size!
Excellent.

oldman4U:
Hi wild dragon. 

You can close this PR. 

Thank you

bigtreetech:
Hello, this has be fixed in other PR,  so I will close here.  Thanks for your hard work

Issue #1112, opened 2020-09-20
Ability to Scroll GCode Command Display Window


ptmy1:
**Is your feature request related to a problem? Please describe.**
After issuing a GCode command using Terminal in the TFT35 touchscreen, a display window is shown.  This window shows the GCode command issued and response received from the printer firmware.  Information often appears and disappears too quickly to be read.  

**Describe the solution you'd like**
The ability to scroll back the display window.  Also, the ability to switch back and forth between the command window and display window.  Currently, if I accidentally touch the display window, system goes back to showing the command window.  The only way to show the display window again is to issue another GCode command.

**Describe alternative used**
Send GCode commands using OctoPi

radek8:
PR #1171

oldman4U:
Hello ptmy1.

Could you please check PR #1171 like radek8 mentioned 5 days ago, if this covers your feature request.

Could you please be so kind and close this ticket after that. Your feature request is alreadycovered in the pinned Feature Request ticket #1177 

Thank you

ptmy1:
I don’t understand this.  #1177 looks like a bug fix unrelated to my feature request.

ptmy1:
I don't understand.  Is my enhancement request a duplicate of a request posted earlier?

oldman4U:
This was a typo. Sorry!

Please check #1170 but for your fr already a PR exist, right?

ptmy1 <notifications@github.com> schrieb am Sa. 24. Okt. 2020 um 05:33:

oldman4U:
And #1171 includes what you have been asking for. So if you compile it you get the feature. Please test 1171 and if it does what you want you can close this ticket. If not let the owner of 1171 know.

oldman4U:
Any news?

oldman4U:
And now it is in master branch. 

Please help the community and close the ticket once you do not need it anymore. Thank you

PR #1171, opened 2020-10-13
Supports Scroll gcode messages in terminal

Subcircuit:
### Description
On the gcode message display interface of the terminal, touch the upper half screen to turn pages up and touch the lower half screen to turn pages down.
```
screen:
        exit botton
+------------------------+
|                        |
|        page up         |
|                        |
+------------------------+
|                        |
|        page down       |
|                        |
+------------------------+

```
### Related Issues
TFT35 E3 , TFT 43 , TFT 50 FLASH overflowed about 300 bytes

### DIFF
--- a/TFT/src/User/API/Settings.c
+++ b/TFT/src/User/API/Settings.c
@@ -169,6 +169,11 @@ void initMachineSetting(void){
 
 void setupMachine(void)
 {
+  // Avoid repeated calls caused by manually sending M115 in terminal menu
+  static bool firstCall = true;
+  if (!firstCall) return;
+  firstCall = false;
+
   #ifdef ENABLE_BL_VALUE
     #if ENABLE_BL_VALUE == 2
         infoMachineSettings.leveling = BL_ABL;
--- a/TFT/src/User/Menu/SendGcode.c
+++ b/TFT/src/User/Menu/SendGcode.c
@@ -1,6 +1,13 @@
 #include "SendGcode.h"
 #include "includes.h"
 
+#define SCROLL_LINE  22
+#define SCROLL_PAGE  1
+char terminalBuf[TERMINAL_MAX_CHAR];
+TERMINAL_PAGE terminal_page = {terminalBuf, 0, 0, 0, 0, 0};
+static uint16_t terminalBufTail = 0;
+uint8_t buf_full = 0;
+
 typedef enum
 {
   SOFT_KEY_123 = 0,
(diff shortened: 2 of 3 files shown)


radek8:
How does the terminal window close?

Subcircuit:
> How does the terminal window close?

Page turning is under the back button, and the back button is still there.

radek8:
OK, I missed the button. Thank you

StarNamer:
@Subcircuit I just submitted a [PR ](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1191)to add an extra column to the terminal keys. Originally (#1181 ), I was just going to add the F key (as I find typing the F parameter common as in 'G0 F60 Z0') but decided to add an extra column give a better layout (X Y Z E on one row) and allow a few extra keys (I, J, P, R). The only only place this conflicts with what you've done is that you've defined the bottom right position as ">". But as far as I can see, you don't use it for anything and I'm not aware of GCode using it. Is there a reason for this key being added? I didn't include it in my change, but since I added several punctuation keys on the ABC page, this could easily replace one of them since the only essential punctuation characters are "!" and "#" (for M32), "/" for file paths and "~" for 8.3 truncated filenames.

oldman4U:
Hello @Subcircuit.

Hope you are still monitoring this PR. Could you please add this to this PR, so that the FR related to the function you developed is automatically closed.

Thank you

resolves #1112

digant73 (changes requested):
hi,
As an enhancement I will add a page/total pages indication in the same bar where you provided the "back" button. It helps the navigation through all the pages

BlackDragon2020:
I will add that this was something that made me sad as using this function was one of the reasons to get it, but with no scrolling it was maybe 75% usable at best.

this was one of the bugs/ issues / FR i was going to look in to or recode. I will give this a look soon. thanks

digant73:
IMHO once the bar providing the "back" button is present, I would place into it also the UP and DOWN buttons. It makes more intuitive the navigation (instead of pressing on the upper or lower part of the screen. I wouls also place the bar in the bottom part of the screen instead of the upper part (that simply to avoid to cover with the finger the screen). I would provide something like:

+------------------------+
|                        |
|                        |
|                        |
|                        |
|                        |
|                        |
+------------------------+
| _1/4______^__v__BACK_|

bigtreetech:
@digant73 hello,  I've rearranged this menu.

digant73:
> @digant73 hello, I've rearranged this menu.

Ok, I will have a look.
Something more I would like to change is also the background and foreground colors. Although I don't have problem with my eyes, the grey background and the black foreground are not the best choice (too less contrast). Normally the preferred colors are black for the background and yellow (or orange) for the foreground. However, I could work on these minor improvements in the future (after this PR is merged).
My idea is also to review the main page. The "send" and "back" buttons should be located in the bottom part. They are more compliant with the standard menus (where the back button is in the bottom part of the screen). Also it makes (IMHO) more comfortable the use of the buttons. Also that blue and yellow colors seems not consistent with the rest of the layout. Just my opinion

PR #1202, opened 2020-10-25
Updated Hungaryan Language

AntoszHUN:
Newly added options have been updated. So everything can be read in Hungarian. I have corrected the incorrectness of the conjugation of a few more words.

By: AntoszHUN


radek8:
Now you have to translate the INI file with Hungarian

AntoszHUN:
> Now you also have to translate the INI file with Hungarian

What? What INI file? So far I haven't had to. What kind of nonsense got into the firmware again?

radek8:
16 languages already took up 1/3 of the space and the storage was completely full. No more features could be added.
Therefore, there will now be only the primary language (English) + one selected language in the FW, which you will upload via the INI file.

AntoszHUN:
> 
> 
> 16 languages already took up 1/3 of the space and the storage was completely full. No more features could be added.
> Therefore, there will now be only the primary language (default English) + one selected language in the FW, which you will upload via the INI file.

So it's understandable. But you don’t have to complicate it to run out of space. It has to be left as is and only the bugs need to be fixed. As with a good program. It is then annually updated and ready. Not possible?

radek8:
the situation was so critical that even errors could not be corrected

radek8:
The second possible solution was to delete only the Hungarian :-)

AntoszHUN:
Okay okay. I Understand.

radek8:
You didn't have to close PR just to add another file edit

PR #1205, opened 2020-10-25
It was missed

AntoszHUN:
A language_en.ini was left out. There’s so much of it already that I didn’t notice. I don't know why so much. But there must be a reason.

By: AntoszHUN

### DIFF
--- a/TFT/src/User/Language Packs/language_hu.ini
+++ b/TFT/src/User/Language Packs/language_hu.ini
@@ -30,8 +30,8 @@ label_point_2:2. pont
 label_point_3:3. pont
 label_point_4:4. pont
 label_point_5:5. pont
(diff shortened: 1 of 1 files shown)


oldman4U:
You mean why there are so many language files?

AntoszHUN:
> You mean why there are so many language files?

Why do I need 3 locations for INI file?

oldman4U:
One in each Icon theme. Otherwise the user has to jump from one folder to another. But this would also be possible. Then the question is why this has to be the case for font and config.ini also.

AntoszHUN:
> One in each Icon theme. Otherwise the user has to jump from one folder to another. But this would also be possible. Then the question is why this has to be the case for font and config.ini also.

I see. :)

Issue #1194, opened 2020-10-21
[FR] Add option to be able to control case light


mak0t0san:
**Is your feature request related to a problem? Please describe.**
Unable to control case light from touch screen.  Have to go into Marlin mode to be able to change case light.

**Describe the solution you'd like**
Add a button (that is only shown if `caseLightsBrightness` in the struct MACHINESETTINGS is set to 1) that will take the user to a Case light menu. The Case Light menu will have options to increase or decrease the brightness, or to turn it on or off.

**Describe alternatives you've considered**
The current alternative is to go into Marlin mode.

**Additional context**
Case light gcode information is here:
https://marlinfw.org/docs/gcode/M355.html

Sending M355 with no parameters will get the current values.  Here are some examples:
```
Send: M355
Recv: echo:Case light: OFF
Send: M355  
Recv: echo:Case light: 255
```

oldman4U:
Whats wrong with using the existing LED light function?

mak0t0san:
> Whats wrong with using the existing LED light function?

The LED light function controls Neopixels RGB LEDs.  This is different compared to case light.  

For example, I have a strip of LEDs connected to an unused heater output and am able to control the brightness by sending M355 gcode commands.  The TouchScreenFirmware doesn't currently support M355.

I am currently working on an implementation to test it out, but would like to go through an official path so that new icons that match the current theme can be created.

oldman4U:
Got it, thx.

Unfortunately there is nothing like an official path. In fact, this repository is a bit like the wild west where tickets are opened but never closed and where developers are working on the same stuff without knowing from each other. This happened several times and people left frustrated. Now we try to make everything a bit more organised and one part of it is to collect all FR's into a single ticket, and also to show, if an engineer is working on it or not. I already added your ticket to the pinned ticket #1170 and marked it as "under development". It would be great if you could close this ticket, but the same time use it for further discussion. Thank you

For a fast start, I would check if one of the existing Icons can be used. Do you know where you would like to add the control buttons in the UI?

mak0t0san:
Yeah. Planning on adding it to the settings screen, after Connection.
[...]

PR #1214, opened 2020-10-29
Add support to be able to control case light

mak0t0san:
### Requirements

### Description
This PR adds a button in Machine Settings that allows a user to view and modify the brightness of the case lights.  The button is only visible if the capabilities reported by the machine indicates that it has a case light.

### Benefits

This allows a user to easily change the case light brightness, or even toggle it on and off. Without this change, a user has a couple of workarounds, none of them very convenient including:
- Switch to Marlin mode screen
- Go to terminal and type in gcode 
- Create custom actions

### Related Issues
#1194 

Here are some pictures of what the button looks like on a TFT35 and TFT24

TFT35
TFT24

And pictures of the controls


digant73 (changes requested):
hi
In all the german language files there is a duplication for
#define STRING_NOTIFICATIONS          "Notifications"
Remove that entry (leave the german translation). Also I would append the new label at the end of each language file (so after the mesh tuner label).
The same in includes.h. I would append the new .h after the mesh tuner and probeHeightControl

mak0t0san:
> hi
> In all the german language files there is a duplication for
> #define STRING_NOTIFICATIONS "Notifications"
> Remove that entry (leave the german translation). Also I would append the new label at the end of each language file (so after the mesh tuner label).
> The same in includes.h. I would append the new .h after the mesh tuner and probeHeightControl

Thanks. I'll fix those up.  I merged in the latest changes from the upstream branch and it brought the mesh tuner strings to the bottom.  I'll update this PR with those changes.

Acenotass:
hi @mak0t0san 
The backlight on my printer is WS2812 RGB Led Strip Light with a choice of colors and / or effects. It would be nice to include support for such functionality.

mak0t0san:
> hi @mak0t0san
> The backlight on my printer is WS2812 RGB Led Strip Light with a choice of colors and / or effects. It would be nice to include support for such functionality.

@Acenotass , does the current RGB functionality do what you need?  The touchscreen firmware has code to be able to control RGB lights via the M150 gcode command.

Acenotass:
@mak0t0san 
At least backlight color control

Issue #361, opened 2020-03-05
[FR] IDEA: remote firmware deploy with build.py


Andrei-Pozolotin:
@guruathwal 

consider inclusion of `build.py` script to automate firmware build and image deployment:

* tool folder:
[v-user-1/buildroot/user](https://github.com/random-builder/bigtree-tft-code/tree/v-user-1/buildroot/user)

* build entry point:
[v-user-1/buildroot/user/build_USER.py](https://github.com/random-builder/bigtree-tft-code/blob/v-user-1/buildroot/user/build_USER.py)

* platfromio integration:
[v-user-1/buildroot/user/setup.py](https://github.com/random-builder/bigtree-tft-code/blob/v-user-1/buildroot/user/setup.py)

* remote host deploy script:
[v-user-1/buildroot/user/setup_disk.py](https://github.com/random-builder/bigtree-tft-code/blob/v-user-1/buildroot/user/setup_disk.py)

oldman4U:
Could you please be so kind and close this ticket. Your feature request is covered in the pinned Feature Request ticket #1177 

Thank you

oldman4U:
Hi. Hope you are doing fine.

Could you please check if this solves your request: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1463

Thank you and happy new year

PR #1215, opened 2020-10-29
Language pack auto-generate script

guruathwal:
- Add language pack auto-generate script to keep the ini files in sync with the source
  The `language_xx.ini` files are now auto-generated while compiling the firmware and are copied to the respective folder. 
  The String size for each keyword is checked to make sure it is within max limit
- Add language flash size check to prevent compiling if the size of total keywords is larger than allocated size.

### DIFF
--- a/Copy to SD Card root directory to update - Unified Menu Material theme/Language Packs/language_am.ini
+++ b/Copy to SD Card root directory to update - Unified Menu Material theme/Language Packs/language_am.ini
@@ -1,5 +1,5 @@
 #### Language Code:AM
-## Language Version:20201001
+## Language Version:20201007
 ## Maximum byte per keyword is 250 Bytes.
 ## Escape characters are not supported except newline '\n'
 
@@ -277,3 +277,4 @@ label_hybrid_threshold:TMC HybridThreshold
 label_notifications:Notifications
 label_mesh_editor:Mesh edit
 label_mesh_tuner:Mesh tuner
+
--- a/Copy to SD Card root directory to update - Unified Menu Material theme/Language Packs/language_ca.ini
+++ b/Copy to SD Card root directory to update - Unified Menu Material theme/Language Packs/language_ca.ini
@@ -1,5 +1,5 @@
 #### Language Code:CA
-## Language Version:20201001
+## Language Version:20201007
 ## Maximum byte per keyword is 250 Bytes.
 ## Escape characters are not supported except newline '\n'
 
@@ -277,3 +277,4 @@ label_hybrid_threshold:TMC HybridThreshold
 label_notifications:Notifications
 label_mesh_editor:Mesh edit
 label_mesh_tuner:Mesh tuner
+
--- a/Copy to SD Card root directory to update - Unified Menu Material theme/Language Packs/language_cn.ini
+++ b/Copy to SD Card root directory to update - Unified Menu Material theme/Language Packs/language_cn.ini
@@ -1,5 +1,5 @@
 #### Language Code:CN
-## Language Version:20201001
+## Language Version:20201007
 ## Maximum byte per keyword is 250 Bytes.
 ## Escape characters are not supported except newline '\n'
 
--- a/Copy to SD Card root directory to update - Unified Menu Material theme/Language Packs/language_cz.ini
+++ b/Copy to SD Card root directory to update - Unified Menu Material theme/Language Packs/language_cz.ini
@@ -1,5 +1,5 @@
 #### Language Code:CZ
-## Language Version:20201001
+## Language Version:20201007
 ## Maximum byte per keyword is 250 Bytes.
 ## Escape characters are not supported except newline '\n'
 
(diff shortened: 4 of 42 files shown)


digant73:
hi, Are you considering the possibility to reduce (maybe in further PRs) the number of language packs? Currently there are 4 packs with the same information and it is uncomfortable to manage all of those identical transalations.

radek8:
When creating a new function, you should only edit the configuration and language INI files in the "TFT \ scr \ user" directory.

Bigtreetech should copy INI files to the "Copy to SD Card root directory to update" and "Copy to SD Card root directory to update - Unified Menu Material theme" as it compiles new BIN files.
Otherwise, there are problems that the user flashes a month old bin file with the current ini file and the display generates errors with unsupported variables.
Unfortunately, this does not happen and users complain about inconsistent files to update

radek8:
The same problem is with the icons when the old Bin files are in the folder, but the icons already have a new name, some are deleted and so on.

guruathwal:
@digant73 This is an auto-generate script which does not require manual editing the `ini` files. All the `ini` files are now auto-generated from the language Headers files and directly copied to the respective folders for easy access during the build process. You will have to manually add/edit at least the C Headers for the keywords which are required for compiling the firmware also.

@radek8 The issue you are explaining is related to the release branch and dev branch of the firmware. it's not feasible to maintain too many copies of the same file at multiple locations. There are different suggestion fro different users. There should not be any compiled binary files in this Master branch. All the compiled files should be in the release branch only. However, BTT should release a new version at least in every 2nd month.

@discip it is from here: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/8a5b9a1ccb0f3c2d3f6b2d14bfdd2e015a6ea0c8/TFT/src/User/API/Language/language_de.h#L98

digant73:
@guruathwal good. So it's the final solution we were waiting for

discip:
@guruathwal
Good morning,
first of all thank you for the updates. 👍

Some questions about the following:
> @digant73 This is an auto-generate script which does not require manual editing the `ini` files. All the `ini` files are now auto-generated from the language Headers files and directly copied to the respective folders for easy access during the build process. You will have to manually add/edit at least the C Headers for the keywords which are required for compiling the firmware also.

I am a somewhat confused right now, because I do not know anymore:
1. which **\*.ini** to configure according to my liking?
2. which **files** do I have to copy to *SD*?
3. where do I set the **language** (`config.ini` or `Configuration.h`)?

Regarding ***language_de.ini***: Than it has to be corrected in **`Line 98`** in ***language_de.h***. 😊

btw:
The latest commit results in this:

crash

oldman4U:
Hi discip.

Which TFT and which commit exactly, master branch?

Ich würde auch noch das r weglassen;-) "Marlin Simulator Hintergrundfarbe**r**"

digant73:
@guruathwal If I can suggest a possible change, I would also consider to apply the same logic to the config.ini file (possibly providing some placeholders; e.g. to differentiate "unified_menu"). Currently there are 3 instances of config.ini. I would consider only TFT/src/User/config.ini and then copy it on the two release directories (Unified and classic material theme). Probably I would move TFT/src/User/config.ini to a directory called TFT/resources. I would made the same for the languages and icons currently provided on TFT/src/User/API/Language/* and TFT/src/User/API/icon_list.inc. I see those as configurable resources more than C code

discip:
@oldman4U 
1. TFT is *MKS_32_V1_4*.
And the commit is this https://github.com/guruathwal/BIGTREETECH-TouchScreenFirmware/commit/79286d545a69b4c8d45e71097314686d7c1520da one.

2. There is a *PR*, concerning **`language_de.ini`**, on its way already.

kind regards

digant73:
@discip you have also to apply the icons. a new one (apply.bmp) is used by the current BTT main branch

discip:
@digant73 
Ah, did not realize *there* where changes applied as well.
But than again, why did the *system* not ask for that, like **it did in the past**, if that was the case.
thank you

discip:
@digant73 
Unfortunately this did not change anything.

radek8:
Here #1225 is an example of the problem I wrote about above.
There is an old version of the precompiled BIN file and it has a new version of the INI file and missing icons that are deleted from the new version.
The new user is confused.

guruathwal:
@digant73 it is feasible for `config.ini` but not for icons as currently the icons are already at one place, instead I would recommend changing the names of the folders "Copy to SD Card root directory to update"  to a more logical one and adding an instruction readme file. 
The `.inc` files work as c header files. The .inc extension is just used to indicate to not add any other code. These files are used by the pre-processor to generate source code for compiling of the firmware. The language files are already in their correct seperate location.

@radek8 This is a known problem. With so many small PRs mergeing every few days the precompiled files will always be outdated. It is also explained by some 3D printing channels on YouTube. One solution can be to remove the precompiled binaries from this `master` branch and keep them only in release branch.

digant73:
@guruathwal yes, I agree with both the binary removal and the "Copy to SD Card..." directories.
I would move the config.ini and Language Packs under the main directory TFT\src\User\resource. At least all the resources (e.g. languages templates, configuration files and any further new resource) will be placed in a common base directory. That config.ini file under TFT\src\User is clearly not the right place (it's mixed with .c files)

oldman4U:
Andrei just closed #361. Isn't this somehow related and also the tickets related to "pre compiled version" issues?

Issue #967, opened 2020-08-13
Duplicate features and icons


rightawy:
### Description
Since last few PRs there are more duplicate features beeng added.
Like the baby stepping is in side both ABL and Manual Leveling, while it should have been in the move menu.
The EEPROM save restore options now available in 3 places: customenu, in parameter settings, and in the Machine settings where the user needs to juggle multiple times to select correct option.
There are two different icons for disabling steppers and save/load buttons but all of them do the same thing.
The new save and load button are both same floppy disk with just different back color.
In the z offset menu there are 4 toggle buttons, it will be easier to go back rather than clicking 4 times to disable steppers then 4 time again to get to the save button, this could have been done by asking the user to save or not when clicking on the back button.
The 100mm button in tuning menu is not useful as it does not check for heating and the same can be done through extrude menu which also allows to change length (between 1mm and 200mm) and select different speed. The tuning menu should be inside the machine settings.

traffic-light:
This is not really a bug.
But I agree with some points.

traffic-light:
Oh. By the way. I'm working on a new version for the extruder tuning.

digant73:
@rightawy 
The correct location for EEPROM save, reset ... should be only on MachineSettings menu. The correct place, easy to see, easy to use (big icons). Opening a small menu, scrolling at the end and then pressing on a small text could be a problem for a lot of people with handicap on their hands or eyes.
I see that in the last PR, by default that menu is now disabled. It seems that people doesn't have any problem here or prefer the most complicate approach.
The function for the steppers are different. One is to disable all steppers, while the other (used in other menu such as Leveling and Z Offset) disable only X and Y axis.
In the Z offset menu, that is the best layout to provide all those features according to the status of z offset process. Exiting from the menu will always disable z offset process, if not already stopped by the user. Asking to save on exit will not allow you to be notified again in case you denied to save the first time. The menus must be contextual. Now in the manual leveling menu, babystep is no more available while it allowed to fine tune the leveling as final step. Another bad choice.

traffic-light:
Currently working on the tune extruder.
You can find it here:

rightawy:
[...]

PR #1227, opened 2020-11-02
improved Terminal menu

digant73:
**IMPROVEMENTS:**

* Improved Terminal menu:
 1) Changed layout providing a control bar. Configurable colors are used such as:
  * infoSettings.font_color: used for keyboard
  * infoSettings.bg_color: used for keyboard
  * infoSettings.marlin_mode_font_color: used on terminal window
  * infoSettings.marlin_mode_bg_color: used on terminal window
 2) page and total page count provided in the terminal window

**PR STATE:**

IMG_20201102_220422_1
IMG_20201102_220442_1
IMG_20201102_220505_1


oldman4U:
Wouldn"t it be good to release it together with #1191?

digant73:
@oldman4U no problem. If the owner of that PR agrees, I can merge his changes in my PR.

StarNamer:
@digant73 Your changes look like what I was think of moving on to if you hadn't already done it and hopefully will be a significant improvement. Moving the non-printing buttons to their own group makes the layout much cleaner.

My changes were basically to include a variation of the number of columns (in ui_draw.h) to better make use of the space on the TFT35 while not making the key areas too small on the TFT24/28. For ease of editing, I moved the array definitions from SendGCode.h to SendGCode.c and, for ease of future layout changes, eventually decided to simply replicate the entire arrays for each variant rather than try to dynamically construct them by macros or multiple tests. I'm sure you'd be able to adopt them easily.

It looks like you've also picked up changes for scrolling  the output similar to [PR 1171](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1171) which I'm looking forward to using.

By the way, I'm still puzzled by the inclusion of ">" on the 123 page since I haven't identified a use for it. What am I missing?

Looks like a great job and I hope to download your changes and take a look later in week.

StarNamer:
Decided I couldn't wait to look at your code. I see, you've moved the arrays/enums into SendGCode.c, the same as I did.

If the TFT24/28 can handle the extra row you've added, then it should work for both the TFT24/28 and the TFT35.

If it can't, you could do something like:
`"1", "2", "3", "4", "M", "G",
`
`"5", "6", "7", "8", "X", "Y",
`
`"9", "0", ".", "-", "E", "F",
`
`"ABC", "Space", "Del", "Back"
`
With something similar on the ABC page.

If 5 rows are OK on the TFT35, then additional keys could be added...
`"1", "2", "3", "M", "G", "S", "T",
`
`"4", "5", "6", "X", "Y", "Z", "E",
`
`"7", "8", "9", "I", "J", "P", "R",
`
`".", "0", "-", "/", "~", "!", "#",
`
`"ABC", "Space", "Del", "Back"
`

I'll try to find time to try out a few ideas in the next few days, based on your code.

StarNamer:
First attempt. Need a few tweaks, but it's 1am!

oldman4U:
It is 3am here reading your posts. I am excited 😆

digant73:
@StarNamer My PR is on top of the merged #1171. Simply provide me your layouts here and I will merge them on my PR

rightawy:
What happened to the concern for people with handicapped and large fingers? https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/967#issuecomment-673647055

oldman4U:
@rightawy 

As far as I understand, StarNamer provides the interface depending on the size of the screen, so it should work like before for the smaller screens. Bigger screens can hold more elements (even the resolution is sometimes lower compared to the TFT35), so it makes sense to allow more elements for those screens (or have bigger buttons).

In general I hope that people like me having big hands do not expect the same functionality independent of the size of the screen. This would be like buying a small car but expecting the comfort of a big car. Somehow this does not match.

My 2 cents like always;-)

bigtreetech:
@StarNamer Hello,  Now I merge this PR, you can make new resolution adaptation changes in #1191, and I will merge #1191 after your optimization

PR #1241, opened 2020-11-10
Disable stepper motor deactivation timeout for M600/M601

azsde:
On Marlin FW, there's a feature to handle stepper motors idling, if they have been inactive for a specified period of time (usually 120 seconds, see DEFAULT_STEPPER_DEACTIVE_TIME in Marlin's Configuration_adv.h file), they are turned off.

This can be problematic when a print is paused, because at filament insertion time, bed or print head can be moved. Even the slightest move will result in layer shifting when the print is resumed.

To prevent this, Marlin implemented the ADVANCED_PAUSE_FEATURE with PAUSE_PARK_NO_STEPPER_TIMEOUT.

When a M600 is detected, it disable the stepper timeout during filament change.

But since this TFT FW override M600 command, all this logic never happens on Marlin's side, so the motors are getting timeouts and are disabled.

To prevent this, when a M600 is detected, we need to send a "M84 S0" command which disable the timeout.

Ideally it would be best to set back the initial value when resuming the print, but I didn't find a way to get the initial value.


azsde:
Sorry did not want to include every commit ...

PR #1243, opened 2020-11-11
Avoid marking coordinate as unknown when parsing a M18/M84 S<timeout>

azsde:
### Requirements

None

### Description

M18/M84 commands are used to disable steppers, however they also are used to set the stepper inactivity timeout:

    M18 S60; Set the stepper inactivity timeout to 1 minute

When only setting the timeout, steppers are not disabled. 

Thus ,if the print is paused, the printhead would not move to a safe spot as the coordinates would be considered as unknown.

To fix this, I added a simple test, in the case a M18/M84 command is seen, it checks for the arguments and if only S is specified, it will not mark coordinates as unknown as it is not needed.

### Benefits

In the eventuality a M18/M84 S<timeout value> was sent during a print, the next pause command would not move the print head to a safe spot.

This commit fixes this behavior.

### DIFF
--- a/TFT/src/User/API/interfaceCmd.c
+++ b/TFT/src/User/API/interfaceCmd.c
@@ -235,7 +235,17 @@ void sendQueueCmd(void)
           break;
         case 18: //M18/M84 disable steppers
         case 84:
-          coordinateSetKnown(false);
+          if(cmd_seen('S') && !cmd_seen('Y') && !cmd_seen('Z') && !cmd_seen('E'))
+          {
+            // Do not mark coordinate as unknown in this case as this is a M18/M84 S<timeout>
+            // command that doesn't disable the motors right away but will set their idling
+            // timeout.
+          }
+          else
+          {
+            // This is something else than an "M18/M84 S<timeout>", this will disable at least one stepper, set coordinate as unknown
+            coordinateSetKnown(false);
+          }
           break;
 
 #ifdef SERIAL_PORT_2


digant73:
can you provide the same logic also for the M600 (starting on line 752 of interfaceCmd.c)?

azsde:
I'm not sure what you mean ?

Do you want me to add a `#ifdef NOZZLE_PAUSE_M600_M601` to this modification ?

If so, I do not find it useful, there could be other scenarios unrelated to M600 / M601 where a M84 S0 could be sent.

digant73:
no, I mean you should also provide mustStoreCmd("M84 S0\n") for M600 an M601 as reported below (in bold the addition) to avoid stepper are disengaged during pause initiated by M600/601

        #ifdef NOZZLE_PAUSE_M600_M601
          case 600: //M600/M601 pause print
          case 601:
            if (isPrinting())
            {
              **mustStoreCmd("M84 S0\n");**
              setPrintPause(true, false);
              // prevent sending M600/M601 to marlin
              purgeLastCmd();
              return;
            }
            break;
        #endif

azsde:
This is what I did on RAWR's firmware.

I'll make another commit for this one.

It is worth to note that there seems to be no way of getting the original value from Marlin, so once we disable the stepper timeout with a M84 S0, there is no way of setting it back to its original value.

digant73:
@azsde I know it is for RAWR (you told me that :-)). however, in this case it's better to not apply that change. As far as I remember the pause and resume are correctly applied with M600

azsde:
@digant73 I disagree with you, it is important to avoid marking coordinate as unknown when parsing a M18/M84 S<timeout>, this is unrelated to the M600/M601 commands.

Let's say I slice a model and I want to put a M18 S60 for some reason in my gcode, if a pause occurs it will not move the printhead to a safe spot.

Regarding the M600/M601 commands, if the configuration NOZZLE_PAUSE_M600_M601 is set, the command will not be parsed by Marlin and will require a M84 S0 command to avoid steppers getting released after a certain timeout.

So, there are two options:

- NOZZLE_PAUSE_M600_M601  is set, `mustStoreCmd("M84 S0\n");` is required
- NOZZLE_PAUSE_M600_M601  is not set, nothing needs to be done regarding the stepper timeout.

But, in any case, it is mandatory to not mark the coordinate as unknown when parsing a M18/M84 S<timeout>.

digant73:
I said to avoid the change for M600 only

azsde:
Oh my bad, well it is always a nice to have feature, I can always do a PR and it will be up to the repository owner to decide.

PR #1261, opened 2020-11-17
Update language_ru.h

Acenotass:
Update language_ru.h


guruathwal:
@Acenotass please try to club all similar changes in one PR instead of pushing PR for every file. (#1262 #1263)

PR #1262, opened 2020-11-17
Update language_ru.ini

Acenotass:
Update language_ru.ini


bigtreetech:
merged in #1261 so close here

PR #1276, opened 2020-11-21
Enlarged DMA-L1 and Terminal buffer, reduced DMA-L2 buffer

kisslorand:
### Requirements

BTT TFT or MKS TFT

### Description

- enlarged DMA-L1 buffer
- enlarged Terminal buffer
- reduced DMA-L2 buffer to accommodate RAM requirements across the boards, it was unnecessarily large anyway

### Benefits

TFT can display larger responses from Marlin.
### Related Issues

Fixes bigtreetech#1267


radek8:
It works. Thank you

oldman4U:
Great work. 

Thank you

digant73:
I would maintain the same buffer sizes for all the TFT models. Otherwise it will provide different behaviours for the different TFT.
Also, RAM is enough for all models, so there is no problem with that

oldman4U:
I am not an expert, but this will also then work with the TFT24 which is known to have limited specs?

digant73:
yes, no problem on any TFT. just 512 more bytes make no difference when you have about of 48% used RAM

kisslorand:
> I would maintain the same buffer sizes for all the TFT models. 

Me too

> Also, RAM is enough for all models, so there is no problem with that

Try to compile at the higher settings (DMA_TRANS_LEN and TERMINAL_MAX_CHAR) with MKS_32_V1_4_NOBL, BIGTREE_TFT24_V1_1 or BIGTREE_TFT28_V3_0 and see for yourself.

kisslorand:
> yes, no problem on any TFT. just 512 more bytes make no difference when you have about of 48% used RAM

You need some free RAM for FATFS.

PR #1284, opened 2020-11-22
Layer refresh on the tft delayed during print

### LINKED ISSUE #1278: [BUG] stuttering Print with BIGTREETECH TFT35-E3 V3.0
If printing from TFT a cylinder for testing spiralize mode and the ttf is at start-screen its stuttering. if i'm in a submenu, only about all 3-10 seconds a short stuttering. If the same file at marlin-mode no stuttering. tried
 higher Baud and the fast mode for gcode, but not made it better.
the High-value at z-axis is flickering, too.

Used the BTT TF Cloud (copied on it before directly at the MicroSD with cardreader, cause Wifi not working to transfer with Win10)

A lot of Blobs and Zits at Surface

Same file from same SD Card at Marlinmode plugged in at Mainboard are working without any Problems.

Marlin and TFT Firmware downloaded today evening.
[Btt.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5578575/Btt.zip)

kisslorand:
### Requirements

BTT TFT or MKS TFT.

### Description

Delays the refresh of the layer on TFT during print.
### Benefits

Avoids stuttering (especially on vase mode when it is updated at every program loop).
### Related Issues
Hopefully fixes #1278


guruathwal:
instead of drawing every 500ms, it would be better to draw only if the value changes.

oldman4U:
Hi Gurmeet, we missed you!;-)

The problem is, that there are prints where the layer changes very often which causes issues #1278. As you know, I am not a big fan to let the majority of the users suffer, just to fix edge cases. But the ability to get a proper Z layer information has been highly requested, so the feature itself is great. So I believe we need both, the information on layer change, but also a limit, how often such a change is displayed. Maybe I remember right, that using Marlin with an LCD display, some of the values shown have been updated most of the times, but not every time. So for example in case a Z-hop was activated, this information sometimes was updated but not every time.

Just my 2 cents as a user and not an expert - as usual...

TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -31,7 +31,9 @@ const GUI_RECT printinfo_val_rect[6] = {
 };
 
 static uint32_t nextTime = 0;
+static uint32_t nextDrawTime = 0;
 static uint32_t toggle_time = 2000; // 1 seconds is 1000
+static uint32_t drawTime = 500; // 1 seconds is 1000
```
guruathwal, line 36: use #define for constant values

```suggestion
#define drawTime  500 // 1 seconds is 1000
```

kisslorand: Good idea!

radek8:
> instead of drawing every 500ms, it would be better to draw only if the value changes.

It's not a good idea.
the user prints a single-threaded cylinder. It was cut like a continuous spiral.
There was a change on the Z axis in each step.
The value was constantly updated in such a way that the TFT did not manage to send data to Marlin because it was overloaded with the CPU updating the icon. 
#1278  Watch the attached video

guruathwal:
@oldman4U I am busy with some important work. 
I remember that the solution I am suggesting was already added by me at the time of developing the new printing screen, Which I think in this case is not helping, as the layer changes are too frequent. But instead of using the time delay it might be useful to add a minimum change in layer height to trigger the redraw, like redraw only if change is above 0.1MM. this way it ensure redraw only when change is detected.

@radek8 A few weeks back I printed two vases with a TFT35_V2 and I did not witnessed any stutter or pause. the layer height was changing at a high rate without any stutter. The firmware on it is updated to this commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1215. 
In the videos you mentioned the print is stuttering even when the TFT is not on the printing status screen. also it is interesting to note that in the second video (https://drive.google.com/file/d/19RmHP9KzZJHkLU3jw9eeTq7snzhh5mT0/view), the pauses occur at almost same time interval as in first video (https://drive.google.com/file/d/19hSKkBuUWvKoLZ9B8F7ySwUMvJO7HY-n/view?usp=drivesdk) and the pauses are in sync with change in page counts in terminal screen. 
Another question arises is that why the print does not pauses in terminal when the full screen (which is larger than 10x the area of the layer height icon) is updated about more than 10 times but pauses only with the page count?

kisslorand:
I just tested with and without the layer change delay. I am on Artillery SWX1, SKR1.4Turbo motherboard. Stutters badly with no delay and plays nice without any hiccup with the 500ms delay in place. With 500ms we can have best of both worlds, very very fast and accurate layer height display and no stuttering.
I'm happy. :)

TFT/src/User/Menu/PrintingMenu.c
kisslorand: > use #define for constant values

Done.

oldman4U:
👍🏻

kisslorand <notifications@github.com> schrieb am Mo. 23. Nov. 2020 um 19:44:

radek8:
Tested and the repair works well

https://youtu.be/UIMzwTqPtLE

radek8:
> @ radek8 Před několika týdny jsem vytiskl dvě vázy s TFT35_V2 a nebyl jsem svědkem koktání ani pauzy. výška vrstvy se měnila vysokou rychlostí bez koktání. Firmware na něm je aktualizován na tento potvrzení [# 1215](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1215) .
> Ve videích, která jste zmínili, je tisk koktání, i když TFT není na obrazovce stavu tisku. také je zajímavé poznamenat, že v druhém videu ( https://drive.google.com/file/d/19RmHP9KzZJHkLU3jw9eeTq7snzhh5mT0/view ) dochází k pozastavení téměř ve stejném časovém intervalu jako v prvním videu ( [jednotka https: //. google.com/file/d/19hSKkBuUWvKoLZ9B8F7ySwUMvJO7HY-n/view?usp=drivesdk](https://drive.google.com/file/d/19hSKkBuUWvKoLZ9B8F7ySwUMvJO7HY-n/view?usp=drivesdk) ) a pauzy jsou synchronizovány se změnou počtu stránek na obrazovce terminálu.
> Vyvstává další otázka, proč se tisk nezastaví v terminálu, když je celá obrazovka (která je větší než 10x plocha ikony výšky vrstvy) aktualizována více než 10krát, ale pozastaví se pouze s počtem stránek?

Guruathwal, yes, you must be right. Updating an icon with a layer height is not the only problem that causes stuttering. Further analysis of the problem would be needed.

PR #1291, opened 2020-11-27
update firmware and folder structure

bigtreetech:
### Description
* Adjust the structure of the precompiled firmware folder. The modified structure is shown in the following figure
image


oldman4U:
OK, now I have to adjust the ReadMe again...;-)

But of course this is much better!

Is there a way that you automated the building of the firmware? Marlin does a nightly build.

bigtreetech:
> 
> Is there a way that you automated the building of the firmware? Marlin does a nightly build.

I plan to write a .py plug-in, just like the [plug-in](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/buildroot/scripts/auto_gen_language_pack.py) that automatically generate the language pack. After compiling, it will automatically copy. Bin to the precompiled folder. In this way, if the firmware changes, the. Bin will also be updated to the latest

bigtreetech:
@oldman4U And thank you very much for your efforts

oldman4U:
This sounds great.

I am happy I can give back a little bit even I can not develop...

Thank you for your great products

Issue #1303, opened 2020-11-28
[BUG] The new Unload/Load function


kisslorand:
### Description

I opened this bug report to be able to discuss how it should be handled

### Steps to reproduce

1.  Press Extrude in the main menu
2.  Preheat automatically appears
3.  After selecting a preheat value the Load/Unload menu appears

or

1. Select Menu -> Heat
2.  Select "Preheat"
3. Select a a preheat option

**Expected behavior**

Extrude should go to Extrude menu, not preheat and than Load/Unload.
After a regular Preheat selection, it should not jump right away to the Load/Unload menu
In the Load/Unload menu "Back" should get you to the previous menu.
Pressing "Load" or "Unload" should not kick you out from the Load/Unload menu. 

**Actual behavior**

Main->Extrude goes to Preheat menu than jumps to Load/Unload menu
Selecting an option in Preheat menu jumps right to Load/Unload menu. If you selected a wrong preheat option there's no turning back.
In the Load/Unload menu "Back" should go to the previous menu, not to the "Status" menu.
Pressing "Load" or "Unload" exits the menu and brings you to the "Status" menu. After an unload the majority of people would like to load a new filament so they have to navigate back to the Load/Unload menu.
### Additional Information

I agree a preheat is needed before load or unload but Marlin takes care of that (M701, M702). It's true it preheats the nozzle to a fixed value (in my case to 180 degree Celsius). However it is not logical to preheat both the nozzle and the bed for a load or unload.
As for the regular extrude, it is the user's duty to preheat the nozzle, it should be the same for load/unload.
Load/unload feature is a welcome feature but I do not think it has a greater importance than the regular Extrude menu, hence I do not see justified its presence in the main menu in the detriment of the Extrude menu, it being  buried deeper in the menu structure.
Also the "Back" button behaves differently than in any other menu, It takes you beck to the Status Screen. I think buttons should behave he same across the menus, it should **just take you BACK** where you were before.

kisslorand:
PR #1304 is there for this bug.

kisslorand:
I found some new issues since PR #1304 was merged, so there will be a new PR. Stand by...
If Load/Unload procedure was initiated, the user can click around on the TFT. Since the printer is in a waiting state (executing M702/M701), it will freeze the TFT eventually until M701.M702 is finished. This can confuse a regular user.

kisslorand:
PR #1317 is there for the latest issues.

[...]

Issue #1302, opened 2020-11-28
Menus not working properly


slwise:
Menus are not working properly.

Installed the latest firmware for the TFT35 E3 V3.0. The update went beautifully, no errors or warnings. I went to change filament and preheated the nozzle by going to the preheat menu. That worked fine. Once heated, I selected the extruder menu to unload the filament. It took me to the preheat screen. Ok, so I selected the correct preheat setting and it took me to an extruder screen with only the unload, load and nozzle buttons were displayed. When I pressed the unload button the extruder didn't move but the z-axis moved up about 10mm and stopped. I tried reinstalling the firmware with the same results. I also compiled a version myself using PlatformIO with the same results. I went back to a previous release and everything is working as it should so it has to be something in the current release.

kisslorand:
It's not a bug, it's an intended function. Your Z moved up and the printer was waiting to reach the selected preheat temperature. Once that temperature reached, filament would load or unload based on your previous selection.

oldman4U:
Thank you. This and some more bugs have been reported already. Please see the tickets from earlier today.

oldman4U:
@kisslorand 

For all direct extruders this is a bit of a strange behavior. Could you please make the extrude function optional?

kisslorand:
Yes, I can restore its (Extrusion icon) original functionality and make the the "Un/Load" function a separate one in the Move menu, where now it is the Extrusion moved. I could even add a "Preheat" and a "Heat" option there.
Ideas? Opinions?

The "Un/Load" automatic menu must go from the Preheat menu! It pops up right after you select some preheat option.

oldman4U:
I BELIEVE that a separate load/unload would be the right way to do it and this should always include the option to heat the nozzle.

The same time I can not really see the need to (un)load filament after the preheat but it would be great if this could be activated optionally.

My 2 cents

kisslorand:
PR #1304 is up, I hope it suits everyone.

oldman4U:
@slwise 
Do you know how to compile firmware using VSC and PIO? Have you tested kisslorands PR?

Please let us know

Thank you

oldman4U:
The fix is merged in the meantime.

Please download the new firmware with the fix and close this ticket afterwards.

Thank you

slwise:
Yes, I can try it this evening. I have a print running right now. I'll close the ticket then. Thanks to everyone.

oldman4U:
[...]

PR #1304, opened 2020-11-29
Load/Unload fix

kisslorand:
### Requirements

BTT TFT or MKS TFT

### Description

The new Load/Unload feature misbehaved, so it needed some taming.

### Benefits

- Preheat option selection will not lead to Load/Unload menu.
- Load/Unload menu has now a Heat and Preheat button.
- Extrude button/menu gets back to its rightful position in the menu structure.
- New Load/Unload button in Movement menu.

### Related Issues

#1302 and #1303

### Additional info
Pressing "Load" or "Unload" in Load/Unload menu will bring a popup window informing the user that Load/Unload procedure was initiated and it has to wait for the nozzle to be heated to the temperature needed for the Load/Unload  process.

Translations are needed, it brought new labels.


GregSKR:
Nice! It's really better that way!

Just 2 suggestions:

1) adding #ifdef LOAD_UNLOAD_M701_M702 in /User/Menu/MainPage.c 
```
line 17

     #ifdef LOAD_UNLOAD_M701_M702
        {ICON_EXTRUDE,                 LABEL_LOAD_UNLOAD_SHORT},
     #else
        {ICON_EXTRUDE,                 LABEL_EXTRUDE},
     #endif

...

line 61

      case KEY_ICON_6:
      #ifdef LOAD_UNLOAD_M701_M702
         infoMenu.menu[++infoMenu.cur] = menuLoadUnload;
      #else
         infoMenu.menu[++infoMenu.cur] = menuExtrude;
      #endif
         break;
```

2) Please do not include platfomio.ini in your PR

Thanks

kisslorand:
> Nice! It's really better that way!
> 
> Just 2 suggestions:
> 
> 1. adding #ifdef LOAD_UNLOAD_M701_M702 in /User/Menu/MainPage.c

You probably meant "UnifiedMove.c" instead of "MainPage.c". Anyhow, great idea! Done.
 
> 2. Please do not include platfomio.ini in your PR
> 
> Thanks

Done

### Additional work
Pressing "Load" or "Unload" in Load/Unload menu will bring a popup window informing the user that Load/Unload procedure was initiated and it has to wait for the nozzle to be heated to the temperature needed for the Load/Unload  process.

kisslorand:
### More work

- added actual and desired temperature display to the Load/Unload menu so people can see something is actually happening
- when exiting, if heaters are still on, there's a popup window asking if you want to turn off the heaters
- preheat button removed, replaced by CoolDown button  (in the preheat menu there's a big chance people will preheat the bed also, which is absolutely unnecessary for filament load/unload)

oldman4U:
I like the way it works now.

Thank you

PR #1311, opened 2020-11-29
Fixed false SFS runout error when G92 E0 is used in start code before…

looxonline:
… any extrude commands are issued.

### Description

When using the smart filament sensor it is possible to have a G92 E0 in the start g-code from the slicer which would be executed potentially prior to any G1 or G0 commands instructing the the extruder to load filament and thereby causing the encoder wheel to turn. This results in the "isAlive" variable within FIL_SmartRunoutDetect being set to false when the G92 command is executed. After the G92 is executed the next update of M114 will see the E count dropping to 0 which could likely cause a difference greater than the required difference for a filament runout. If "isActive" is false then this will cause a runout.

This change causes "isActive" to be set to true for the start of a print. This results in the G92 E0 command triggering the algorithm but not generating a false alert. If no G92 E0 is used then this is still no issue since after the configured amount of filament is extruded the algorithm will once again set "isActive" to false.

### Benefits

No false filament runout when using the smart filament sensor and issuing a G92 E0 as one of the first start g-codes after heating and homing.

### Related Issues

N/A

### DIFF
--- a/TFT/src/User/Hal/stm32f10x/Serial.h
+++ b/TFT/src/User/Hal/stm32f10x/Serial.h
@@ -10,10 +10,10 @@ typedef struct
   uint16_t rIndex;
 }DMA_CIRCULAR_BUFFER;
 
-#if (defined(TFT28_V3_0) || defined(TFT24_V1_1) || defined (MKS_32_V1_4_NOBL))
-  #define DMA_TRANS_LEN  3072 // for lower RAM
+#if ((defined(MKS_28_V1_0)) || (defined(MKS_32_V1_4) && !defined(MKS_32_V1_4_NOBL)))
+  #define DMA_TRANS_LEN  4096 // for 64k RAM
 #else
-  #define DMA_TRANS_LEN  4096
+  #define DMA_TRANS_LEN  3072 // for 48k RAM
 #endif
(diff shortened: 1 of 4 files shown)


bigtreetech:
@looxonline Hello, Can you confirm whether https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1311/commits/c94544dedb664ecf12706d6c700f979b5597d416 is valid

looxonline:
> @looxonline Hello, Can you confirm whether [c94544d](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/c94544dedb664ecf12706d6c700f979b5597d416) is valid

Sure. I'll check it a little later this evening and give you some feedback.

looxonline:
I didn't have a chance to test tonight yet but I did have a chance to read through all of the code and it looks as if it should work perfectly. The solution is very similar to the one I proposed except it is even better because it is more general and will catch all G92 E events regardless of whether they happen at the start of the print or during the print. I will install the firmware tomorrow morning (South Africa time - midday Shenzhen time) and let you know the results.

bigtreetech:
OK, I will merge this PR after you confirm that have solved your problem

looxonline:
Testing failed. The sensor triggers each time I start a print when it is printing the skirt. If I resume the print it is able to complete the model.

Issue #1253, opened 2020-11-15
Styleguide


traffic-light:
Hello everyone,

Because BTT doesn't have a styleguide I'll make some sort of styleguide here.
It's by far not ready. But it's more then nothing.
More styleguide parts are alway's welcome.
Place them in the comments and I'll bring them into this post.

I've added this styleguide into my current PR ;-)

## turn off auto indent etcetera in vs code
To prevent vs code adjusting to much code you need to turn it off.
To do this....
### Comments in code
Comments in the code should always be in English. After the comment marks '//' their should be a space.
Example:
```
//Wrong comment
// Correct comment
```
## text styles
Using uppercase for functions and lower dash for variables.
Example:
```
Void MyFunctionName (void);
Uint8_t my_variable_name;
```
### If else statements
A if else statement should always look like this:
```
if (statementA == checkup)
{
  //Do something for B
}
else if (statementB == checkup)
{
  //Do something for B
}
else
{
  //Default
}
```
_Notice the space in between the if and '('_

When using a if else statement you should always use the curly braces. Even if it's one line!
Some examples:
```
if (whattodo == true) return; // This is correct
if (whattodo == true)
  return;                                // This should be with curly braces
if (whattodo == true)
{
  return;                                //This is correct. 
}
```

Some more examples
```
         Wrong                     |          Correct
if (checka == true)            | if (checka == true)
{ if (checkb == false)         | {
    doActionA();                 |    if (checkb == false)
}                                       |    {
                                        |       doActionA();
                                        |    }
                                        | }
```

## functions
When creating functions without parameters there should be 'void' in the declaration/ function
Example: `void MyFunctionWithoutParameter (void);`

In between the function name and the first brace should be a space as you can see in the example above

oldman4U:
👍🏻

If possible I would like to see the known bugs pinned but what about adding this information to the FR and Bug ticket!?

traffic-light:
Yeah. I understand that. In my opinion not the best location for a styleguide. But I don't know where it should be.


Maybe a styleguide.md in the files instead..

traffic-light:
@oldman4U as we speak, it's already in my PR ;-)

oldman4U:
👍🏻

Peter van Weeterloo <notifications@github.com> schrieb am So. 15. Nov. 2020
um 21:07:

digant73:
[...]

PR #1321, opened 2020-12-01
cleanup & remove duplicate code, reduce global variables & reduce code size.

guruathwal:
- Move const values out of functions to reduce program size
- Cleanup & remove duplicate codes
- Reduce global variables.
- Replace UNIFIED_MENU option with STATUS_SCREEN option.
- Remove obsolete/ unused code.
- Fix const declarations.


digant73:
I see you made some variables global instead of local but const initializers are always maintained on psp. It seams to me not an optimization. Am I wrong?
I know anyone has a preferred programming style, but compressing the code removing blank lines goes not in the direction of the recent style guide. E.g. All the new lines I normally put in my code are to clearly separate varaible definition, initialization, external function invocation, gcode functions invocations, GUI function invocations etc... Unless of wrong indentation, tabs etc... if you could avoid to change the syntax followed by the author is preferrable, at least for the code written by me. Not an offense of course, your contribution is always welcome, but consider that, for me at least, is quite boring that someone changes someting that is properly wriitten and (personally) more readable. As you see I normally limit my changes, on code not written by me, to wrong indentaion and presence of tabs avoiding to drastically adapt the code to my style guide.
Also I didn't yet understand what is the style guide you use. E.g. for "switch" statement. sometimes the "case" statement is at same indentation of {, sometimes it is indented (as I suppose it should be) and it makes to me more hard to integrate something in your code trying to respect the style followed in the code.

oldman4U:
Maybe it would be best to have a kind of guideline fro this repository. @traffic-light already started with it and I would recommend and ask guruathwal to make such a document, because he is the one who contributed most of the code.

@guruathwal I know you are very busy at the moment, but would you please be so kind and define such a document? This could be very useful and maybe would also help to reduce the amount of issues in the future.

Thank you

traffic-light:
@oldman4U yes. I started on it.

For everyone (@digant73 @guruathwal ) who's having ideas:
Feel free to comment on my issue https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/1253

guruathwal:
@digant73  If you declared a const inside of a function, every time that function is called, it would allocate a new place in RAM for that const and eat up CPU cycles during the process. This defeats the point of a const. 
Code optimization itself means changing code to reducing RAM and memory usage.
It is good to add blank lines to distinguish between code but not when it results in blank line after every line of code. Instead it will be better to change line spacing in editor itself.

@oldman4U I have started as per your suggestion. Should be completed in about 2 - 3 days.

digant73:
@guruathwal 
Well, in the best case, the compiler normally optimize the "const" avoiding to allocate memory and accessing directly the const data from the psp area. In that scenario, for both global and local variables, no memory allocation shold be performed but moving a const from local to global simply changes the visibility scope (it is increased when not necessary).
In the worst case, the compiler will also allocate memory for both global and local variables (psp is still there for both) but the global will permantly remain while the local will be released (it's on the stack) when exiting from the function. In this scenario, I see an increase of mem usage by the presence of global variable and (as for the first scenario) a change on the visibility scope.

For the code, I already explained my style. I don't put any random empty line here and there. For that aspect, anyone has his idea  and I normally don't change it on my review.

In the Mesh Editor menu, you moved some data structues on the .h file. I normally always avoid to provide structures in the .h when not necessary as in that case (all of that structures does not need to be normally tuned by user). The info that can be tuned were reported in the Configuration.h file.

traffic-light:
> @digant73 If you declared a const inside of a function, every time that function is called, it would allocate a new place in RAM for that const and eat up CPU cycles during the process. This defeats the point of a const.
> Code optimization itself means changing code to reducing RAM and memory usage.
> It is good to add blank lines to distinguish between code but not when it results in blank line after every line of code. Instead it will be better to change line spacing in editor itself.
> 
> @oldman4U I have started as per your suggestion. Should be completed in about 2 - 3 days.

So, you're making a style guide too?

guruathwal:
@traffic-light  I am writing coding standard which would cover all aspects of this repository.

Issue #1218, opened 2020-10-30
[BUG] Verified Bugs


oldman4U:
Dear reader,

This list of verified bugs has three "functions":

1: It allows users to check, if a certain bug has already been requested, before starting a new ticket.

2: To mark bugs which are already under development to avoid, that several developers are working on the same function.

3: There is now a BOT which closes inactive tickets. Some of them are probably useful, so it makes sense to make them available at a more prominent place.

Notes:
Make sure you are using the latest firmware available for your mainboard and TFT, before adding a new bug report.
Read the [Read me before posting!](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/1047) to see if you found an already know issue.
To join an existing bug report, please click on the link below. Even the ticket is closed, it does not mean that it can not be used anymore.
Feel free to add your comments related to this BUG ticket below.
New bug reports will be added as fast as possible, once more than one user verified the same issue.
Bug reports shown in Bold are under development and if possible the developer is mentioned next to it.
Notes from the community are added to each BUG in CAPITAL LETTERS.
Bugs which have been merged into the master branch, will be removed from this list.

General Bugs:
#417 TFT35 stops responding with red 'Busy processing please wait....'
#540 and #593 TFT not reading filament detector when connected to SKR
#571 Firmware assumes heated bed - SOUNDS MORE LIKE A FEATURE REQUEST
#893 (MARLIN 2.5.4 - M997 freeze SKR1.4 Turbo with BTT TFT35-E3 V3.0 only in touch mode)
#1199 Terminal outputs "Busy paused for user" after every print
#1225 Installing pre-configured firmware give errors/warnings
**#1228 Flow randomly changed from 100% to 10% during a print** Solution available in ticket

Touchscreen Related Bug:
#629 Filament motion sensor TFT display is not interested in... 
#642 Communication timeout - reset send buffer block
#1139 in print fan % change issue
#1193 TFT 35 E3 v3.0 + skr 1.4 Power loss recovery Problem - NO support for UPS module
#1245 8.3 file names on TFT35 printing screen when print from board SD

Marlin Screen Related Bugs:

Bug, Feature or Known Issue?:

PS: A user asked me what happens with this list in case I do not want to maintain it anymore or when I die. Well, I am younger than my nickname makes you believe, but this is a valid question, so here is my answer.

[...]

Issue #1170, opened 2020-10-11
[FR] Feature Request List


oldman4U:
Dear reader,

This Feature Requests list has three "functions":

1: It allows users to check, if a certain feature has already been requested, before starting a new ticket.

2: Mark features which are already under development to avoid, that several developers are working on the same function.

3: There is now a BOT which closes inactive tickets. Some of them are probably useful, so it makes sense to make them available at a more prominent place.

Notes:
Feel free to start a new FR Ticket, in case your feature or function is not listed here.
To join an existing FR, please click on the link below. Even the ticket is closed, it does not mean that it can not be used anymore.
Feel free to add your comments related to this FR ticket below.
New feature requests will be added as fast as possible.
Feature Requests shown in **Bold** are under development and if possible the developer is mentioned next to it.
Notes from the community are added to each FR in CAPITAL LETTERS.
Features which have been merged into the master branch, will be removed from this list.

General Features:
#290 Add a character generator in order to use klipper
#524 Vertical Screen GUI
**#634 The TFT display does not support the progress LED** PR #1209 available
#804 Transfer a file via Wifi ESP3D to the TFT-SD and/or TFT USB-Stick
**#1072 Smart RGB encoder control** PR #1209 available

Touchscreen Related Feature:
#296 Idle and back to main menu
#601 Invert filelist
**#732 Simplified Status Screen while Printing**
#763 Recover possibility to have the custom menu with icons
#810 Octoprint Print Status Screen
**#834 Filament change buttons** PR #1272 available 
**#865 Print finished dialog and jump back to Info Screen on TFT35**
**#1026 Alternate elapsed/remaining time during printing**
#1053 Option to deactivate the TFT LCD menu item
#1102 standard deviation on repeatability test
#1121 Runout Sensor On Board
#1137 Add ability to get to home screen with 1 click from any screen please
#1150 Include G34 Auto-Alignment to Movement/BedLevel UI
**#1166 Pre heat beep/tune/tone to acknowledge the set temperatures have been reached** PR #1209 available
**#1178 usb printing not able to control Neopixel PRINTER_EVENT_LEDS on tft35 v3 and poss other screens with usb drive ports** PR #1209 available
#1182 Show CHAMBER temperature alternately with Bed temperature
#1193 TFT 35 E3 v3.0 + skr 1.4 Power loss recovery Problem
#1211 Move X, Y, and Z Axes While Paused
[...]

PR #1336, opened 2020-12-03
New pull to replace the old smart filament sensor bug fix. 

looxonline:
### Description

This replaces the previous PR to fix the SFS. It was just easier for me to create a new PR off the current codebase. 

All of the @bigtreetech changes have been integrated but in addition the datatype bugs have also been fixed. There were two bugs:

1.) The filament position was being passed as a unit32 when it should have been a float. If a retraction occurs directly after a G92 E0 then the filament position can go negative and this resulted in the uint32 suddenly jumping to a massive number and then jumping back to a small number. This triggered the SFS algorithm twice which resulted in one of the triggers creating a false pause.

2.) The input to the FIL_SFS_SetAlive function was being declared as uint8 but passed as a bool. It is not passed and declared as the same datatype.

### Benefits

Tested and confirmed that there are no longer smart filament sensor false triggers.

### Related Issues

N/A


oldman4U:
Hi.

I am not using a filament sensor, but thank you for your contribution.

The checks have failed because there is an unsupported update. Please change the 15 env. related entries from "platform      = ststm32" to "platform      = ststm32**@9.0.0**"

In case there are open tickets related to your fix, please add "resolves #ticketnumber" to the description of this PR.

Again thank you for your help and please check #1170 and #1218 in case you have some time to spare;-)

digant73:
@looxonline In the past I got some wrong filament runout messages (so the print was paused and needed to be resumed). For me it seems that the problem was solved just increrasing the value for "fil_noise_threshold" to 200 (default value in config.ini was 100).
It seems you find and fixed a bug on filament runout feature that should allow me to reduce again "fil_noise_threshold" to 100. I will try to test it and verify if the wrong runout detection is finally fixed

looxonline:
> Hi.
> 
> I am not using a filament sensor, but thank you for your contribution.
> 
> The checks have failed because there is an unsupported update. Please change the 15 env. related entries from "platform = ststm32" to "platform = ststm32@9.0.0"
> 
> In case there are open tickets related to your fix, please add "resolves #ticketnumber" to the description of this PR.
> 
> Again thank you for your help and please check #1170 and #1218 in case you have some time to spare;-)

All of the envs are using the correct platform but I only ever tried to build for the TFT35 V3 E3 which worked. I've now built for a bunch of other envs and it seems as if anything in the 1.X range fails at the linker stage. I'm not familiar with that HW series. Was there less flash available on those CPUs?

oldman4U:
I do not know what exactly finally changed, but my PR also failed and changing the config.ini like described above solved the problem. @kisslorand knows more about it

kisslorand:
FIL_SFS_SetAlive is not recognized in interfaceCmd.c

oldman4U:
@AntoszHUN

Could you check if this solves your problem with the sensor.

AntoszHUN:
> @AntoszHUN
> 
> Could you check if this solves your problem with the sensor.

I'll see you as soon as I have time.

looxonline:
> FIL_SFS_SetAlive is not recognized in interfaceCmd.c

The thing is that it is actually recognized at a compiler level but it breaks at a linker level. Also, it only breaks at linker level for the V1.X hardware environments. All other environments build without any issue. This is why I think that there is some sort of an issue with NVMEM allocation with the earlier hardware revisions when trying to link this particular code. I noticed that the python build script is different for all of the V1 hardware versions which seems to indicate that they may have used an MCU with less flash although I have not read the script in detail in order to confirm this. Nevertheless, if this is the case then it could be possible that there is not enough flash to link the final binary on the V1 hardware and that is where the error is coming from.

Issue #1300, opened 2020-11-28
[BUG] open Extruder Menu after Clicking on Preheat Bed, nonstopping beep and layerhigh TFT35 E3 V3


Bastlwastl84:
clicking on preheat Bed opens Extruder Menue
[Configuration.h TFT35E3V3.txt](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5610674/Configuration.h.TFT35E3V3.txt)
[platformio.ini TFT35E3V3.txt](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5610676/platformio.ini.TFT35E3V3.txt)
Video:
https://drive.google.com/file/d/1AFLJ0R-qH5XDzpzdTQB7rrH-FDbx7M5g/view?usp=sharing

Also happens that sometimes the TFT does a long Beep (happens at 115200 Baud also) after clicking at a Button till i click again at any other Button

Time is incorrect  and Layer. Did a Printjob before and it showes 2mm at Layer 1 (0,16mm Layerhigh). Print starts at USB_Stick at TFT. Sometimes the Display jumpes to correct high 

![20201128_145214](https://user-images.githubusercontent.com/69717507/100517225-73528300-3189-11eb-8158-827800bebb4c.jpg)

kisslorand:
Check PR #1304. It might help you.

oldman4U:
Hello @Bastlwastl84 .

What is the status of the reported issues?

1: Extruder Menu when clicking Preheat button - should be fixed with #1304 - THX @kisslorand 
2. long beep - is this a new issue or have you seen this also with a previous version?
3. Time is incorrect? Is this a new issue or.... What exactly is wrong with it? It should show the elapsed time.
4. Layer hight. Are you sure it does not show the layer hight at a Z hop retract?

More information is needed, so please let us know.

Thank you

oldman4U:
Regarding 4. The layer  hight (when printing from TFT SD or uDisk) is only updating "by mistake" every few layers. Leaving the main screen and going back refresh the Layer hight to the correct value temporarily.

Bastlwastl84:
> Hello @Bastlwastl84 .
> 
> What is the status of the reported issues?
> 
> 1: Extruder Menu when clicking Preheat button - should be fixed with #1304 - THX @kisslorand
> 2. long beep - is this a new issue or have you seen this also with a previous version?
[...]

Issue #1294, opened 2020-11-27
TFT35 - Stop Print from Board SD Card - Stuck in "Loading..." Screen


kageryu42:
### Description

Hi, I just installed a TFT35 V3 display and an SKR E3 Turbo board for my Ender 5 Plus.  I used Cheetah to compile the software and enabled all option mentioned in the config.ini file.  So, when I start a print from the board's SD card slot and stop it, I get stuck in a "Loading..." screen.  My only option at this point is to either push the reset button or turn the machine off/on.  I've noticed the behavior is a little erratic where sometimes it will not get stuck on this screen.  When it doesn't get stuck it sends me to the screen where you chose the menu type (touch vs analog).  I've noticed there is a yellow dot in the upper right corner of the screen.  I'm not sure if this is a clue or not to what is going on.

This does not happen if I print from the display SD card slot.

### Steps to reproduce

1. Start a print from the board's SD card slot
2. Stop the print

**Expected behavior**
From the way printing behaves when printing from the display SD card slot, I should be sent back to the SD card's menu of items to be printed.

**Actual behavior**
The screen gets stuck in the "Loading..." screen or I'm taken to the screen to pick between the touch screen mod and the analog mode.  If I'm sent back to the menu to pick between the touch screen mod and the analog mode, choose touch screen, try the print again and stop it, I get stuck in the "Loading..." screen.

### Hardware Variant
TFT35 V3
### TFT Firmware Version & Main Board Firmware details

BIGTREE_TFT35_V3.0.26.x 02-Nov-2020 12:01AM - Compiled by radek8
Cheetah Marlin 2.0.7.2

### Additional Information
[config.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5608456/config.zip)
![IMG_0438](https://user-images.githubusercontent.com/75130858/100462842-d9131200-3098-11eb-913a-fd1bf03fa254.JPG)

oldman4U:
Hi.

Another user reported similar issues and has been able to fix it by:

I needed to format the SD Card (not fast formating fat32 and standard-blocksize and set to 250000 Baud, than i coulded Connect to the sd card onboard from tft.

Please try this and let us know

kageryu42:
I just tried reformatting the SD card.  I removed the tick for a Quick format, confirmed fat32 and set allocation unit size to "Default allocation size".  Unfortunately, the issue is still happening.

[...]

Issue #1293, opened 2020-11-27
TFT35 - Stop Print - Stuck in "Loading..."


kageryu42:
### Steps to reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior**

**Actual behavior**

### Additional Information
* Include a ZIP file containing your `Configuration.h` or use [Pastebin](https://pastebin.com/) and paste a link in this issue.
* Provide pictures or links to videos that clearly demonstrate the issue.

kageryu42:
details missing

PR #1344, opened 2020-12-04
Lighter ram memory allocation/possible long beep fix

kisslorand:
### Requirements
BTT TFT of MKS TFT

### Description

This is an attempt to reduce the amount of memory allocated for serial ports DMA's buffer. My guess is that only Marlin would send huge chunks of data (see M503 response), the other peripherals would send way much less/shorter data. I do not know what other people have connected to the TFT and what does what. I only assume those deives would be more shy in filling those buffers. I can only test with SD and USB connected to my TFT.

Also I did some changes that I thought would eliminate the random long beeps. While testing my theory I couldn't make my TFT to beep, Spent 2 hours pressing like crazy on the buttons, doing the most abnormal button/menu combinations while printer was printing, halting, heating, moving, etc.  Not a single long beep.

### Related Issues
#1300 #1294 #1293

### To be done

This has to be tested seriously because it's prone to introduce new, unforeseen issues., so people willing to test are really welcome.


oldman4U:
What exactly has to be tested. I can help but never saw the long beep on my printer.

kisslorand:
It should be more onboard SD friendly, that was the first thing in my mind that lead to this PR. The whole purpose is to generally have more usable RAM for present and future features.
What to test? Just simply if it doesn't bring new issues, main,y with things connected to the TFT's serial ports.
If you have a perfectly OK system it will not bring anything new, at least for now, it just leaves room for the future addons.

digant73:
@kisslorand 
I would define the size of each queue as a new attribute in DMA_CIRCULAR_BUFFER (in Serial.h). E.g. "uint16_t cacheSize;".
Also I would define the size for each queue in Configuration.h (e.g. #define SERIAL_PORT_QUEUE_SIZE 3096) so anyone having specific requirements on some queues size can tune them as they need.
Also MKS TFT have been limited to 48KB and not 65KB. The limits are defined in the .ld spec file in the project. You can easily verify that simply defining a new array of about 8KB (e.g. uint8_t buf[8000];) and you will get and heap/stack RAM error due to no minimum space left for heap (minimum 8Kb). also 8KB is the minimum for the stack. So the maximum non dynamic RAM you can use is 48KB - 16KB. When you reach this limit you get a compile error because no more minimum space can be reserved to heap or stack

Bastlwastl84:
> What exactly has to be tested. I can help but never saw the long beep on my printer.

i had tested it with the Buid from 3.12. and war reducing the beep-time to 2ms at tft, than there was no long beep since that, but than i can't hear filament-change beep

kisslorand:
Than that beep for filament runout must be rewritten, because it was in fact the effect of the bug. There is no long beep defined in this FW.

kisslorand:
> @kisslorand
> I would define the size of each queue as a new attribute in DMA_CIRCULAR_BUFFER (in Serial.h). E.g. "uint16_t cacheSize;".
> Also I would define the size for each queue in Configuration.h (e.g. #define SERIAL_PORT_QUEUE_SIZE 3096) so anyone having specific requirements on some queues size can tune them as they need.

Done.

> Also MKS TFTs have been limited to 48KB and not 65KB. The limits are defined in the .ld spec file in the project. You can easily verify that simply defining a new array of about 8KB (e.g. uint8_t buf[8000];) and you will get and heap/stack RAM error due to no minimum space left for heap (minimum 8Kb). also 8KB is the minimum for the stack. So the maximum non dynamic RAM you can use is 48KB - 16KB. When you reach this limit you get a compile error because no more minimum space can be reserved to heap or stack
> 

Done.  MKS TFT now uses its full 64K RAM.

> EDIT: I forgot to mention that the new (un)load menu requires FILAMENT_LOAD_UNLOAD_GCODES enabled in Configuration_adv.h (marlin FW). It should be reported in the config.ini file

Food for thought for another PR.

oldman4U:
Hi.

I have been testing @mehmetsutas PR for some hours, and there is still one issue left. Hopefully I have some time this coming Tuesday to test your PR. Do you believe your PR could help with the lagging issue, every time the speed is changed, for example? This is the worst issue for me, once all those PR's are merged.

kisslorand:
> . Do you believe your PR could help with the lagging issue, every time the speed is changed, for example? This is the worst issue for me, once all those PR's are merged.

I am not aware of such an issue. Could you please elaborate it?

kisslorand:
So...
I took some time to check each board. I was shocked that the majority has a very capable MCU with 128K RAM but BTT decided to use only 48K. I guess this is for keeping the compatibility with older TFTs. Only BIGTREE TFT35 V1.0, BIGTREE TFT35 V1.1, BIGTREE TFT35 V1.2, BIGTREE TFT35 V2.0, BIGTREE TFT28 V1.0 have 48K RAM. All the others have 128K RAM !!!
I think it is not anyone else's job but theirs to do something about this, so I will not invest more time in BTT boards RAM usage.

digant73:
@kisslorand they probably limited the resources just to avoid to abuse of them too quickly

oldman4U:
> @kisslorand they probably limited the resources just to avoid to abuse of them too quickly

You really believe so...;-)

kisslorand:
Probably the truth is more simple... cough-lazi-cough-ness-cough

oldman4U:
You mean something like - no time to do it right - or so....

;-)

radek8:
@kisslorand 
Do you want to simulate a long (permanent) beep?
Open a terminal on the TFT
send command M300 S440 P300
a short beep sounds.
exit the terminal with the back button.
A continuous tone sounds.

kisslorand:
@radek8 
Thank you for the hint, good catch!
I was unable to properly diagnose the problem without a repeatable method to trigger the long beep.
Having the possibility to trigger the long beep at will, I was able to find at least one problem that led to that. Hopefully it was the only problem. 
The Buzzer.c got a little overhaul, fixed some not obvious bugs also.

radek8:
I found another place where there was a continuous beep, but your fix solved it. Thank you.

kisslorand:
The beep was driving me crazy happening at the least expected times, I really hope it will be in the past once this PR is (if) merged.
Regarding the main thing of this PR, the RAM usage, not much feedback came. It's been almost a week, I think no new volunteers will try this PR, so I changed it from draft to be ready to be reviewed.
Will see (if merged) how it will improve or not the "onboard SD loading hanging" issue.
Really thanks for the help guys!

radek8:
Loading the integrated SD onboard is fine

ChuckNoxis:
It's beeping a lot less now thanks ! <3
I still had a random beep when navigating through the menus but I don't remember how/where ^^

PR #1346, opened 2020-12-05
FIX Italian language

BsCmOD:
Fix Italian language


guruathwal:
Please always try to include similar changes in single PR instead of creating separate PR for each file (#1347 )

BsCmOD:
ok but I don't know how to do it

BsCmOD:
Close to new PR to merge different files

PR #1347, opened 2020-12-05
FIX Italian language in language_it.ini file

BsCmOD:
FIX Italian language in language_it.ini file


BsCmOD:
Close to new PR to merge different files

PR #1351, opened 2020-12-06
config.ini cleanup

digant73:
A general cleanup on config.ini file:
* Title on section part ALWAYS uses capital letters for any word
* Title on paramter part ALWAYS uses capital letters for any word
* Title on paramter part NEVER provides any comment/description
* A parameter comment/description is provided ONLY on paramter body part
* Parameter qualifiers such as:
  * Format
  * Unit
  * Value range
  * Options

  are ALWAYS provided on parameter body part when neeeded (e.g. °C, mm, seconds etc...)

**PR STATE:**

\
\
\
SECTION PART:

\#--------------------------------------------------------------------
\# <TITLE> (e.g. Supported TFT Variants)
\#--------------------------------------------------------------------
\# <comment/description>
\#___________________________________________________________________<- this empty comment line is always present
\
\
\
PARAMETER PART:

\#### <TITLE> (e.g. Unified Menu / Classic Menu) 
\# <comment/description>
\#
\# Format: [...]
\#
\# Unit: [...]
\#
\# Value range: [...]
\#
\# Options: [....]
\#___________________________________________________________________<- this empty comment line is always present

### DIFF
--- a/Copy to SD Card root directory to update/config.ini
+++ b/Copy to SD Card root directory to update/config.ini
@@ -1,282 +1,413 @@
-#### Default Config file for BigTreeTech TFT Controllers
+#--------------------------------------------------------------------
+#
+# DEFAULT CONFIG FILE FOR BigTreeTech TFT CONTROLLERS
+#
+#--------------------------------------------------------------------
+
+
+#--------------------------------------------------------------------
+# Supported TFT Variants
+#--------------------------------------------------------------------
 #
 # BIGTREE_TFT35_V1_1 / BIGTREE_TFT35_V1_2    / BIGTREE_TFT35_V2_0
 # BIGTREE_TFT35_V3_0 / BIGTREE_TFT35_E3_V3_0 / BIGTREE_TFT28_V1_0
 # BIGTREE_TFT28_V3_0 / BIGTREE_TFT24_V1_1    / MKS_32_V1_4
 # MKS_28_V1_0
 #
-# Firmware Source: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware
+# Firmware source: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware
+#
 
 
 #--------------------------------------------------------------------
-# Supported Marlin Firmware Version
+# Supported Marlin Firmware Versions
 #--------------------------------------------------------------------
-#  Minimum Marlin Firmware Version: 2.0.5.4
-#  Distribution date: 2020-05-12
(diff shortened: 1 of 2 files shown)


oldman4U:
Hi.

It seems that 90% o the configuration.h is covered by the config.ini but not all of the functions. Wouldn't it be good to change configuration.h to make it better recognisable which features are not covered?

digant73:
Well, I currently reviewed the config.ini because it is the only configuration file that acts as a configuration/user manual for any kind of end user. So it is important that at least it is compliant and coherent to common rules and format all around the file.
Configuration.h has also different settings and its target is developers that know the meaning (or are able to find and understand) of the settings

Copy to SD Card root directory to update/config.ini
```diff
@@ -1,282 +1,413 @@
...
-# Pause feed Rate # format      [ X<feedrate mm/min> Y<feedrate mm/min> Z<feedrate mm/min> E<feedrate mm/min>]
+# Format: [pause_retract: R<retract length> P<Resume Purge length>]
+#         [pause_pos: X<position in mm> Y<position in mm>]
+#         [pause_z_raise: Z<distance in mm>]
+#         [pause_feedrate: X<feedrate mm/min> Y<feedrate mm/min> Z<feedrate mm/min> E<feedrate mm/min>]
+#
+# Unit: [position & length: mm, feedrate: mm/min]
+#
+# Value range: [min: 0, max: 2000]
+#
```
guruathwal, line 385: Do not remove the description/name of the config lines.

digant73: not my intention. Moved param names at the beginning of the section for Pause and Leveling settings

raufemann (approved):

oldman4U:
Who is raufemann!!??

Sounds like an angry german man;-(

Copy to SD Card root directory to update/config.ini
guruathwal: instead of a separate list for names and a separate list of formats, it would be better if the param names are clubbed together with their respective format.

digant73: well, those must be  param description not param names. Param names are already reported in the format. Also reporting them in the same line will make it less readable. You now always easily identify tags such as Format, Unit etc, while they were previously  mixed with param name/description etc... For the 2 cases you refer to, it should be better to provide a description instead of a fully qualified param name (for that reason I previosly removed them. They didn't add any description)

PR #1360, opened 2020-12-08
some cleanup after last PRs

digant73:
A general cleanup after last PRs

@bigtreetech:
* By a mistake on a previous PR, the file BIGTREE_TFT35_V3.0_E3.26.x.bin was also provided in the main project directory. It was now removed
* All bootloaders directories (e.g. "TFT35_V2 Bootloader fix") where moved to a common directory called "Bootloaders" under the main project directory (I adopted the same approach used for "Copy to SD Card root directory to update")

**NOTE:**
* It should be preferrable to merge this PR before others just to eventually let them allow to be aligned (current active PRs provide less code, so more easy to be eventually adapted to this one)

**PR STATE:**


TFT/src/User/API/FanControl.c
```diff
@@ -28,40 +31,49 @@ void fanBuildList(void)
...
     );
 }
 
-bool fanIsType(uint8_t i, uint8_t type) {
+bool fanIsType(uint8_t i, uint8_t type)
+{
   return (fanType[i] == type);
 }
 
 void fanSetRcvSpeed(uint8_t i, uint8_t speed)  // by Lori
```
kisslorand, line 72: Please delete "// by  Lori" comments, I left it there by accident, it's meant only for my personal mods

digant73: ok I will delete it from all files

looxonline:
That was my bad. I committed my branch with the bin in the main dir and issued the PR with it still there. BTT just needs to remove though.

digant73:
> That was my bad. I committed my branch with the bin in the main dir and issued the PR with it still there. BTT just needs to remove though.

ok thanks. I removed it and moved all bootloaders under a new Bootloaders directory

traffic-light:
Thanks digant73!

vitaliy172 (approved):

TFT/src/User/API/Temperature.c
```diff
@@ -122,7 +126,7 @@ void heatSetUpdateSeconds(uint8_t seconds)
   if (infoMachineSettings.autoReportTemp && !heat_update_waiting)
   {
     heat_update_waiting = true;
-    storeCmd("M155 ");
+    storeCmd("M155\n");
   }
```
bigtreetech, line 130: Hello, There should not be '\n', because in the `sendQueueCmd()` function, the latest temperature update time will be obtained and sent in the form of "M155 Sxx\n"
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/ce2ea18b94abe8e94b80e69e405999ded1b8c73a/TFT/src/User/API/interfaceCmd.c#L524

Adding '\n' to this gcode means that `"M155\n"` will be sent directly to Marlin, and the default `'S'` value of Marlin is 0. So automatic temperature reporting will never be available in this way,

digant73: my bad, I will rollback the change

bigtreetech: I'm doing it

PR #1361, opened 2020-12-08
Improve Ukrainian language

vitaliy172:


Antti (approved):
LGTM

Antti:
@vitaliy172 I think you also need to run `buildroot/scripts/auto_gen_language_pack.py` to generate the language packs.

vitaliy172:
Yeah but autogenerated translations isn`t that good. I`m native Ukrainian so I can translate a bit better

Copy to SD Card root directory to update/Language Packs/language_uk.ini
```diff
@@ -202,8 +202,8 @@ label_move_speed:Швидкість (X Y Z)
 label_knob_led:Підсвічування енкодера
 label_knob_led_idle:Простій підсвічування енкодера
 label_m0_pause:Пауза по команді M0
-label_send_start_gcode:G-код перед друком
-label_send_end_gcode:G-код після друку
+label_send_start_gcode:Розпочати G-код перед друком
```
Antti, line 205: I don't think these are correct. You don't "begin" or "end" a G-code.
It toggles sending start/end g-code On or Off.

vitaliy172: English version:
`label_send_start_gcode:Start Gcode before print
label_send_end_gcode:End Gcode after print`

Antti: I think "Start Gcode" means the gcode, which is exected on the start.
Same with the "End Gcode" - gcode, which is executed on the end.

Antti:
`label_send_start_gcode` and `label_send_end_gcode` are incorrect 😞

PR #1376, opened 2020-12-12
Higher calibration resolution in the Extruder Calibration Menu

NicoletoAerials:
### Requirements

### Description

When calibrating the extruder steps, we are not actually extruding filament but changing the stepper steps. In such case, it is better to use icons relating to steps movement than extrusion. In addition, it is better to be able to calibrate steps down to 0.01 resolution.

### Benefits

Higher calibration resolution. From 0.01 to 10 mm


traffic-light:
You're pretty good in measuring when you can measure 0.01 precisely on extruded fillement 😜

oldman4U:
Maybe he is using inches.

🤔

NicoletoAerials:
> You're pretty good in measuring when you can measure 0.01 precisely on extruded fillement 😜

Jajajajaja, Yes, it may be an overkill. However, when calibrating steps I've noticed that it is better to al least be able to measure from 0.1 to 10 millimeters so one can input steps with 000.00 resolution.

In any case, the idea here is that in reality when calibrating E steps, we are calibrating length and stepper motors steps and no really extruding, that's why I think is better to use movement than extrusion icons.

PR #1379, opened 2020-12-12
resolves #1377 + cleanup

### LINKED ISSUE #1353: Question about "fil_runout:" in the config.ini file
I have a TFT35 E3 V3.0 updated with the latest firmware which includes the filament sensor (#1336)bug fix. I also have the BTT SFS connected to the TFT. I changed the "fil_runout:0" in the config.ini file to "fil_runout:2" for the smart sensor. After the firmware update is done installing, I go into the settings menu and it only shows the filament sensor as being "on" and not set as "smart". I have to manually select "smart". Is this the correct way it is supposed to work? Would this have any effect on how the sensor responds?

I am still getting false positives when printing. I think the main reason may be I have a Micro Swiss direct drive/hotend and the SFS is mounted where the extruder use to be on my Ender 3 Pro. I feed the filament from the side due to height limitations where the printer is located. I have the SFS connected to the direct drive through a bowden tube as a guide. I have tried different values in the "fil_noise_threshold:" and "fil_runout_distance:" with limited success. If anyone has any suggested settings for a direct drive I would really appreciate it.

### LINKED ISSUE #1363: Z-offset menu bug, cause X,Y axis Impact structure
### Description

When activating Z Offset via TFT firmware with existing negative BLTouch offsets, the TFT sends GCODE commands to the printer which seem to drive it beyond the stop movement limits.   

### Steps to reproduce

1. Via GCode, configure X, Y and Z offsets for BLTouch location to all be negative
1. Via touch screen UI, go to Menu > Movement > Bed Leveling > Z Offset
6. By default, Z Offset will be in "OFF" state, and correct negative Z Offset displayed at top of screen
7. Press Z Offset OFF button to turn it on

**Expected behavior**

- Printer homes X, Y, and Z to zero based on limit switches (controller board firmware to use BLTouch for zeroing Z is NOT being used)
- Z Offset modification is now enabled
- Z offset value can be changed via the UI and fine tuned to optimize bed clearance as needed

**Actual behavior**

- Printer homes X, Y, and Z axes, stopping at limit switches
- Printer then proceeds to try and move Z axis below level of bed and past stops
- Neither X or Y homing put the extruder tip on the bed, so there is no impact with the bed, however extruder tip clearly moves to a position several MM below the bed
- Horrible noises ensue due to it trying to exceed mechanical limits
- Printer powered down at this point to avoid mechanical issues
- As a result, one is never able to actually modify the Z offset value via the UI, and must use GCode messages to change values.

Note that printer and BLTouch work just fine and do exactly what they should do when printing.  Printer is able to automatically probe print bed, compensate accurately, and otherwise print without issue.

### Hardware Variant

- Ender 3 Pro
- SKR mini E3 V2.0
- TFT 35 V2
- BLTouch 3.1 with mount-specific X, Y and Z offsets set via direct GCode and all values negative (-42, -10, -1.18)

SKR mini E3 V2 firmware with BLTouch NOT used for z homing (https://github.com/bigtreetech/BIGTREETECH-SKR-mini-E3/blob/master/firmware/V2.0/firmware-bltouch.bin)

TFT 35 V2 firmware (https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/Copy%20to%20SD%20Card%20root%20directory%20to%20update/BIGTREE_TFT35_V3.0_E3.26.x.bin)

### Additional Information

Reviewing the TFT firmware source code, when you select the Z Offset ON option via the TFT, it calls probeOffsetEnable() in src/User/API/ProbeOffsetControl.c 

When you do this, here is the gcode that is sent to the printer.  I've added what function the gcode is generated in via # type comments.

    ; In probeHeightEnable()
    M211 S0; disable software end stops

    ; In probeOffsetEnable()
    G28; home, which takes all axes to the point where they hit their stops

    ; In probeHeightStop()
    G91; set relative position mode
    G1 Z<z_probe offset;  move nozzle to relative z offset coordinate
    G90;  return to absolute position mode

    ; In probeOffsetEnable()
    G91; set relative position mode
    G1 X<probe offset> Y<probe offset>; move nozzle to x and y probe coordinates

    ; In probeHeightStart()
    G90; absolute position mode
    G1 Z0;  move nozzle to absolute Z0 coordinate
    G91;  back to relative position mode

I believe what is happening is this...
- It's disabling software end stops
- It is taking X, Y, and Z all the way to zero based on the mechanical stops (micro-switches) via G28 command
- It then tries to move relatively by the X, Y and Z offsets
- Since the offsets are negative, it is trying to move PAST the X/Y/Z mechanical stop limits, which causes the noise

Here is a Dropbox link to a video showing the issue -> https://www.dropbox.com/s/6mv4gh2ja8723hk/Video%20Dec%2008%2C%208%2049%2021%20PM.mov?dl=0

I am not 100% certain which axis is being exceeded to cause the noises.  It may actually be the X axis limit, since it is larger.

### LINKED ISSUE #1364: Zoffset menu issue : the nozzle dives into the bed after modifications on hotend
I have TFT35 3.0, I like the Zoffset menu which enables to live test it, but I think it would be great to add some safety option. What I mean is that if the Zoffset is at the moment of -3mm and if I make some changes on the carriage and need something like a -2.8 Zoffset the nozzle will dive into the bed by 0.2mm (-2.8 -(-3)) before I can add the 0.2mm to the Zoffset on this menu. So it's a serious problem here. Actually what I do now is that I send some Gcode to change the Zoffset because of that issue, and then tune it with the actual menu of the TFT.

I propose to add an option in the Zoffset tune menu to change directly before launching the {homing+G1 Z0}, and then tune it like it's actually made.

### LINKED ISSUE #1377: [BUG] (MBL - No Mesh Saved - No Mesh In Editor)
When using Manual Mesh Bed, before you have to choose what slot to save to, and you could even edit the mesh from the TFT, with newest version - Compiled Today 12/12-2020 you don't have the possibility to choose what slot to save to, and the Mesh Editor just show a screen with no dots and then freeze, so when start printing no mesh is used.

digant73:
**BUG FIXES:**
* resolves #1377 (Mesh Editor freeze after one of the last merged PR)
* resolves #1364: Z Offset will start from probing Z0 point instead of absolute Z0 point. It means original Z offset is not applied when Z offset is enabled on Z offset menu
* resolves #1363 (probably solved now)
* resolves #1353 (value "SMART" for "fil_runout" is now correctly read from config.ini)
* MBL safe starting Z point: initial Z points are now set to 1mm instead of Z0. This should make more safe the MBL process in case of warped bed or not good manual leveling
* some cleanup

**PR STATE:**


ETE-Design:
@digant73 See my last comment on #1377 Seems like it dosen't fix it :-(

radek8:
The repair works.
Problem #1377 solved.

ETE-Design:
@radek8 Still have the problem even with this fix :-(

digant73:
The previous code was restored and is working properly now. You probably missed something. Retry please

szepnorbee (approved):

Octet-nl:
This does not fix #1363.

I had to remove this line (around line 20) in ProgeOffsetControl.c:
`  mustStoreCmd("G1 X%.2f Y%.2f F%d\n",

    getParameter(P_PROBE_OFFSET, X_STEPPER),

    getParameter(P_PROBE_OFFSET, Y_STEPPER),

    infoSettings.axis_speed[infoSettings.move_speed]);     // move nozzle to XY probing coordinates and set feedrate
`
The nozzle has just homed (line 15) and I don't see the sense in moving it beyond this point. The P_PROBE_OFFSET values are ( in my case) beyond the physical capability of the nozzle, these offsets are used to calculate the distance from the probe to the nozzle. Trying to move the nozzle to the probe position is not a good idea and it will cause these horrible grinding noises.

digant73:
@Octet-nl 
After homing, the nozzle is moved on the probing point just to start using the usual paper to tune z offset.
In #1363 it was primarly reported the problem was on Z axis. If XY offset cannot be applied because they will be beyond the limits, it has no sense to set homing position on the edge of bed but it's enough to enable Z_SAFE_HOMING on Marlin FW (that is strongly recommended by Marlin FW when using a probe)

Octet-nl:
As you mentioned correctly in #1363 the problem was not on the Z axis but on X and Y. At first it looked like a Z axis problem because it was caused by enabling Z offset.

When I had my BL-touch connected to the Z-endstop connector, it made sense to enable Z_SAFE_HOMING. Now I have the original Z-endstop connected, and the BL-touch is connected to the PROBE connector on the BTT SKR mini E3. This is working great and I don't want to change this. As a wise Marlin person once said: A probe is not an endstop and an endstop is not a probe.

(Reason for this is that the BLtouch once failed as an endstop so the nozzle drilled a 1 sq mm hole in the middle of the build plate. So for me no more homing there, it is not safe :p )

Apart from that, I don't see a reason for the movement of the nozzle to an area that can't be reached. I removed the offending statement and now everything works as expected. Is there any reason why I should not do this?

Octet-nl:
As an experiment I removed the complete contents of probeOffsetEnable and probeOffsetDisable, except for the statements `probe_offset_enable=false/true`. This seems to work fine, probe offset can be toggled between on and off and there are no strange side-effects, the probeOffset can be set just like in the Marlin screen. Is there any reason why I should not do this either?

digant73:
Understood your setup and the problem you had on the bed. The better solution is to make XY offset configurable based on preference/need. However, an HW setup forcing to use a  probing point that is never reachable and far to be a good probing point for z offset is quite useless for me and I would tried a different setup

Octet-nl:
XY offset is now measured with respect to the probe position as per Marlin config instructions.
Will keep this setup for the moment to see if anything strange happens. Thanks for thinking along.

digant73:
> XY offset is now measured with respect to the probe position as per Marlin config instructions.
> Will keep this setup for the moment to see if anything strange happens. Thanks for thinking along.

try PR #1415. It should solve your problem (set the new parameter on config.ini to 0)

PR #1384, opened 2020-12-13
Add cold extrusion check, bug fixes, Cleanup and remove duplicate codes

### LINKED ISSUE #1353: Question about "fil_runout:" in the config.ini file

guruathwal:
- Cleanup and remove duplicate codes.
- Fix `fil_runout:` read incorrectly (also mentioned in #1379).
- Change font colour logic for live icons.
- Add cold extrusion check.
- Fix load-unload popup (#1378).
- Remove duplicate global variables.
- Fix wrong settings after (#1333).

Fixes #1353, #1378


oldman4U:
Hi Gurmeet.

So after this version, a user should get a warning again in case the firmware and the config.ini (or TFT) does not match, right?

Thank you

kisslorand:
@guruathwal 
You enabled Cold Extrusion Prevention in the Load/Unload menu. Did you know M700 and M701 takes care of that automatically?
If nozzle not heated, it preheats it to 180 grade Celsius before proceeding further.
Just want to know why did you enabled it, I am about to tweak a little more that menu, so I need the logic behind that activation of yours. It was so convenient for me to not press many buttons for Load/Unload.

guruathwal:
@kissorland Yes I know, I added it because it is better to make the user do it manually because M701/M702 makes marlin unresponsive during heating which is making the interface unusable.

@oldman4U It will not ask for a config file. it will reset the settings. it asks for a config file only when the signatures of settings stores on SPI flash are changes. currently, the new parameter added is saved on program memory which resets if the signature mismatch.

guruathwal:
@kissorland Yes I know, I added the check because it is better to make the user do it manually because M701/M702 makes marlin unresponsive and making the interface unusable makes the user think something is wrong with it.

@oldman4U It will reset the settings. it asks for a config file only when the signatures of settings stores on SPI flash are changes. currently, the new parameter added is saved on program memory which resets if the signature mismatch.

kisslorand:
When using directly the heat function of M700 and M701 the screen let the user know that there's an ongoing process and wouldn't let the user to navigate away until M700/M701 is finished.  So the screen is not frozen, the user knows what's going on. If the user would want in the future to click around whilst the nozzle is heating before the Load/Unload, he can use the "heat" in advance and return/go to the Load/Unload menu once the desired temperature is reached.
There are other situations when Marlin gets unresponsive until the given task is finished (Homing, ABL, etc) so it's not something new to any user. In every case there are obvious signs, messages that something is going on, the printer is not frozen.
The way it is now, it takes away an option, the one that makes Load/Unload a more automated procedure, no need to click multiple times for the desired nozzle temperature. As far as I know any material can be loaded or unloaded at 180 degree Celsius.
I am inclined to let the users make the choices, but, making sure if possible, not stupid ones.
This is just my personal view, not a complaint.

guruathwal:
The heat button is provided in load/unload menu for easy access already. what's the point in blocking the user from clicking on all buttons. In other situations, the user is still able to navigate through the menu even if marlin is busy and not accepting commands. but here it is taking control away from him by not allowing to click on any option, not even the back button to allow going to the status screen while it loads/waits for heating. isn't this stupid?

kisslorand:
Good points.  I might have an idea to have best of both situations.
Thanks for the explanations.

PR #1394, opened 2020-12-15
minor cleanup

### LINKED ISSUE #1386: [BUG] Onboard SD print - Layer display stops at x height
### Description

Printing a small object from onboard SD and the layer height stops at 0.80 mm. If I do nothing it stays that way, if I click "More" and than "Back" the layer height is updated until the end of the print.

### Steps to reproduce

1. [Start print]
2. [Watch layer height]
3. [and so on...]

**Expected behavior**

 Layer to update until the end of the print job.

**Actual behavior**

Layer display stops at 0.8mm
### Hardware Variant

Clone of the MKS TFT 28

Latest build as of 14.XII.2020
### Additional Information

Here's the gCode:

[Strap6mm.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5685793/Strap6mm.zip)

digant73:
just a minor cleanup (not committed on my just merged PR)

fixes #1386

### DIFF
--- a/TFT/src/User/Menu/TuneExtruder.c
+++ b/TFT/src/User/Menu/TuneExtruder.c
@@ -145,27 +145,31 @@ void menuTuneExtruder(void)
 
       case KEY_ICON_6:
         {
-        char tempMsg[120];
+          char tempMsg[120];
 
-        if (heatGetTargetTemp(c_heater) < infoSettings.min_ext_temp)
-        {
-          labelChar(tempStr, LABEL_TUNE_EXT_TEMPLOW);
-          sprintf(tempMsg, tempStr, infoSettings.min_ext_temp);
-          popupReminder(DIALOG_TYPE_ALERT, tuneExtruderItems.title.index, (u8 *) tempMsg);
-        }
-        else if (heatGetCurrentTemp(c_heater) < heatGetTargetTemp(c_heater) - 1)
-        {
-          popupReminder(DIALOG_TYPE_ALERT, tuneExtruderItems.title.index, LABEL_TUNE_EXT_DESIREDVAL);
-        }
-        else
-        {
-          labelChar(tempStr, LABEL_TUNE_EXT_MARK120MM);
-          sprintf(tempMsg, tempStr, textSelect(LABEL_EXTRUDE));
-          setDialogText(tuneExtruderItems.title.index, (u8 *) tempMsg, LABEL_EXTRUDE, LABEL_CANCEL);
-          showDialog(DIALOG_TYPE_QUESTION, extrudeFilament, NULL, NULL);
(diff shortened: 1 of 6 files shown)


TFT/src/User/Menu/TuneExtruder.c
```diff
@@ -145,27 +145,31 @@ void menuTuneExtruder(void)
...
+          }
+          else
+          {
+            labelChar(tempStr, LABEL_TUNE_EXT_MARK120MM);
+
+            sprintf(tempMsg, tempStr, textSelect(LABEL_EXTRUDE));
+
+            setDialogText(tuneExtruderItems.title.index, (u8 *) tempMsg, LABEL_EXTRUDE, LABEL_CANCEL);
+            showDialog(DIALOG_TYPE_QUESTION, extrudeFilament, NULL, NULL);
+          }
```
guruathwal, line 170: why are you adding space after every line?

digant73: I corrected the indentation of that piece of code starting from the original code. My pr was still on draft when merged. However, if any style guide will be defined/applied it must be applied by anyone working on the project and on the whole project otherwise it is simply a waste of time for me, you and everyone. It should be mandatory to enable and use a defined beautifier on vscode. It should be better to have a defined style guide from now on. But in the meanwhile  I think it is better to limit changes only to clear wrong format. E.G. I usually always separate variables definition from code by a blank line etc._.. That is not a format violation but simply a personal preference I adopted and imposed in all of my teams just because it allows to make more uniform and more clear the code written by people otherwise using totally different approaches

guruathwal: I request you to not take it on you personally. Adding space to differentiate between assigning values and function calls is a good thing to do, but do you think it is good if adding line spaces results in space after every line or if there are just 2-3 lines per block? This results in less code lines being shown on the screen which makes it hard to analyse and edit the code. Instead wouldn't it be better to increase overall line spacing in the editor iteself. 
Should we not keep the new code match the code already existing in the project? What if more contributors keep insisting on keeping their personal style on files added by them? even I have many personal preferences for writing codes which are not according to style for this opensource project but still use them on personal projects and I always keep trying not to incorporate them, but this should not mean that I should get bored or it becomes a waste of time.

digant73: but I agree to use a common style (standard rules are present since C was born). The problem is that in the whole project I see different styles (often home made) even inside the same file. Currently there is no style for this opensource (it was open even a bug to discuss/adopt a style for the project). That was my explanation. If a style for this project is provided I would apply it. In the absence of that I simply said that it is a waste of time for all of us. It's not a blank line a violation to style guide but the problem relies in a mix (often home made) of styles used by a lot of (often occasional) contributors.

PR #1407, opened 2020-12-19
add M73 support [WIP]

belese:
### Description
Add M73 support.
Format M73 PXXX RXXX  (P: actual progression of print (%) - R: remaining time (minute)

PrusaSlicer (and probably others but i don't use) can add progress gcode, and at least for PrusaSlicer and my printer, time is very accurate.

When a print start, if a M73 command is see, it will switch to "slicer_progress mode", and don't use internal progress.
if we are in that mode, the printing gui display the remaining time instead of ellapse time (if set)

### Benefits

Get an accurate print time

### Related Issues

I've some questions, 
my merlin didn't accept the M73 command, and i imagine most Printer with this screen has marlin compiled without LCD

In this code, i didn't forward it to Marlin, and i don't know if it make sense to forward it (if two screens on printer?)

But in more general case, is a blacklist of gcode to not send in printer in config.ini a good option?

it could be useful for others commands too.
my slicer add G21 to set in mm.
and marlin is not compiled with  INCH_MODE_SUPPORT, so don't accept the M21 command.
so i've got an warning each time i start a print.

PS.
Sorry, make a mistake with git and commit are not good.
If you are interested with it, il will redo a clean PR

### DIFF
--- a/Copy to SD Card root directory to update/Language Packs/language_ru.ini
+++ b/Copy to SD Card root directory to update/Language Packs/language_ru.ini
@@ -289,3 +289,4 @@ label_print_summary:Время печати: %02u:%02u:%02u\nПотрачено 
 label_print_complete:Печать окончена
 label_ext_templow:Температура сопла ниже\nминимальной (%d℃).
 label_cold_ext:Экструзия холодным\nсоплом запрещена
+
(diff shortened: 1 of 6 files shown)


ETE-Design:
@belese  Neat feature to have :-) Would like to try it, so could you pls. merge with newest version?

belese:
I redo a clean PR.
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1410

Issue #1409, opened 2020-12-19
[BUG] During printing not possbile change paramters


DanielSan1000:
Description
During printing (when started) not possbile to change paramters
- change fan speed
- baby step
- print speed/load

Expected behavior
It should be possbile

Hardware Variant
BIGTREETECH-SKR-mini-E3
BIGTREETECH-TFT35-V3.0
Marlin 2.7.0.2

radek8:
Is the More icon not working for you?

DanielSan1000:
> Is the More icon not working for you?

Is strange because after some times it is possible to change parameters. Changes are not working immediately.

radek8:
you are allowed in Marlin?
REPORT_FAN_CHANGE (in Configuration_adv.h)

kisslorand:
It is possible to change them but it takes time. The most important thing during printing is the printing itself. Other messages are sent to the printer only  when it's possible. I agree, it can be improved, I also find it cumbersome to set the flow and speed during print. However Baysteps ar working just fine for me.

oldman4U:
The delay is a known issue. Also changing feedrate and so is delayed. Printing fron on board sd is even worse.

oldman4U:
Please see #1139 and please help the community and close the ticket once you do not need it anymore. Thank you

DanielSan1000:
Thanks for help

kisslorand:
Even if it's closed, I started working on this. There's room for improvement, we'll see how much...

oldman4U:
DanielSan1000. Please give #1417 a try. It works GREAT;-)

DanielSan1000:
> DanielSan1000. Please give #1417 a try. It works GREAT;-)

I will check it and will give feedback

Issue #1139, opened 2020-09-28
[BUG] in print fan % change issue


BlackDragon2020:
### Description

If you change the fan % while printing you have to change with full, half, off or you have to change in small increments of 1% then wait a few moments to increase another 1%. If you try to hit the increase button more than once it looks like it work but after a few seconds it will change to show it only it increases 1%.

### Steps to reproduce
Start a print
Change fan speed with 1% steps add 5% or 10%

**Expected behavior**
To change the fan speed by increased amount

**Actual behavior**
% jumps back down to original value or original value +1%

### Hardware Variant
Tft35 e3 v3
Current build download today and updated no change.

Will make video if need be.

oldman4U:
This is "normal" behavior and there have been other tickets related to it. It affects a more busy system more and also happens with other settings like print speed and flow.

traffic-light:
What oldman said. 

The mainboard is busy processing data. A microcontroller is not like a computer. It can't run tasks next to each other.

(Close the ticket when not needed anymore ;-) )

BlackDragon2020:
I did not have this issues with my old screen and stock board. Not sure if it's related to the tft or the skr. 

Seems to me after reading @oldman4U the code just needs tweaked. Instead of reading the user input and changing the % instantly it should do something like this rough code:
If Input change
Loop Read input, end after input stops changing for x seconds
Take last input send that as the change.

That seems it should fix it, I will poke around in the code. I have been but it's been years and I'm rusty and never got into GUI, but I maybe able to address this.

traffic-light:
I serial always on for you?

#843

oldman4U:
Resources are very limited on the TFTs and a numpad would be the best solution I believe also.

BlackDragon2020:
@oldman4U I don't know if it's because I'm rusty and not remembering how to do stuff 100% or b cause it just so full already, but a small function with 3 variables added to the fan menu cause a flash overflow problem. What it did was do a quick loop that counts the number of times you push the increase or decrease button then pass that to the switch code, but I can't get it to compile due to flash over flow and it's 20 bytes lol. I will keep playing around see if I can find out what's up.

oldman4U:
Well. You are a black dragon and I am a old man. So you have to know and do it...😁

traffic-light:
The flash overflow is a known issue. Btt is working on it by removing all theblanguages on compile. At least that's sort of what they told.

[...]

PR #1417, opened 2020-12-21
Flow,speed,fan - reworked

kisslorand:
### Requirements

BTT TFT or MKS TFT

### Description

This is a major rework of the  Speed, Flow and Fan adjustment. It uses a different approach of stacking consecutive speed, flow and fan related commands. It is making sure the desired value is sent to Marlin, feedback is used. No more fiddling and guesswork.

### Benefits

Users can now adjust  the FeedRate, FlowRate, FanSpeed during print without hiccups, the adjustment is straight forward. the desired and actual values are displayed so no guessing what's going on, just as in the heating menus (actualValue/desiredValue). User can abuse the adjustment buttons, desired value will refresh instantly. Rapid firing those buttons will not clog the print queue, only the relevant values are sent.

### Related Issues

#1139 #1409

### Remarks

The unit of the adjusted values are displayed above the "actualValue/desiredValue" because of design constrains. It's the best I could come up with.


digant73:
I would use the prefix "target" and "current" for the variables and funtions names instead of "set" or "cur"

kisslorand:
Good advice. I started with "desired" and "current" than changed my mind because of length considerations.

oldman4U:
@kisslorand I have no idea what you have done, but it works GREAT!!!

Tested printing from on board and TFT based storage and the values are taken over reliable, does not matter how fast I click or when I use the keypad.

The only problem I have seen on MY machine is, when I increase the Flow value over 800%, something which is completely useless but hey, it is possible to change it;-) So changing above 800% makes my motor grind and higher values stop the motor. Likely other motors or TMC settings will make this problem occur earlier or later, so I am fine with it. This is a HUGE step forward, THANK YOU!

Once released, I will make some additional tests with the print finished function (Marlin update is still pending...)

@bigtreetech Please merge this great fix asap.

kisslorand:
Thanks for the feedback. 
The "problem" you mentioned is rather a misuse of a feature. There is "SPEED_MAX 999" defined in the TFT's firmware. The flow rate is not an absolute value, it is a factor. If let's say extrusion speed is at 1mm/s than a FlowRate of 800% would make it 8mm/s which is still reasonable. However if you are already at 10mm/s than it would be absurd to have 80mm/s extrusion speed.
If needed, "SPEED_MAX 999" can be reduced to a reasonable value like "SPEED_MAX 200" but max speeds should have been defined according to the machine's spec already in Marlin. It's Marlin's job to supervise max speeds. 
The TFT FW is not machine specific, each machine should have it's Marlin configured properly to its components capability and limits.

oldman4U:
I can not see this is real problem and agree 100% what you say.

Really cool;-)!!!

PR #1427, opened 2020-12-23
Add coding standards, language pack size check during compiling, fix incorrect temperature for mixing hotend, reduce program memory usage

### LINKED ISSUE #1419: Multi extruder Cold extrusion prevent
KakaoTalk_20201223_081853317

I am using a multi 4 color extruder.

Raise the hot end temperature and test Extruder No. 1, and it works.

After that, if you change it to 2,3,4, an error is displayed as shown in the figure.

I want to know how to fix it.

There were no errors in previous versions of the build.

guruathwal:
- Add Coding standards.
- Reduce code and program memory usage.
- Add Language pack size check during compiling.
- Fix incorrect temperature for mixing hotends / multi extruders with single hotend #1419 
- Update language pack size for TFT70
Fixes #1419


digant73:
In some menus you now passed an unitialized buffer to numPad funtion. Also in PID menu the tool index is now wrongly set  befor the loop detecting it. Just a very quick check, so probably other errors are present. 
I would avoid that comment style because it will require frequent indentation changes when a new or a change in a code line is made. Also in case the code is well commented (as it should be) it makes the code pretty much distracting and confusing.
The conditional code must be aligned to the rest of the code. So the best way is to always put conditional clause (e.g. #if._.) on column 1. That solution allows nested conditional code more easily. It was also adopted on ST libs if I am not wrong or in general in any commercial sw

traffic-light:
Small feedback on the guidestyle:
* Header files: instead if define and ifndef we should use #pragma once (already implemented in my pr)
*  When you described camel case you said that we can use _ in between words. I think it's better to use one. Only camel case or underscore for variables

guruathwal:
@digant73 Editing code line will affect the comment alignment in every style even when the comments are aligned to the far right side. it will also require frequent indentation changes when a line is changed.  The benefit of the style suggested in this PR is that it can be easily programmed via clang-format to auto-indent the comments removing the need to manually fix everything manually.
The Marlin Firmware prefers Indenting the preprocessor directives with the code.

@traffic-light  I did not add `#pragma once` as you already have a PR for that. You are right about the camel case. I added the _ as it is already used in the firmware in a lot of places. But it should be removed from the guide.

PR #1431, opened 2020-12-24
2020 xii 24 merge fix

kisslorand:
The merges from 24.XII.2020 broke PR #1417. This PR fixes it.
@bigtreetech please merge this


TFT/src/User/Menu/Speed.c
```diff
@@ -52,8 +52,7 @@ void menuSpeed(void)
 
   percentageItems.items[KEY_ICON_5] = itemPercent[percentSteps_index];
 
-  storeCmd("M220\n");
-  storeCmd("M221 D%d\n", heatGetCurrentTool());
+  storeCmd("M220\nM221 D%d\n", heatGetCurrentTool());
```
bigtreetech, line 55: `storeCmd` will directly store the commands in the cache and send them directly. Therefore, it will send out two gcodes at one time. Therefore, it is better to store them separately or use `mustStoreScript.` This function will automatically split the gcode commands according to `'\n'`

kisslorand: Sending more than one command at once with storeCmd() is also used in sendParameterCmd() where it is working just fine.

Issue #1433, opened 2020-12-24
[BUG] Screen freezes in Marlin mode while starting to draw status screen.


kad:
### Description

I have two TFT35 E3 V3.0 and both after recent firmware upgrade started to behave problematically.
Reverting to 1bd14113e088aac2430f2b8152784e361bebdec0 state helps.

Overall: display is used in touch mode - it is ok. But if it is in Marlin mode and printer boots with both EXP3 and 5pin TFT cables connected - once boot sequence is over and marlin starts drawing status screen (about 1/4th) the TFT freezes. it is impossible to change something or click something, impossible to change modes.

### Steps to reproduce

1. latest master build
2. SKR mini E3 board, 2.0.7.2 release, CR10_STOCKDISPLAY config
3. connect by both EXP3 and TFT cable.
4. Boot with Touch mode by default. verify that it works (moves, printer info, fan control, etc).
5. Switch to Marlin mode
6. Press reset button to reboot.
7. Observe boot sequence in marlin mode (custom boot screen, marlin boot screen, status screen).

**Expected behavior**
No freezes, correct drawing in Marlin mode.

**Actual behavior**
If booted with Marlin mode:
Custom boot screen shows, Marlin boot screen shows. Status screen starts to appear and display freezes.
Screen shows 1/4 of status screen, rest is from Marlin boot screen. No response on clicks.

If booted with Touch mode: it works 90% of times. Some times, after boot trying to initiate move commands from menu (e.g. Home), leads to screen freeze.

I tried to revert firmware to commit 1bd14113e088aac2430f2b8152784e361bebdec0 It works without issues.

In addition: while printer is booting and tft in marlin mode, it is possible to switch to touch mode (before status screen appear).
However, most often that leads also to tft freeze once it establishes serial link to the board. Reseting and booting completely in touchscreen mode works. 

### Hardware Variant
TFT3.5 E3 V3.0

### TFT Firmware Version & Main Board Firmware details
Marlin 2.0.7.2, bugfix-2.0.x same effect.

### Additional Information
* Include a ZIP file containing your `Configuration.h` or use [Pastebin](https://pastebin.com/) and paste a link in this issue.
* Provide pictures or links to videos that clearly demonstrate the issue.

kad:
![image](https://user-images.githubusercontent.com/41858/103102058-ff3dba80-4622-11eb-915c-258d88fb407a.jpeg)
[...]

PR #1359, opened 2020-12-07
Print finished

kisslorand:
### Description
 It's a tiny convenience update for the printing menu after the printing is finished. It doesn't add much just the option to go back to the File Selection menu or go straight to the Status menu.
The Pause and Babystep buttons switched place for graphical coherence.

### Benefits

After a finished print users do not have to click multiple times the Back button to get beck to the Status menu.

### Related Issues
 It isn't related to any issue.

### Additional info
In the process of this PR the automatic jump back to the Status menu after the print is finished when printing from onboard SD is removed.
There are two options for this PR after the print is finished:

- only going back (File Selection or Status menu)
- leave the Babysteps and More menu active 

I can find reason for both of them, but cannot make a logical decision.

Babysteps and More left alone:
Babysteps and More Enabled

Babysteps and More reomved:
Babysteps and More Removed

I am waiting for your feedback. Thx.

### Things evolved to something bigger so here are the final things that were done.

I. Print is Finished

1. Additions:

a. New buttons
Babystep, Pause, More, Stop buttons are removed after the print is finished, they are replaced by Main and Back button and an Infobox. Main will get you to tha Mains screen, Back will get back to the file selection menu where you selected your file that was just printed.
There's no automatic jumping to the Main screen after the print finished, the user has the choice where to go.

b. Statistics after the print is finished:

Statistics as filament length, filament weight and filament cost can be embedded into the gCode. These statistics will be presented after the print is finished in the new infobox area as a scrolling text. You can click that infobox and a popup will present you the printed filename (limited to the first 25 characters), the time needed for the print, the filament length used, the filament weight and its cost.

The statistic data in the gCode must have the following format*:

M118 P0 filament_data L:{12.3456}m
M118 P0 filemant_data W:{1.23456}g
M118 P0 filament_data C:{0.1234}

*a good practice would be to include this at the beginning of the gCode. 

L: represents the length in meters, W: represents the weight in grams, C: represents the cost without a unit.
The values of every filament data can be in a brackets, parentheses, apostrophes, etc. or without them, measurement units can be there or not. 
So M118 P0 filament_data L:(12.3456)m, M118 P0 filament_data L:12.3456meters, M118 P0 filament_data L:[12.3456] and so on are all valid formats.

The inclusion of the filament data into the gCode can be automated. In Cura all you have to do is to insert the following into the Start G-Code:
M118 P0 filament_data L:{filament_amount}m
M118 P0 filament_data W:{filament_weight}g
[...]

PR #1438, opened 2020-12-26
Fix: temperature auto-report timeout check, Fix invalid settings parameters

### LINKED ISSUE #1434: [BUG] Bad hotend and bed temperature update on status screen.
Bad hotend and bed temperature update on status screen.
If I turn on the hotend heating and the bed turns off after reaching the temperature, the temperature on the status screen is updated and it can be seen that the temperature is falling.
If I restart the printer when the temperature is reached, the printer will display a high temperature. The temperature of the hotend and the bed drops, but the display still shows the high temperature it read after the reset.

SKRV1.4 + TFT35E3V3.0
Marlin Bugfix 2.0 from today
TFT own compilation from today:
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/46d7434d8b463f160130543a2c5751aaeff1fd0c

guruathwal:
- Fix for temperature auto-report disabling due to Mainboard reset or through any other serial controller (#1434). Now it checks for a timeout after last reported temperature status and resends `M155` command.
- Update old settings signature causing invalid parameter reading due to changes in parameters via PR #1359 (related detail in https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/1433#issuecomment-751341268)
- Update config.ini
- Fix leveling feed-rate for probe height/offset control
- Clean code for marlin mode
- add missing xy_offset_probing parsing in config.c

Fixes #1434


kisslorand:
I am also working on this with a different approach. Will make a PR (Draft) also and let you know.

TFT/src/User/API/Temperature.c
```diff
@@ -210,6 +214,14 @@ void loopCheckHeater(void)
       heat_update_waiting = true;
     } while (0);
   }
+  else // check temperature auto-report timout and resend M155 command
+  {
+    if (OS_GetTimeMs() > AUTOREPORT_TIMEOUT && !heat_update_waiting)
+    {
+      heat_update_waiting = storeCmd("M155 ");
+      updateNextHeatCheckTime();  // set next timeout for temperature auto-report
```
kisslorand, line 222: I suggest:
`if (storeCmd("M155 ")`
`{`
`__heat_update_waiting = true;`
`__updateNextCheckTime();`
`}`

or
`if (heat_update_waiting) updateNextHeatCheckTime();`

So if stroeCmd() is unsuccesfull it will not wait another 3 seconds to try to resend "M155 ".

kisslorand:
Great work! It also removes the "No printer attached!" reminder after the motherboard is disconnected and reconnected.
Yummy!

TFT/src/User/API/Temperature.c
```diff
@@ -210,6 +214,14 @@ void loopCheckHeater(void)
       heat_update_waiting = true;
     } while (0);
   }
+  else // check temperature auto-report timout and resend M155 command
+  {
+    if (OS_GetTimeMs() > AUTOREPORT_TIMEOUT && !heat_update_waiting)
+    {
+      if(heat_update_waiting) updateNextHeatCheckTime();  // set next timeout for temperature auto-report
+      heat_update_waiting = storeCmd("M155 ");
```
kisslorand, line 222: `heat_update_waiting = storeCmd("M155 ");`
`if(heat_update_waiting) updateNextHeatCheckTime();`

The order should be reversed like above, otherwise "if(heat_update_waiting) updateNextHeatCheckTime()" is never executed.

guruathwal: 🤦‍♂️it was supposed to be this way but somehow the lines got interchanged accidentally.

kisslorand:
With this PR if you send M155 S0 from TFT, it will be sent periodically forever, without a stop. If you try to set M155 Sx from outside, directly to the mainboard (OctoPrint, Pronterface) the TFT will send again M155 S0, so M155 cannot be enabled again from outside, only from the TFT.
So, if M155 was switched off by a deliberate action, automatically receiving temp informations should not be checked again.

I suggest:
` if ((OS_GetTimeMs() > AUTOREPORT_TIMEOUT) && (!heat_update_waiting) && (heat_update_seconds > 0))`

AntoszHUN:
Another fix? Super! Maybe you can help me why you don't update the layer height during printing?
IMG_20201227_163056
IMG_20201227_163104

It is clear from the picture that it is not 0.8mm. True?

kisslorand:
This was fixed 9 days ago when PR #1394 was merged into master.

AntoszHUN:
> This was fixed 9 days ago when PR #1394 was merged into master.
> It was caused in PrintingMenu.c by
> `if(curLayer != (infoFile.source == BOARD_SD) ? coordinateGetAxisActual(Z_AXIS) : coordinateGetAxisTarget(Z_AXIS))`
> which has been corrected to
> `if (curLayer != ((infoFile.source == BOARD_SD) ? coordinateGetAxisActual(Z_AXIS) : coordinateGetAxisTarget(Z_AXIS)))`.

I'll see then. Thanks! I haven’t had time to upgrade lately.

kisslorand:
The fix for config.ini and inital baudrate is:

In config.ini:
`#### Baudrate`
`#Optiuns: [2400, 9600, 19200, 38400, 57600, 115200, 250000, 500000, 1000000]`
`#`
`baudrate:115200`

and in config.c:
`case C_INDEX_UART_BAUDRATE:`
`__SET_VALID_INT_VALUE(infoSettings.baudrate, item_baudrate[0], item_baudrate[ITEM_BAUDRATE_NUM - 1]);`
`__break;`

The flash is expected to store the actual baudrate, not its order in a list.
This wasn't caused by "Update old settings signature causing invalid parameter reading due to changes in parameters via PR #1359".

oldman4U:
Hi Gurmeet.

Hope you are fine.

So this PR is going to change the way the "kind" of menu used is "named", right?

From:
Unified Menu / Classic Menu
Select a UI menu flavour
Options: [Unified Menu: 1, Classic Menu: 0]
unified_menu:1
to:
Enable Status Screen
Status screen or home screen displays the current temperature, fan and speeds.
If this disabled, the main menu will become the default home screen.
Options: [Enable: 1, Disable: 0]
status_screen:1

Is there also any kind of functionality or look which changes or is it really only how it is named in config.ini?

guruathwal:
@kisslorand The parameters were read wrong after #1359 [af86e84a7](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1359/commits/af86e84a7d80a8dfa66540c6688ee9c90c98d1a7) because the setting signatures in `flashstore.c` was not updated after removal of `print_summary` which caused wrong reading of subsequent parameters listed after `print_summary`. 

Yes, the problem for the baud-rate issue was in `config.c`. The logic for reading and validating the baud-rate settings should not be as complex as it is. The baud-rate is checked against the predefined list, but in the config file, The option asks for a value instead of the index. The firmware compares the value with all the available options to finally store the value only if it exists in the predefined list. What is the point of asking to enter the value when it is the index which suffices? This also gives it ambiguity that any value can be provided for the baud-rate. The same issue can occur with the original code if the baud-rate is set to any other value not provided in the list. The logic was changed by me a long time ago when I added the config file feature but it was changed again sometime after that.

@oldman4U  yes, the icons for the classic menu and the menu layout were removed so there is no classic menu any more. The only thing that option now does is switch the status screen on and off.

kisslorand:
> @kisslorand The parameters were read wrong after #1359 [af86e84a7](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1359/commits/af86e84a7d80a8dfa66540c6688ee9c90c98d1a7) because the setting signatures in `flashstore.c` was not updated after removal of `print_summary` which caused wrong reading of subsequent parameters listed after `print_summary`.

If there's a config.ini, the reading and storing of parameters from config.ini restores the order of the subsequent parameters.
flashStore.c -> storePara()

All I said that PR #1359 is not the cause of baudrate parameter being wrong after config.ini is loaded.

Yes, you are right, any value can be provided as baudrate, which might be fine if someone uses a non-standard baudrate in its Marlin. Same goes for selecting the index of the baudrate, a user can provide also the wrong index. Using indexes makes the TFT to work only at a few predefined baudrates, which is totally fine for general users.
I do not see anything wrong storing the real baudrate nor storing the index from a predefined list. I just found the bug, I didn't created it. :)

In case of storing only the index the fix would be:

- in Configuration.h:

 `*Options: [2400: 0, 9600: 1, 19200: 2, 38400: 3, 57600: 4, 115200: 5, 250000: 6, 500000: 7, 1000000: 8]`
 `*/`
`#define BAUDRATE 5`

 - in flashStore. c ->readStoredPara():
 
 `infoSettings.baudrate             = item_baudrate[byteToWord(data + (index += 4), 4)];`

Of course there are other options too for the fix, use "item_baudrate[infoSettings.baudrate]" instead of "infoSettings.baudrate" later in the whole code but that would be counterintuitive, in that case "infoSettings.baudrate" should be renamed to something like "infoSettings.baudrate_index"

oldman4U:
kisslorand used #1359 to silently remove a feature mehmetsutas and I have been introducing after spending days and hours of developing and testing, because he does not like it. By doing so, he messed up config.ini.

@guruathwal In case there is no technical reason for removing this feature, would you please be so kind and reverse what kisslorand did and reinstall the ability to choose, if a print finished information should be shown at the end of a print or not.

Thank you

kisslorand:
@guruathwal 
Check PR ##1441, it has been reworked to adapt your ideas about using only indexes. Thank for insisting, it seems better this way, even the compiled flash is smaller now.

PR #1441, opened 2020-12-27
Fix: Baudrate config

kisslorand:
It fixes wrong parameter type stored in flash for baudrate. It now stores the baudrate itself rather than it's order in a list.


kisslorand:
Latest commit changes the of baudrate storage. The baudrate index is stored/read to/from flash and real baudrate is passed through only when needed (creating the selection list in parameter settings and initializing the serial port). 
Thx @guruathwal !

kisslorand:
Not needed anymore.

PR #1450, opened 2020-12-29
Add Automatic binary updating using github actions

Codel1417:
### Description

After every push to master this action will generate all binaries listed in platformIO and push them to master

### Benefits

Binaries are always up to date with master

### Related Issues

Binaries not defined in buildBinary.yml are deleted


Codel1417:
In the future I may look into combining buildTest and buildBinary into a single file where pushing only happens on pushing to master

PR #1459, opened 2020-12-30
config.ini cleanup

### LINKED ISSUE #1467: [Q]  Option "level_edge_distance:
@guruathwal Why did you remove the option "level_edge_distance:" from the Config.ini file?
PR #1438

### LINKED ISSUE #1474: [BUG] Typo German language Pack
Typo. Should be "Dateien als Liste" and not "Datein als Liste"

digant73:
Just a cleanup on last config.ini
fixes #1467
fixes #1474

### DIFF
--- a/Copy to SD Card root directory to update/config.ini
+++ b/Copy to SD Card root directory to update/config.ini
@@ -140,10 +140,10 @@ files_list_mode:1
 # NOTE: The OFF value is applied to any ACK message type (e.g. even to known echo ACK).
 #       It means that any kind of ACK message is silently discarded
 #
-# Options: [OFF: 0, POPUP: 1, TOAST: 2]
-#   OFF:   No notification. The message is ignored
-#   POPUP: Display a popup window for user confirmation
-#   TOAST: A non-blocking Toast notification is displayed for few seconds. No user interaction is needed
+#   Options: [OFF: 0, POPUP: 1, TOAST: 2]
+#     OFF:   No notification. The message is ignored
+#     POPUP: Display a popup window for user confirmation
+#     TOAST: A non-blocking Toast notification is displayed for few seconds. No user interaction is needed
 ack_notification:1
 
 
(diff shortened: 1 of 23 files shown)


Copy to SD Card root directory to update/config.ini
```diff
@@ -55,12 +55,12 @@
 
 #### Enable Status Screen
 # Status screen or home screen displays the current temperature, fan and speeds.
-# If this disabled, the main menu will become the default home screen.
+# If this disabled, the main menu will become the default home screen
```
thisiskeithb, line 58: Why remove periods in some places & not others? The line above this still has one.

Copy to SD Card root directory to update/config.ini
```diff
@@ -231,12 +231,12 @@ fan_ctrl_count:0
 max_temp:T0:275 T1:275 T2:275 T3:275 T4:275 T5:275 BED:150 CHAMBER:60
 
 #### Cold Extrusion Minimum Temperature
-# Minimum temperature need to extrude/retract. Any extrude/retract below this temperature will be prevented.
+# Minimum temperature need to extrude/retract. Any extrude/retract below this temperature will be prevented
```
thisiskeithb, line 234: Same here. Periods are used to end sentences, so why not keep them?

digant73: It is removed always on the last sentence. So if you have only one sentence, "." is not used (as it was in almost all the config.ini file). If there are more sentences (as in one you reported), "." is removed on the last one (it means there is no continuation)

rightawy: Periods should be added instead of removing them to make the sentence look complete. why would there be a pattern for not having the periods on the last sentence? if it is missing somewhere then it should be added not removed from other parts. Deriving patterns from everything is not a great way to follow. Somethings look better when they are a little bit out of order.

thisiskeithb: > Periods should be added instead of removing them to make the sentence look complete.

I agree. Periods end sentences and removing them here is odd.

digant73: It is not a problem at all for me to adopt one solution or another but it should be applied uniformly on the whole file. In most part of the file it was omitted. If we should make the file more readable it should be most important to use a separation between a description and param format qualifier instead of scratching everything as it is now

PR #1463, opened 2020-12-30
Improve "Build Binaries" and "Build Test" GitHub Actions

thisiskeithb:
### Description

1. Copy `config.ini` when building binaries so the version located in `Copy to SD Card root directory to update/` won't have to be updated each time the main version in `TFT/src/User/` is updated.

2. Ignore changes to the following files/folders for automatic binary/config uploads:

    - .devcontainer/
    - .github/
    - Bootloader/
    - Copy to SD Card root directory to update/
    - images/
    - readme/
    - *.md

3. Ignore changes to the following files/folders for build tests:

    - Bootloader/
    - Copy to SD Card root directory to update/
    - images/
    - readme/
    - *.md

Tagging @Codel1417 as they introduced automatic binary uploads in #1450

### Benefits

Automate updating `config.ini` in the "Copy to SD Card" folder and only run CI when needed & don't upload new binaries when readmes/non-code changes are committed.

### Related Issues

None.

### DIFF
--- a/.github/workflows/buildBinary.yml
+++ b/.github/workflows/buildBinary.yml
@@ -3,7 +3,10 @@ on:
   push:
     branches: [ master ]
     paths-ignore:
+    - .devcontainer/**
+    - .github/**
     - Bootloader/**
+    - images/**
     - readme/**
     - '**/*.md'
 
--- a/.github/workflows/buildBinary.yml
+++ b/.github/workflows/buildBinary.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Copy Files
         run: find .pio/build/ -name '*.bin' -exec cp -vf '{}' "./Copy to SD Card root directory to update/" ";"
       - name: Stage Changes
-        run: git config user.email "actions@example.com" && git config user.name "Github Actions" && find "Copy to SD Card root directory to update/" -name '*.bin' -exec git add {} \;
+        run: git config user.email "${GITHUB_ACTOR}@users.noreply.github.com" && git config user.name "${GITHUB_ACTOR}" && find "Copy to SD Card root directory to update/" -name '*.bin' -exec git add {} \;
       - name: Push
         uses: actions-x/commit@v2
         with:
--- a/.github/workflows/buildBinary.yml
+++ b/.github/workflows/buildBinary.yml
@@ -13,7 +13,8 @@ on:
 
 jobs:
   main:
-    name: Main
+    name: Build Binaries
+    if: github.repository == 'bigtreetech/BIGTREETECH-TouchScreenFirmware'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
--- a/.github/workflows/buildBinary.yml
+++ b/.github/workflows/buildBinary.yml
@@ -6,6 +6,7 @@ on:
     - .devcontainer/**
     - .github/**
     - Bootloader/**
+    - Copy to SD Card root directory to update/**
     - images/**
     - readme/**
     - '**/*.md'
--- a/.github/workflows/buildBinary.yml
+++ b/.github/workflows/buildBinary.yml
@@ -2,6 +2,11 @@ name: Build Binaries
 on:
   push:
     branches: [ master ]
+    paths-ignore:
+    - Bootloader/**
+    - readme/**
+    - '**/*.md'
+
 jobs:
   main:
     name: Main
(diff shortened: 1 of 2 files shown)


Codel1417:
I like this, i need to open another pr to copy the config.ini so we are not maintaining 2 versions of it

thisiskeithb:
> I like this, i need to open another pr to copy the config.ini so we are not maintaining 2 versions of it

What about changing the build action to zip up & tag releases instead? That'll keep the repo size down as well.

Codel1417:
i worry about posting releases too fast for little features + handling release names. I would like to see more releases though

oldman4U:
bit off topic and maybe just my 2 cents, but at the moment we have 2 config.ini files and maybe it would be possible to remove the one under the _TFT/src/User_ folder and keep only the one in the _Copy to SD Card root directory to update_ folder

thisiskeithb:
> bit off topic and maybe just my 2 cents, but at the moment we have 2 config.ini files and maybe it would be possible to remove the one under the _TFT/src/User_ folder and keep only the one in the _Copy to SD Card root directory to update_ folder

I’d keep the one in the `TFT/src/User/` folder so it doesn’t become “out of sight, out of mind” and like @Codel1417 mentioned, just script an action to duplicate it to the `Copy to SD Card root directory to update` folder on each automatic build.

thisiskeithb:
Speaking of the `Copy to SD Card root directory to update` folder, that could probably be excluded from the CI/build test too.

I’ll update the PR when I’m back at a computer and can push another commit.

thisiskeithb:
The last commit limits binary building to only run in this ([bigtreetech/BIGTREETECH-TouchScreenFirmware](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware)) repo, since you end up with double commits/replacement when syncing or rebasing your fork.

thisiskeithb:
@Codel1417: The last commit copies `config.ini` from `TFT/src/User/` to `Copy to SD Card root directory to update/` when running the build action.

oldman4U:
Sorry to be such a pain, but can I ask why the two config.ini's are needed?

I am not a complete newbie but it happened to me twice that I edited the one and copied the other to the SD card. It is just confusing to have the same file twice.

Codel1417:
> Sorry to be such a pain, but can I ask why the two config.ini's are needed?
> 
> I am not a complete newbie but it happened to me twice that I edited the one and copied the other to the SD card. It is just confusing to have the same file twice.

Its easier for users that want something to just work without looking for files.

oldman4U:
Not sure I understand.

Easier it would be to have one, like it is the case for all other files.

Codel1417 <notifications@github.com> schrieb am Do. 31. Dez. 2020 um 23:39:

thisiskeithb:
> Easier it would be to have one

I agree.

Ideally, the `Copy to SD Card root directory to update` folder wouldn’t exist and BigTreeTech would use GitHub’s Releases feature to attach precompiled firmware, config, images, etc. for SD updating as well as the source code, but that’s not the process here.

The duplicate `config.ini` issue exists because one sits along side `Configuration.h` in the source code folder for developers and the other sits in the `Copy to SD Card root directory to update` folder alongside the firmware binaries so it’s in one place for users.

Since this PR now duplicates `config.ini`, it’s a non-issue.

oldman4U:
A developer is used to work with files in different folders. The common
user is having issues to find the compiled firmware once he came to the
point being able to compile at all.

I am afraid that you implemented a solution for a problem which finally
does not fix the root of it.

As always - just my 2 cents as a dump user.

Keith Bennett <notifications@github.com> schrieb am Fr. 1. Jän. 2021 um
09:18:

guruathwal:
@oldman4U  as explained by @thisiskeithb, while navigating through the file list tree in VS Code and when there are many files opened in the editor it is easier to find **config.ini** when it is in the **src** directory. The config.ini file in **Copy to SD Card root directory to update** is in a different hierarchy.  A copy is placed in **Copy to SD Card root directory to update** so that all the files needed by the user to flash/update the device are at one place.

@thisiskeithb please add `'**/*.ini'` also to `paths-ignore:` in buildBinary.yml

oldman4U:
I am concerned about the situation once the average user finished the download and starts working on it.

thisiskeithb:
> please add '**/*.ini' also to paths-ignore: in buildBinary.yml

I pushed this change initially, but there's a potential for the latest `config.ini` file to not get pushed out to the `Copy to SD Card root directory to update` folder since the "Build Binaries" actions wouldn't run if that's the only file edited in a PR.

PR #1471, opened 2021-01-02
Fan Menu: change to fix fan set speed not updating when sending an M106 from a pronterface.

NAPCAL:
### Description

An issue with the fan set speed not updating with the current fan speed when sending an M106 command from a pronterface.

### Benefits

When wanting to increase or decrease the fan speed from the touchscreen fan menu after an M106 send via pronterface. The menu would currently hold the last set value via the fan menu and not the new current fan speed.
So if the menu set speed is 0 and the issued M106 sets the speed at 45, increasing via the menu would start from 0, instead of 45.

### Related Issues

none


kisslorand:
This will broke flow of the Fan adjustment. It was like this (as you changed it) before but it got reworked. The fanSetSpeed variable is the variable that stores the speed the TFT wants to be set. M106 from outside sets the current speed, not the speed the TFT wants to be set.
Just test your PR while printing and adjust rapidly the fan speed from TFT. You will get inconsistent readings and value jumping up and down.  I've been there, spent a lot of time to smooth out Fan speed adjustment from TFT.
If ain't broken, don't fix it.

If you want to send M106 from Pronterface, just get out of Fan speed adjustment menu. If you are in that menu it means YOU want to adjust the Fan speed from TFT and DON'T WANT that some external M106 to mess with the target speed that YOU want for the fan.

NAPCAL:
Thanks for the feedback. The correction was for the TFT to match the current M106 setting no matter the source.

Please check the video; the TFT firmware does this without the change; when you exit the fan menu to the status screen the re-enter the fan menu, you will see the set fan speed matching the current fan speed. So the current set speed doesn't stay with the last TFT fan speed setting.

Being able to adjust the current fan speed will help in fine-tuning parts cooling during the print. Or if you started a print and forgot to turn it down, up, or off.

kisslorand:
Did you check adjusting it rapidly whole printing?

NAPCAL:
Once again thanks for your feedback. If it makes thinks worse then don't commit it to the master. But please provide a better way of fan control. Your statement that the variable is to hold what the TFT set speed would hold water if when you exit the fan speed menu and a
re-enter that menu it would really hold the fan speed values! 

Maybe you need to submit a pull request to comment this line out in the menu fan.c
Line 59 
fanSetSpeed(curIndex, fanGetCurSpeed(curIndex));

Because it makes the fan set speed to match the current fan speed.

kisslorand:
Did you check adjusting it rapidly while printing? What about the situation I mentioned?

NAPCAL:
No. Please post a video of adjusting +/- 1% over and over so we all can see the effect on a print.

If I were going to adjust a cooling fan, I wouldn't be doing at 1% or using the encoder; I would tap on the center status and directly enter a speed.

kisslorand:
OK. I have nothing new to add. We've been there, we already saw it.

NAPCAL:
You might be miss understanding.

The current code (without my change) changes the current fan set speed to match the current fan speed when entering the fans menu. 

Any M106 issued before entering the fans menu will be the set speed when entering the fans menu.

It is only when in the fans menu when an M106 is issued that it doesn't update the set fan speed but does update the current speed via the M106 being echoed back to the TFT.

My change was only to make sure you are adjusting from the current fan speed. Yes, if there is an M106 issued for a 20% speed and you what it to be 80%, then you would adjust from the numerical input and not the +/- or encoder.￼

kisslorand:
I do not misunderstand the code. I worked enough time on Fan speed, I had even a few merged PR's about it, one had the exact changes you made. It was unpopular. I really do not have anything else to add. We've been there, we saw it.

PR #1483, opened 2021-01-05
Add Z Steppers Auto-Alignment (G34)

### LINKED ISSUE #1150: Include G34 Auto-Alignment to Movement/BedLevel UI
Please include the command G34 into the bedlevel menu for better user-experience for user with more than one Z-axis.

Guilouz:
Enable/Disable via config.ini
Disabled by default.

Resolve #1150


ETE-Design:
Great to see this feature implemented :-) Is it possible direct from the display to tell where on the bed it should meassure?

Guilouz:
> Great to see this feature implemented :-) Is it possible direct from the display to tell where on the bed it should meassure?

Not easy because XY positions for mesure also depend on the offset of the bltouch, i don't know what a bad value can do but one thing is certain is that at compilation if the values ​​are not correct it falls into error

The best way is to use probe limits by default in Marlin.

ETE-Design:
@Guilouz Was just so you don't have to flash Marlin until you have found the spot you wanted.

Guilouz:
> @Guilouz Not a big deal, but was just so people don't have to flash Marlin a coulpe of times of they don't know the exact meassure point they want to use from the start.

No needed to flash Marlin every time just enable #define Z_STEPPER_ALIGN_XY in Marlin and use M422 command to adjust https://marlinfw.org/docs/gcode/M422.html

radek8:
After all, it is not a problem to set these values in the Parametersetting.c menu
Just know what format the data should be.

ETE-Design:
Mabye make 2 value boxes for each side, to input x/y values, a button to each named check so it places the sensor over each point... Then a save Button to save the values...

radek8:
If you have 2 stepper motors on the Z axis,
can you enter the answer to your question here?
M503
M422

Guilouz:
> If you have 2 stepper motors on the Z axis,
> can you enter the answer to commsnd here?
> M503
> M422

No there is no M422 command with M503 report, I try it

It quickly becomes hellish because Marlin allows up to 4 steppers Z1, Z2, Z3 and Z4 with each of the X and Y coordinates for this function and since it's a feature that not everyone can use (one Z stepper or double stepper on same driver or no bltouch), I did not complicate myself.

radek8:
You're right.
Therefore, once the sensor points are defined, there is no need to change it.

Guilouz (commented):
Update flashStore.c

Guilouz:
@bigtreetech Can you merge it please ?

PR #1468, opened 2020-12-31
save/load infoSettings Struct instead of save/load every variable individually, Fix Screen freezes/Click to Resume message/damaged M105 command in Marlin mode

guruathwal:
- The settings are stored in the last 2KB of the program memory. earlier a separate entry was used to save/ load each variable. With this the **infoSettings** Struct is copied at once to/from buffer instead of saving/loading every variable individually, removing the need to add save/load entries for each variable. This also eliminates bug related to wrong/invalid settings being loaded. The prebuild settings size check now works accurately.
- ~Add missed level_edge_distance (#1467)~
- ~Prevent initial gcode queries when not in touch mode (This may be causing freezes due to long M503 response and other initial gcode responses when the TFT boots in marlin mode.)~
- Fix Screen freezes in Marlin mode (#1433) cause by serial connection not getting initialized while booting in marlin mode after PR #1438).
- Fix Click to Resume" message in marlin mode (#1383) and "damaged M105 command" (#1297) caused by `M105` delay set to **0** in PR #1244.
- Cleanup **Configuration.h**

Fixes #1433 , fixes #1383, fixes #1297

PR #1488, opened 2021-01-08
Onboard SD print - Long filename; TFT aware of other printing host

### LINKED ISSUE #1245: [Q] 8.3 file names on TFT35 printing screen when print from board SD
On the print screen top left corner I can see only the current gcode file in 8.3 format when I print from the motherboard's SD card. In the file list when I browsing the sd for the *.gcode the long filenames are appears but when I choosing one in the popup window also the 8.3 format appears and then on the printing screen too.

I'm using a TFT35 v3.0 with latest master branch FW (TFT35, firmware, lang, config.ini are updated to latest)
and SKR v1.3 with latest bugfix-2.0.x FW

in config.ini:
long_filename_support:2
in board Configuration_adv.h:
#define LONG_FILENAME_HOST_SUPPORT
#define SCROLL_LONG_FILENAMES

What else am I set for display long file names everywhere?
Thanks

kisslorand:
### Requirements

BTT TFT or MKS TFT

### Description

If printing from onboard SD (from TFT), printing screen title and filament data displays the long filename (if exists).
I spent more than 10 hours but couldn't make it to display long filename if using Octoprint (or any other USB host) and printing from onboard SD. Sorry for that.

With this PR the TFT is aware when onboard SD print is started from another host (Octoprint, Pronterface, etc). File name, print time, filament usage statistics are shown even if printing from onboard SD from another host.

### Benefits
 Everyone printing from onboard SD using the TFT or another host (Octoprint, Pronterface, etc).

### Related Issues

Fixes #1245

### RAM & Flash impact

**RAM** -> static: ~~256~~ 0 (ZERO) bytes extra, dynamic: (theoretically) 2 bytes less
**Flash** -> 276 bytes extra 

### Additional info

NO, the things introduced in this PR are NOT user selectable, there is NO adjustable thing in config.ini nor in parameters settings. There's NO additional popup. Nothing was removed.


TFT/src/User/API/Vfs/vfs.h
```diff
@@ -18,6 +19,7 @@ typedef enum {
 typedef struct
 {
 	TCHAR title[MAX_PATH_LEN]; // max path depth
+	TCHAR cur_file[255]; // current file selected for print
```
guruathwal, line 22: Why did you add `cur_file` when `title` already exists for the same purpose? This is not a good use of already limited RAM.

kisslorand: Because I cannot change title to long filename as it is used in multiple places for different purposes. I also first started to use only an index of the current file and use it to get the long filename based on that index (from infoFile.Longfile) but that way I couldn't store the filename if printing from other USB host (ex Octoprint) that uses onboard SD.

guruathwal: There will be no info about the file list in Onboard SD if the print is started by another host.  In **parseACK.c** `cur_file` is assigned same value as `title`. how is it different?

kisslorand: There will be info about the file list in Onboard SD if the print is started by another host. Just test it, I did. In fact if you start the print from another host than TFT and you print from the onboard SD, you will have the Print Menu active when print is started, just as you would have if you start the print from the TFT.
In **parseACK.c** `cur_file` is assigned almost the same value as `title` ONLY IF `cur_file` is empty. It's empty because the file selected for printing was not done on the TFT but on another host. M33 is useless once you start the print from another host, hence I cannot retrieve the long filename in that case.

guruathwal:
The same result can be achieved at the cost of just 4 bytes of RAM.

kisslorand:
Rewritten with the original idea to use only indexes. It uses now indeed only 4 bytes more RAM. I finally find a way to use only a file index and in the meantime do not disturb the file name when printing from an outside host (Octoprint, Pronterface, etc).
Filament data also displays the long filename if printing from onboard SD.

As I side note, I still couldn't find a way to display the long filename if using onboard SD from an USB host.

A short video will follow to show how it behaves when using the TFT and when using an external USB host.

kisslorand:
Video showing how it works. This PR shows the long filename when TFT is used for onboard SD printing and doesn't interfere with filenames when using other USB host (in the video Octoprint from OctoDash).

https://www.dropbox.com/s/9ej0ku0cz3vyukx/PR1488.mp4?dl=0

blueeagle69:
Brilliant.
Just what I wanted.
Will be trying this as soon as possible.

kisslorand:
Thank you.

kisslorand:
Latest commit brings extra static RAM usage to 0 (ZERO) and makes the TFT aware that onboard SD print was started from another host (Octoprint, Pronterface, etc).

blueeagle69:
Had a strange issue with this firmware flashed to my Biqu B1 (TFT35).
I realised that the printer wouldn't boot(Marlin) unless I changed to TFT mode. Once I switched to TFT Marlin would boot in the background as normal.
If I started the printer while the screen was in Marlin mode it would result in a black screen.
Flashed back to the latest master and the printer booted as normal while in Marlin mode.

Will try again tomorrow with this build.

Thanks.

kisslorand:
@bigtreetech 
This PR is ready to be merged, no conflicts with master, 0 (ZERO) RAM impact, no bugs reported nor found after extensive rigorous test patterns.

bigtreetech:
@kisslorand Hello, Thanks for your work, Is there any repetition between this PR and #1488 ?

kisslorand:
@bigtreetech 
Sorry, I do not understand the question. Right here, this one is the PR #1488.

oldman4U:
> Had a strange issue with this firmware flashed to my Biqu B1 (TFT35).
> I realised that the printer wouldn't boot(Marlin) unless I changed to TFT mode. Once I switched to TFT mode Marlin would boot in the background as normal.
> If I started the printer while the screen was in Marlin mode it would result in a black screen and wouldn't boot.
> Flashed back to the latest master and the printer booted as normal while in Marlin mode.
> 
> Will try again tomorrow with this build.
> 
> Thanks.

There has been a bug related to "serial always on" and booting in Marlin mode, which caused exactly what you describe here. This issue was fixed with #1468 I believe. kisslorand just merged the master branch into this PR 2h ago, so maybe you want to try it again.

;-)

cjj25:
I've had numerous issues with connecting OctoPrint directly to the TFT35 module. I connect via UART and found that all the HOST_ACTION_COMMANDS were not being sent to OctoPrint. The TFT35 would respond with ack: OK and append it to the string, but then send the previous line and the OK to OctoPrint on the same line causing all kinds of issues.  Examples:

Printing...OK
Bed Heating...OK

Notice how the OK is on the same line, it should have a line break so OctoPrint can understand. I only print directly from OctoPrint, whenever I tried to print I would have to click the "Fake Acknowledgement" to allow OctoPrint to start printing. 

My fix is very simple and involved changing the lines in "parseACK.c"
```
  //Parse error messages & Echo messages
      else if (ack_seen(errormagic))
      {
         ackPopupInfo(errormagic);
      }
    // if no known echo was found and processed, then popup the echo message
      else if (ack_seen(echomagic))
      {
        if (!processKnownEcho())
        {
          ackPopupInfo(echomagic);
        }
      }
```
becomes 
```
  //Parse error messages & Echo messages
      else if (ack_seen(errormagic))
      {
        // ackPopupInfo(errormagic);
      }
    // if no known echo was found and processed, then popup the echo message
      else if (ack_seen(echomagic))
      {
        if (!processKnownEcho())
        {
          // ackPopupInfo(echomagic);
        }
      }
```

This stops the TFT35 responding to the HOST_ACTION_COMMANDS and successfully relays the messages to OctoPrint.

I've posted it so users can see the fix and apply it themselves. Would it be possible for you to add a configuration option in config.ini along the lines of 

```
#### Proxy Mode
# Ignore all HOST_ACTION_COMMANDS and error messages and relay messages to connected devices.
# This allows a device (such as OctoPrint) connected to the TFT35 through UART to handle all HOST_ACTION_COMMANDS commands 
#   Options: [enable: 1, disable: 0]
proxy_mode:1
```

PR #1518, opened 2021-01-15
Added Home Offset menu + bugfixes and cleanup

### LINKED ISSUE #1281: THT35v3E3 can`t activate mesh leveling menu on touch mode
SKRmini e3v2 with TFT 35 v3e3 
1.board firmware prepare to use mesh bed leveling with 6x6 grid
2 after factory reset of motherboard then  prepare to upload TFT firmware
3 I used 3 firmwares Unified , "normal" ,Red  
 a.on unified - missing few bmp and wrong keyword mesh_min_color mesh_max_color
 b.on normal wrong keywords mesh_min_color and mesh_max_color 
 c.Red - bltouch menu present (without blt connected )
4. When I switch to Merlin mode - mesh leveling present and used to calibrate bed . Stored to eeprom, activate eeprom.
5.Switching back to touch mode - mesh leveling menu not present , mesh grid not present.
off course always factory reset , reset eeprom etc.

### LINKED ISSUE #1429: [Q] MBL (manual) Zoffset code does not change from M851 to G29 S4
Marlin 2.0 from 23.12.2020
BTT TFT from 15.12.2020
No Zprobe on my printer.  Manual MBL 3x3
Q:
Babystepping still uses M851. Why?? It need to be G29 S4 Z
Mesh editor when editing point moves beyond printing area. and I prefer manual correcting.

As I see the solving may go at least 2 ways.
1 Change all M851 Z  to G29 S4 Z and imitate Zprobe.
Also change marlin response G29 S4 (make it as M851) for easy parsing 

2 Write a new code with "MBL Adjust".

### LINKED ISSUE #1469: [BUG] mesh leveling showing wrong point index with bilinear bed leveling
### Description
with bilinear bed leveling  the mesh editor show the point index as I:1 J:65535 when trying to edit a point and the extruder will not move along the X axis only the **I** values will show correctly.

### Steps to reproduce

1. perform ABL with Bilinear bed leveling
2. go to Mesh editor and try to edit a point.

**Expected behavior**
The indexes should be updated to the actual selected point.

**Actual behavior**
the J index is always 65535

### Hardware Variant
TFT35 v2
updated to latest commit on 31-December-2020
Bilinear leveling enabled on Marlin bugfix from 28-December-2020.

### LINKED ISSUE #1534: Babystep function on real time printing don't save to eeprom
Hi everyone, I updated with one of the latest firmware with the upgrade issue for the filament runout beeps and it work fine, but when i adjust the babystep when printing, it's dont still saved on the eprom with the modified value.

But the stored message appears but the value don't be updated.....

If I make a correction outside the babysteping option and make the modification under the menu SETTINGS-----MACHINE----PARAMETER----- probe offstet.... Z.... and save the upgrade it stored on eeprom is OK.

I'm loosing something or it's a firmware bug????
regards!

digant73:
**NEW FEATURES:**
1) Added Home Offset menu. Previous ProbeOffset menu is now a generic Z Offset menu (just to not duplicate code) that will invoke Home Offset or Probe Offset API. Both the Home and Probe Offset menus are available on the Bed Leveling menu. Furthermore, Probe Offset menu is available only if probe is enabled (supported) in Marlin fw.
2) Added support to MBL Z offset (G29 S4 Zxx) in Babystep menu. Babystep menu will allow to change/save Z offset based on the following rules (matching sequence):
 *) Probe Offset: If probe is supported, Probe Offset is used.
 *) MBL Z offset: Otherwise, if MBL bl is supported, MBL Z offset (G29 S4 Zxx) is used.
 *) Home Offset: Otherwise, Home Offset is used.

**FIXES:**
fixes #1469
resolves #1429
fixes #1281 (already fixed by a previous PR)
fixes #1534 The save button didn't save the new Z offset for some bl type (e.g. MBL). Also Z offset was not properly handled for Home Offset

**PR STATUS:**
Ready to be merged

**NOTE:**
I ask people using Home Offset to test this PR and provide me some feedback for the following features/menus:

**New Home Offset menu:**
* let me know if it works properly or something has to be provided (e.g. raise the nozzle before moving to the probing point, apply also XY offset as it is possible for Probe Offset etc...)

**MBL and Mesh Tuner menus:**
The sequence followed by the current menus is:
1) Homing (G28)
2) raise the nozzle (to avoid to touch the bed when performing step 3)
3) move the nozzle to the desired probing point (for MBL, the nozzle is moved to Z 0.2 mm just to avoid to touch the bed in case manual leveling was not properly made)
4) you can adjust Z offset pressing on UP/DOWN buttons
5) repeat steps from 2 to 4 for the other probing points (for MBL only)

* let me know which of the above steps give some problems, if any.
* let me know if something has to be provided/removed (e.g. don't raise the nozzle before moving to the probing point etc...)


ETE-Design:
@bigtreetech PLS. merge :-)

guruathwal:
@digant73 Can you catch me up to speed with this? is it intended for use in place of z offset?
>New Home Offset menu (it is the Z Offset menu but it will use M206 instead of M851):
>
>- let me know if it works properly or something has to be provided (e.g. raise the nozzle before moving to the probing point, apply also XY offset as it is possible for Probe Offset etc...)

digant73:
@guruathwal based on #1429, printers not having a probe relies on Home offset to set z offset. I simply made the existing Z offset menu general. It will detect if probe is supported or not (using infoMachineSettings.zProbe). It will then use HomeOffsetControl or ProbeOffsetControl API to adjust Z offset. Unfortunately I cannot set up my printer to use Home offset so I cannot test it by myself.
According to clarifications in #1429, I made the PR but it should be clearly tested by people having that setup

sarvenn:
Hi @digant73,
I compiled your PR and have some try outs.
First of all I did a bed level on the corners using the TFT menu (manual level). My Z raise value is set to Z0 (code below) for manual leveling, default value was Z0.2 in the config. So I leveled my bed using the knobs at the corners while the nozzle is just touches to the bed. My nozzle is set to 130C and bed is set to 60C while doing this adjustment. I also keep my nozzle and bed heated while doing the MBL as well. While doing manual leveling and mesh bed leveling I do not use a paper between bed and nozzle, I lower my nozzle until it touches to bed, just at the right time. The reason I do like this is to be sure that my first layer height is absolute right. Otherwise paper thickness should be adjusted as a negative offset value while printing.  This method works really, because after completing these steps, I print a bed level stuff and layer infills were always perfect without a babystep adjustment need (tolerance is +/-0.02mm).

I added my customized config.ini file as attachment.

#   Format: [level_z_pos: Z<position in mm>]
#   Unit: [distance in mm]
level_z_pos:0
[config.txt](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5825081/config.txt)


> _New Home Offset menu:
> 
> let me know if it works properly or something has to be provided (e.g. raise the nozzle before moving to the probing point, apply also XY offset as it is possible for Probe Offset etc...)_

Added new offset menu: Nozzle goes to home position for setting the offset. Bed center would be better. Because at Ender 3 Pro, home position nozzle is off the bed. So nozzle is in the air. So just like a BL touch installed printer, this adjustment should be done in the center of the bed.

However, we are missing something more important, I just figured it out. 
Setting this offset is much more important at printing screen because it is the only way to fine tune the nozzle offset while watching the first layer, especially while doing a first layer infill. In marlin mode it is called Bed Z in print Tune menu (it changes G29 S4 when store setting command is used after adjusting the value). While printing something, only choice for me to fine tune my first layer is to go to Babystep menu at TFT mode. When I dial the encoder knob I can tune my first layer perfectly, however when I click on the save button, it doesn't update the home offset. It just resets the Babystep number and home offset value to it's preset value from epprom. After that, I have no idea what it my nozzle height relatively. Confusing.
I think it would be better to set the G29 S4 command to adjust the nozzle height from bed surface, not the home offset from MBL offset menu. In babystep menu it may adjust the home offset but whenever I click on save, it should be stored at G29 S4 Z offset(aka bed Z). It would be inline with the Marlin mode. Otherwise if I go to marlin mode and adjust bed Z it will add on to home offset value and things will become complicated and confusing.
In addition to that, in printing babystep menu, it is crucial that when I click on save that new offset value should be saved to epprom. In marlin mode, while printing, if I adjust the Bed Z and then click on store setting, G29 S4 value is updated with the new offset. Due to that new logic should not adjust home offset.

MBL and Mesh Tuner menus:
The sequence followed by the current menus is:

> Homing (G28)
> raise the nozzle (to avoid to touch the bed when performing step 3)
> move the nozzle to the desired probing point (for MBL, the nozzle is moved to Z 0.2 mm just to avoid to touch the bed in case manual leveling was not properly made)
> you can adjust Z offset pressing on UP/DOWN buttons
> repeat steps from 2 to 4 for the other probing points (for MBL only)
> let me know which of the above steps give some problems, if any.
> let me know if something has to be provided/removed (e.g. don't raise the nozzle before moving to the probing point etc...)

There is problem with the step 3. It existed in master firmware as well as I told in the issue before.
After nozzle moves to the probing point it goes down to Z0, depending on the warp of the bed on some points it touches and pushes down the bed then the nozzle raises to Z0.2 and waits for my adjustment. Expected behavior is nozzle should be only lowered to Z0.2. Pushing down the bed may ruin the static position of the bed (spring adjusted beds are vulnurable to this).

In Mesh Edit menu, after I did mesh bed leveling for 3x3 grid, I opened the mesh editor, I clicked on a value, nozzle goes to the relevent probe point. I noticed that nozzle stops at a distance to my bed even I adjusted my nozzle barely touching to the bed while doing the MBL sequence.

After these try out I encountered some serious bugs. I think there is a significant problem with the touch screen calibration, I could not click on buttons. I went to terminal screen and clicked on every number/letter with a pointy stylus pen and noticed that some squares way off calibration. For example I click on 9, it writes A on the screen etc.
I switched to Marlin mode, after a while screen went crazy. This touch screen calibration issue occured at prebuild firmware as well so I don't think it is related with your update. I think I need to open a new ticket for this serious issue.

I would like to take videos and explain these things more visually but unfortunately I can not add a video above 10MB. If you aks for a video, I may take a video and upload to youtube and give you a link. I would be happy to help and do as much as I can.

digant73:
@sarvenn thanks for your feedback

here my answers/questions:

**new Home  Offset Menu:**
1) IMHO for Home Offset, the XY home offset should not be applied (as it is possible for Probe Offset) because XY home offset are specifically  set due to specific HW constraints or limitations. However, if required by users, this feature could be easily added in a next PR as it was done for Probe Offset (based on xy_offset_probing configurable parameter)
2) in my last commit 97bab25, the bug in the babystep menu related to save button not saving the offset was fixed. However, for Home Offset it doesn't work properly because the offset is increased when raising the nozzle and decreased while lowering the nozzle (it should be the opposite as in the Home Offset menu)
3) yes I agree to change the logic in babystep menu for Home Offset change/save. My idea is to use M851 if probe is detected (so the same logic as in the current release) and to use G29 S4 Zxx in case MBL bl type is detected (so even in case Home Offset is supported). In order to use G29 S4 Zxx I need to retrieve the current offset in order I can add babystep value to it. Is there a gcode related to G29 S4 Zxx to retrieve the z offset? Or I have to parse it from output provided by G29 S0?

**MBL menu:**
1) it is the G29 S2 that will move the nozzle to the last set Z position (e.g. if previous point was set to Z-0.2 then for the next one the nozzle will be moved to the same Z position). After G29 S2 I always move Z to 0.2 just to avoid the nozzle can be too close to the bed. It seems that I cannot fix that because I cannot issues a gcode such as G29 S2 Z0.2

**Mesh Tuner menu:**
1) strange because I simply move the nozzle to the target Z height

For the calibration problem, I don't know. It seems an HW problem

digant73:
@sarvenn
I just fixed the z offset handling for Home Offset in Babystep menu. It is now handled as in the Home Offset menu. So when nozzle is raised z offset is decreased and viceversa.
If you can, please test my last commit. At least Home Offset handling should be correct.
I'm also working on the possibility (for MBL bl type) to use G29 S4 Z in babystep menu but it seems not easy to implement because I don't see any gcode command to retrieve the current z offset for MBL.
I think this PR can be merged as it is now and in the future it will be improved when possible. In the meanwhile we will collect other feedbacks from users

digant73:
@sarvenn 
Just added support to MBL Z offset (G29 S4 Zxx) in babysteo menu

ETE-Design:
@digant73 It was you who made the Mesh Editor right, really great work :-) But have noticed that in the last couple of PR merged to the Master... Now the Hotend move really slow around inside the Mesh Editor when you choose a Point to edit... The rest of the time the printer moves as expected, only in Mesh Editor it seems extremly slow...

digant73:
@ETE-Design no changes on Mesh Editor. In my printer the speed is correct as in the past

ETE-Design:
@digant73 Then it must be the New Marlin Bugfix... Do you know what to chance for then, cause the printer Works fine with normal speed elsewhere...

digant73:
Using marlin-bugfix of few days ago and no problem on that. Mesh Tuner uses G42 Ixx Jxx to move to the target point

sarvenn:
> @digant73 It was you who made the Mesh Editor right, really great work :-) But have noticed that in the last couple of PR merged to the Master... Now the Hotend move really slow around inside the Mesh Editor when you choose a Point to edit... The rest of the time the printer moves as expected, only in Mesh Editor it seems extremly slow, is this with purpose or is this an issue...

I also encountered same bug a couple of times. After seeing this issue. I tried to print something and still the print head was moving so slow. I figured out speed value has changed. I don't know how that happens. May be while turning the encoder knob at mesh bed leveling sequence. I use the precompiled marlin fw in btt repository. I will try to compile my own fw from marlin bug fix fw for my skr mini e3 and try again with this PR's latest fw and give my feedbacks.

digant73:
I don't use rotary. please try not using rotary

ETE-Design:
@sarvenn Mine only happend when I'm tyring to edit a mesh point inside the Mesh Editor, the rest of the time the printer run at normal speed...

ETE-Design:
@bigtreetech Time to look at the PR and merge some again :-)

mpcolman:
Good Job!!!

Guilouz:
@digant73 What exactly is Home offset? I use a Bltouch and now I have H offset et P Offset (formerly Z offset) in ABL menu

sarvenn:
@digant73 well done! You nailed it. New feature works great.
However I still have problems while doing the manual mesh bed leveling. Nozzle goes down to Z0 and touches to bed at some points and lifts up to Z0.2(I guess) and lets me adjust the height on that grid point.

digant73:
@Guilouz it's the Marlin feature based on M206. It is more usefull for some printers (usually based on Z max endstop).
You (as me) probably don't need it (use Probe Offset with BLtouch)

digant73:
@sarvenn Unfortunately MBL is entirely based on Marlin. With G29 S2 the nozzle is moved on previous Z. I then move to Z0.2 just to avoid the nozzle is to close to the bed.
Maybe in the future the G29 S2 will be ehnanced giving the possibility to set also the Z height

pfunl9091:
@digant73 I have little experience with firmware for prntert and etc., so I have question. Whre I can find your fixes for firmware ? Here : https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware ? For now I have Marlin-2.0.7.2 and I have a lot of bugs z offset and etc.

digant73:
@pfunl9091 move to current official Marlin 2.0.9.1 and TFT fw. The TFT fw requires at least Marlin 2.0.8.1

PR #1535, opened 2021-01-18
Notification_icons.

radek8:
For a printer, 2 hotends add an icon with the temperature of the second hotend to the notification bar.
For a printer with 1 or 3 or more, it remains unchanged

image

### DIFF
--- a/TFT/src/User/Menu/Parametersetting.c
+++ b/TFT/src/User/Menu/Parametersetting.c
@@ -366,13 +366,23 @@ int16_t drawTemperatureStatus(void)
   int16_t x_offset = LCD_WIDTH - 10;
   if (!temperatureStatusValid()) return x_offset;
 
-  uint8_t tmpHeater[3]; // chamber, bed, hotend
-  uint16_t tmpIcon[3];
+  uint8_t tmpHeater[4]; // 1-2hotend, bed, chamber hotend
+  uint16_t tmpIcon[4];
   uint8_t tmpIndex = 0;
   if (infoSettings.hotend_count)
   { // global hotend
+    if (infoSettings.hotend_count == 2)  // dual hotend
+    { 
+    tmpIcon[tmpIndex] = ICON_GLOBAL_NOZZLE;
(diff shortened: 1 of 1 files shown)


oldman4U:
Soon you will need a bigger display my friend;-)

guruathwal:
it will occupy the whole title bar when all the heaters are turned on.

radek8:
> it will occupy the whole title bar when all the heaters are turned on.


That's not a problem.
This is important information that should be displayed.
Not everyone has all the heaters, only a few users will see them all.

radek8:
> it will occupy the whole title bar when all the heaters are turned on.

If you are afraid of too long information areas.
I can return to a maximum of 3 icons, with a condition
if there are 2 hotends and no chamber, the icons 2x hotend + bed will be displayed.
Would that be acceptable to you then?

oldman4U:
It is acceptable for me either way.

Perfect would be to iterate if there is more than one nozzle as well as bed
and chamber if such exists.

Radek <notifications@github.com> schrieb am Sa. 23. Jän. 2021 um 13:23:

guruathwal:
> That's not a problem.
> This is important information that should be displayed.
> Not everyone has all the heaters, only a few users will see them all.

There is always a possibility for everything.  And it is also important to display the title/name of the current menu also. It should not hijack the space meant for displaying the title of the menu.

> If you are afraid of too long information areas.
> I can return to a maximum of 3 icons, with a condition
> if there are 2 hotends and no chamber, the icons 2x hotend + bed will be displayed.
> Would that be acceptable to you then?

I do not have 2x nozzles and chamber heater to weigh the pros and cons for displaying and not displaying the info. It could be better to have 2 icons and switch between two nozzle temps and bed + chamber temp. in this condition, all 4 info can be displayed in 30% area making it cleaner.

radek8:
OK
I returned the display of a maximum of 3 icons.
the temperature of the second hotend is only displayed if the chamber is not active.
Switching icons would also be a good solution

PR #1541, opened 2021-01-20
FIX #1539

### LINKED ISSUE #1539: [BUG] update firmware cause issue with input when screen was rotated.
I am just now getting back at it. Still a bit too busy to be of any help, but think may have found a bug. I will test for sure with in a few days. I just wanted to get it down incase I forget.
### Description

I had my screen rotated last time it was all up and running. I just updated to the newest version of BIGTREE_TFT35_V3.0_E3.26.x.bin from the pre made folder. the screen did rotate back to normal but it seems to remember the autocalibration from being flipped. top left button was giving me the bottom right button.  

### Steps to reproduce
I would think  rotate screen 
update firmware

**Expected behavior**
screen go back to stock

**Actual behavior**
screen goes back to stock, but the input for the touch is still filliped. 

sorry i dont have more info. I will get around to testing this more latter.

rixnco:
### Description

The Touch Screen Calibration was not reset when the stored parameters were reset. (ie when parameters signature was not correct).

### Benefits

The TSC is now also reset when the config is reset.

### Related Issues

Issue #1539


digant73:
Under Related Issues, report "fixes #1539" instead of "Issue #1539" so the bug will be automatically closed when the PR is merged by BTT

PR #1553, opened 2021-01-23
Update language_de.h

discip:
As the headline states.


rebel1 (changes requested):
Hi,
freistehende Zeilen werden wie Satzanfänge behandelt: Das erste Wort von Überschriften, Titeln oder sonstigen Ankündigungen beginnt mit einem Großbuchstaben. Von daher habe ich die Übersetzung entsprechend geändert. Warum möchtest du das wieder ändern ? Das sieht für mich nicht nur falsch aus, sondern ist es auch.

discip:
@rebel1
Good evening,
initially I changed that some time ago, to the state you found it, to make it look more natural, hence grammatically correct.

I have to be honest with you, I did no research prior to doing so.

But even after having done this, I am still not convinced, that the way you put it, is the only one, since there is no distinct rule, which could be applied to this kind of text. (Its not always headline, title or announcement.)

If I am searching for
>freistehende Zeilen

I always end up with enumerations! (Would be thankful, if you would provide some ***trustworthy sources***, proving me wrong.)
Even for this seemingly declared topic, there is no clear settlement.
And the type of text we are dealing with, is clearly not that!

I  absolutely do not want to start a fight over this!
And as stated above, I would be grateful, if you could provide me with some sober proof.

rebel1:
Sorry ,but I don´t wanted to start a fight over this.

For me, it looks more natural to start a menu point with a capital letter and this was good practice all over the time when i saw a german menu.

bigtreetech:
Hello, @discip @rebel1  I'm glad that we can discuss the problem rationally. Is there a conclusion about whether capital letters are needed?
I learned from the search engine that the first letter of German nouns needs to be capitalized, so most menu labels should be capitalized?

discip:
@rebel1
I am not holding on to the idea of displaying each of the words in their grammatically correct notation, if it comes to unify the overall look. Except for sentences! My intention might have been a little to nitpicky.
And no worries, if you prefer to change it back to capitals.
*By the way: I made some additional changes, other than changing capitals.*

@bigtreetech
You are right, in German nouns are capitalized.
I will have a look into this, one more time, and *eventually* change a bit here and there.

kind regards

digant73:
@bigtreetech Due to rearranged/cleanup on language pack made on my PR, in case this PR is merged, let me know if you need I will adapt this translation in my PR just to avoid a wrong merge for the two PR. I will need some time to adapt this merged PR to mine

digant73:
@bigtreetech attached this PR reworked to be compatible with my PR.
It is enough to merge this file on main branch and my PR will be already compatible.

[language_de.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5895075/language_de.zip)

rebel1 (approved):

discip:
I had to open another PR https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1593.

PR #1556, opened 2021-01-23
Reset speed&flow after print finished

kisslorand:
### Description

It's nothing fancy, just resets the speed & flow parameter after the print is finished.

### Related Issues

 
There's no issue about it, it was a discussion (#1406) started more than a month ago but there was no pro or contra, only one single like.


kisslorand:
@bigtreetech
This PR is mergeable.

bigtreetech:
Hello, thank you for your work, but if the M220 / M221 command should been added to "start_gcode" in config.ini will be more general? Because some customers will use this screen on CNC and other machines, these commands are not necessarily universal, maybe in the config.ini is more flexible than set it in the program?

kisslorand:
It can be made as a toggle parameter. I had in mind those who adjust the flow and speed during print. They might forgot about it and their next print would be ruined.
I see no harm to noone if the flow and speed is reset to 100% after the print (or any other job) is finished. 
If you put  M220 / M221 on the start gCode it might not be OK for those who already have a gCode for a given filament but want to use another one and they already know the needed Flow rate and they preadjust it before the print. 
If you put  M220 / M221 in the end gCode it's not OK, because it makes the filament length, weight and cost data to not be calculated according to the adjusted flow rate.

bigtreetech:
> If you put M220 / M221 on the start gCode it might not be OK for those who already have a gCode for a given filament but want to use another one and they already know the needed Flow rate and they preadjust it before the print.

start_gcode is executed before the gcode file that actually printed, so if the flow needs to be adjusted, M221 should be included in the gcode file, or the flow should be adjusted manually after start printing.

kisslorand:
That is right. I remember a user arguing with me that he is adjusting the flow during the print, on certain parts of the print. 
I am against flow adjusting anyway, in my opinion it is only for calibration purposes, not to use in a real print scenario.
This PR had in mind those who like to play with flow during print and might forget to reset it.
When you start the printer the flow and speed is at 100%. With this PR it would be the same for every started print job not only when you start the printer, it would give a consistent behaviour in this manner. User can still add M220/M221 wherever he wishes.
Anyway, if this PR seems useless, it can be closed, no harm done.

PR #1558, opened 2021-01-24
Reduce global variables and cleanup

### LINKED ISSUE #1568: Mesh data not displayed correctly
Installed latest firmware for BIGTREETECH-TouchScreenFirmware-master on TFT 3.5 V2
Mesh data display is one row and not a full grid.
Flashing firmware does not remove old data of stored mesh data

### Steps to reproduce

1. Flash firmware
2. navigate to : menu, movement,  bed level, mesh data
3. grid is not displayed. Only one row of select able points exists.

**Expected behavior**
9X9 grid to appear

**Actual behavior**
See above

### Hardware Variant

TFT_35_V2_.0.26.x Jan 23 2021

### Additional Information

guruathwal:
- Reduce/remove global variable to reduce RAM usage.
- Fix Busy popup preventing the user from exiting load-unload menu if machine/printer is busy or not accepting commands.
- `stringToDouble` now ignores tab character. (UBL mesh data reported by printer uses tab character instead of spaces causing only first value to be parsed) #1568 

Fixes #1568


ETE-Design:
Seems like there is also an issue with M600, the nozzle moves to the given point, and show the button reheat cause the temperature drop... But then the screen freeze and you can't use the reheat button... Only option is to shut down your printer...

bigtreetech:
You seem to have deleted the functions of "1. scrolling the gcode file name. 2.pulling out the SD card to leave the file list menu automatically"

guruathwal:
I intentionally removed scrolling gcode file name for icon mode because it scrolls only the first file name and the rest of the names are displayed normally.
I did not intend to remove the SD card part. it was something I was testing and I forgot to restore.

bigtreetech:
> I intentionally removed scrolling gcode file name for icon mode because it scrolls only the first file name and the rest of the names are displayed normally.

Hello, only one filename will be scrolled at the same time, because if all the filenames are scrolled at the same time, it will be dazzling. When you click the text area of the filename instead of the icon area, the selected file name will start scrolling.

guruathwal:
Hi, The full file can be read by clicking on the file icon. Do you think it is better to click on a tiny space to read a long name in a 10 character width and wait for it to scroll? isn't it faster to click on the file icon and read full name all at once?

PR #1563, opened 2021-01-25
Added new param in FeatureSettings menu + toggle ACK in Terminal menu…

### LINKED ISSUE #1298: [FR]  the button terminalACK
@digant73 your modification of the menuTerminal is great.
Could it be further improved with a button that would turn "terminalACK" on and off directly from the terminal environment?
Thank you

### LINKED ISSUE #1504: [FR] Access to Smart Filament Detection settings in screen settings
Would it be possible to allow the modification of the smart filament detector parameters in the machine settings on the screen?
These settings may need to be changed depending on the material specifications or printing type.
Downloading the firmware each time is not functional, especially since it is not possible during printing.

digant73:
**NEW FEATURES:**
1) Added in FeatureSettings menu the enumerated parameters defined in config.ini such as:
 * enabled, disabled
 * enabled, disabled, auto
 * etc...
 The purpose of this addition is to allow to set some settings without the need to use so frequently the config.ini.
2) Added in Terminal menu a button to toggles the terminal ACK. The changed value is not made permanent on flash (it's meant just as a temporary setting change) . This means that in order to make the change permanent on falsh the usual FeatureSettings menu has to be used
3) Added "MBL offset" parameter in ParameterSettings menu (MBL offset is already managed on Babystep menu)
4) Rearranged all the language files grouping all the entries based on:
 * config.ini parameter / feature settings
 * machine paramater settings
 * standard parameter values (e.g. ON, OFF etc... used in many menus)
 * navigation buttons
 * action buttons
 * etc...
 The purpose of the changes is to better reuse (when possible) the existing labels instead of defining new ones (often) with an already existing meaning or to put in the proper group new entries (e.g. feature parameter, machine parameter, menus etc...).

resolve #1504 (partially): custom (requiring to edit the value) paramter settings could be added in the future
resolve #1298

**PR STATUS:** ready to be merged


ETE-Design:
@digant73 Sorry to highjack this PR... Could you mabye also make an G26 Menu / Tool?

TFT/src/User/API/MachineParameters.h
```diff
@@ -10,29 +10,29 @@ extern "C" {
...
-  P_OFFSET_TOOL,
   P_HYBRID_THRESHOLD,
   // Keep below items always at the end
   P_SAVE_SETTINGS,
   P_RESTORE_SETTINGS,
   P_RESET_SETTINGS,
-  P_ITEMSCOUNT,
-  P_MBL_Z_OFFSET
+  P_ITEMSCOUNT
 } PARAMETER_NAME;
```
guruathwal, line 36: the goal should be to keep less common settings at the end and maintain the sequence as in M503 reporting and in the marlin firmware's settings menu.

TFT/src/User/Menu/common.c
```diff
@@ -82,6 +82,39 @@ const ITEM itemExtLenSteps[ITEM_EXT_LEN_NUM] =
...
+};
+
+const LABEL itemMoveSpeed[ITEM_SPEED_NUM] =
+{
+  LABEL_SLOW_SPEED,
+  LABEL_NORMAL_SPEED,
+  LABEL_FAST_SPEED
+};
+
 // Check time elapsed against the time specified in milliseconds for displaying/updating info on screen
```
guruathwal, line 118: please do not add items in common.c/common.h files if they are used in only one file.

TFT/src/User/API/MachineParameters.h
digant73: yes, I agree. The offset / ABL params are also used in different menus (Z offset menu, BL menu etc...), so frequently used/changed. For that reason I ordered (the offset params) and moved some those few params on top of the list.
Also, in the language file, I grouped and ordered all the parameters (machine param and config.ini param) in the same order they are provided in config.ini / ParameterSettings menu. In FeatureSettings menu I also show the items in the same order as in config.ini. It's more clear to navigate and see the param in the same order as they appear in config.ini. Also it make clear and esay to provide new param in the language files. I hope you agree on that.

TFT/src/User/Menu/common.c
digant73: hmm I re-used itemToggle in Z offset menu. IMHO those values could be used in many other future menus. However, if you don't agree, I can move them in FeatureSettings, Let me know

TFT/src/User/API/MachineParameters.h
guruathwal: I do not have any issue in reorganizing the language keywords. The parameter settings should be arranged according to the sequence in marlin and the most basic parameter should be on top. like for eg. majority of the machines are set up with ABL, putting P_MBL_OFFSET on top is not good. Even in the configuration.h file, the offsets, leveling points, etc are placed after the basic settings. Home offsets are set only once during the first setup of the machine, putting it on top is not good logic. The arrangement should be consistent with that of marlin firmware. In the same way in config files, most common settings are placed on top followed by less and less common/device-specific settings.

TFT/src/User/Menu/common.c
guruathwal: when the future comes they can be placed here. For now, only one file is using them

TFT/src/User/Menu/FeatureSettings.c
```diff
@@ -295,79 +442,154 @@ void loadFeatureSettings()
...
         #endif
 
+        case SKEY_HEATED_BED:
+          settingPage[item_index].icon = iconToggle[infoSettings.bed_en];
+          break;
+
+        case SKEY_HEATED_CHAMBER:
+          settingPage[item_index].icon = iconToggle[infoSettings.chamber_en];
+          break;
+
```
guruathwal, line 511: there is no point in placing these in menu. these are already being set through config file.

TFT/src/User/API/MachineParameters.h
digant73: I just arranged some params (offset tool, current, bump sensitivity) in ParameterSettings according to Marlin (sequence in ParseAck). But it's not clear to me which param you consider basic. The offset and ABL params are changed more frequently than the other machine settings such as steps, acceleratation etc (those for me were never changed because already properly set on Marlin fw).
Currently, only home, probe and mbl offset were put on top just because they seems to be the only parameters that you can change more frequently (the offset normally needs to be changed according to the env temperature).
It'is not a problem for me to follow exacly the same sequence as in Marlin but doing this it seems to me that in that case your suggestion (provide on top the most frequently used params) is disowned

guruathwal: in parseACK the parameters are arranged according to the gcode sequence numbers. When some parameters like ABL state are already available in other menus then why put them in the top? Why do homes offsets need to be changed so frequently? it is only the z offset which may change, but it also has its own menu. then why put it on top everywhere? the probe offsets also does not change after setup..

TFT/src/User/Menu/FeatureSettings.c
digant73: entries removed

TFT/src/User/Menu/common.c
digant73: moved on FeatureSettings

TFT/src/User/API/MachineParameters.h
digant73: Home and Probe offset in ParameterSettings allow also to change XY. Those offset are changed more frequently than any other setting in that menu.
A good optimization on parseAck would be to move on top of the check list the ack messages instead of the machine and capabilities that usually are triggered only at startup or to refresh the machine params.
The ack messages are typically exchanged during a print so when the system is stressed. Moving on top tha ack messages handling will improve the responsivness

sarvenn:
Is this merged with master fw? Can I install the pre-compiled fw or compile my own using platform.ini file ?

oldman4U:
sarvenn, you can click on the repo and download it from there.
Screenshot 2021-01-29 at 20 03 20

Guilouz:
Mistake in your language_keywords.h, missing ":" for this strings :

#define LANG_KEY_CONFIRM                      "label_confirm**:**" //Confirm Sign
#define LANG_KEY_CANCEL                       "label_cancel**:**" //Cancel Sign
#define LANG_KEY_WARNING                      "label_warning**:**" //Exclamation Sign
#define LANG_KEY_INFO                         "label_info**:**" //Info Sign
#define LANG_KEY_POWER_FAILED                 "label_power_failed**:**" //Question Sign

bigtreetech:
> Mistake in your language_keywords.h, missing ":" for this strings :
> 
> #define LANG_KEY_CONFIRM "label_confirm:" //Confirm Sign
> #define LANG_KEY_CANCEL "label_cancel:" //Cancel Sign
> #define LANG_KEY_WARNING "label_warning:" //Exclamation Sign
> #define LANG_KEY_INFO "label_info:" //Info Sign
> #define LANG_KEY_POWER_FAILED "label_power_failed:" //Question Sign

fixed now, thank you

digant73:
@Guilouz, @bigtreetech Thanks for the double check and the fix. Although I see now that those keywords were already wrong at the origin (in the main branch)

bigtreetech:
@digant73  Hello, Thank you very much for your help in the project. Do you have Facebook to communicate with? If it is not convenient to publish, you can send your contact information to our email “Tech@biqu3d.com”

PR #1588, opened 2021-01-31
Bugfixes, cleanup

kisslorand:
### Requirements

BTT TFT or MKS TFT

### Description

* fixed "Loading..." screen hanging after canceling a print from onboard SD
* fixed "Loading..." screen hanging after pausing and than later canceling a print from onboard SD
* fixed nozzle moving away from parking position after "Pause" when `PARK_HEAD_ON_PAUSE` is enabled in Marlin
* fixed some memory allocation not being released after no longer needed
* fixed `request_M0` not being parsed
* fixed a hidden bug that could become active in a future development (`requestCommandInfo` is used ONLY when printing from TFT and onboard SD, in no other case)
* tiny code cleanup
### Benefits

- everyone

### Related Issues

Fixes #1582
Fixes #1523
Probably fixes other or some issues related to "Pause".

PR #1580, opened 2021-01-30
Fix pausing print from onboard SD card, update French translation

Guilouz:
Pause when printing from onboard SD card have issue when using M25 command.

Marlin related issue here : https://github.com/MarlinFirmware/Marlin/issues/18191

Solution is to use Head Park command (M125).

Before :
- Pressing pause on screen
- Print is paused and Hotend is parked but go back immediately at the last position where the pause was made without user interaction
- Purge is done above the print
- Screen tell to user to Continue

After :
- Pressing pause on screen
- Print is paused and Hotend is parked
- Screen tell to user if he want to purge or continue printing
- After pressing Continue, hotend go back to the last position where the pause was made and printing

Tested on 2 printers (CR-10S Pro and Sidewinder X1), working like a charm.

Fix #488
Fix #1523

PR #1598, opened 2021-02-02
Restore "pause fix", extrude menu selected nozzle heat fix, polling fix

kisslorand:
### Description

Merging PR #1580 broke the pause fix made by PR #1588.
This PR restores pause fix, so Marlin parks the nozzle and stays there if `PARK_HEAD_ON_PAUSE` is defined or just simply pauses the print if `PARK_HEAD_ON_PAUSE` is not enabled.
Also fixes a bug in the Extrude menu. If the user selected the preheat option in the minimum extrusion temp warning popup, not the selected nozzle was preheated but the nozzle selected in the (Un)Load menu.

Includes [suggestions](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1598#issuecomment-771471930) by @digant73.

Includes [polling fix](https://github.com/luc-github/ESP3D-WEBUI/issues/158#issuecomment-771965307) by @luc-github.


digant73:
in that case, please in config.ini add

PARK_HEAD_ON_PAUSE  (in Configuration_adv.h)

just after

ADVANCED_PAUSE_FEATURE (in Configuration_adv.h)

kisslorand:
@digant73 
It's done.
There are more problems in the "language_hu.h" but I won't touch it (ex. "Load" is translated as a noun by the meaning of "the force of a weight", not as a verb). Mr. Antosz doesn't like someone else to edit/comment the "language_hu.h" file.

digant73:
@kisslorand For hungarian translation I cannot help you for sure :-). Not a problem the comments were translated by a mistake.
Just a possible cleanup. the function setHotendMinExtTemp() is defined in menu LoadUnload.c and it is then used also by Extrude.c. Probably it's better to move it on the API Temperatue.c

kisslorand:
Not possible because parameter cannot be parsed in a dialog's OK action and you have to send what nozzle do you want to heat. Your observation just made me discover a bug. In the Extrude menu the nozzle selected in the (Un)Load menu was heated.
I am working on fixing it.

digant73:
does your PR fix 1528 and 1590? eventually, it should be it important to identify and fix that bug

kisslorand:
I don't know because I cannot reproduce any of those bugs.

bigtreetech:
@kisslorand Hello, Thank you very much for your help in the project. Do you have Facebook to communicate with? If it is not convenient to publish, you can send your contact information to our email “Tech@biqu3d.com”

PR #1602, opened 2021-02-02
fix labels in language_de.h

rebel1:
This fix the labels in german translation


digant73:
in STRING_FILE_LIST_MODE  you put an extra space char before the text translation

rebel1:
> in STRING_FILE_LIST_MODE you put an extra space char before the text translation

Thanks for reviewing

PR #1604, opened 2021-02-03
label_pl_recovery_en bad label

CyberMODE:
will be label_pl_recovery_en

### Requirements

label_pl_recovery_en is the correct label, or send a error when install.


kisslorand:
The problem is in "language_es.h". The file "language_es.ini" is generated upon compiling based on "language_es.h".

Issue #1497, opened 2021-01-09
[BUG] Filament runout Bug to resume


Bastlwastl84:
Marlin Mode: starting Print
Long touching TFT. 
Switched to Touchmode
Filament Runout happens
At infoscreen:
![20210109_203027](https://user-images.githubusercontent.com/69717507/104107128-6eb9d800-52ba-11eb-8d3a-88f9098a4a86.jpg)
Pressing play -button doesn't take effect. Short showing pause -button, then resume-button.
Cant resume.
Can't switch back to marlin mode.

Opening extruder Menu make the Tft freeze

Skr Mini E3 V2
Marlin Bugfix 2.0.8.x from Yesterday
TFT 35 E3 V3
Firmware downloaded yesterday
Config at next posting

Bastlwastl84:
[2021-01-08 skr mini v2 marlin + tft35e3v3.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/5791706/2021-01-08.skr.mini.v2.marlin.%2B.tft35e3v3.zip)

3dfixer:
I've been pulling my hair out on this for days. got a sensor on e0det plug, which is where it should go.
this works fine, doing a m119 sends back open or triggered depending if theres filament.
the marlin side of the tft see this and does a filament change, then resumes where it left off. this is all done in config.h in marlin firmware.
All seems ok then??
However, the fancy icon side of the tft is looking for a sensor connected to the tft itself. As there is not one, it will trigger a false runout all the time. Doing a m119 reads the sensor on e0det, not a sensor on the tft.
Perhaps BIGTREETECH will look at sorting this out???!!

also, whatever side you start a print from, its best to stay there. Strange things happen when you swap.

hope this helps people

kisslorand:
If you have the filament runout sensor connected to the motherboard, try flipping the `fil_runout_inverting:` parameter in `config.ini`. I cannot remember which one but one of the settings will give you false alarm.

kisslorand:
Please try with M75 before print starts. I think this was what happened to me also and made me send M75 before every print.

Bastlwastl84:
At beginning of the startcode once?

kisslorand:
Yes, that is enough.
You can also include it into the start gCode in config.ini.

gerthalberg:
[...]

Issue #1605, opened 2021-02-03
[FR] Information on the filament consumed in two-color printing


radek8:
Please complete the final information on filament consumption in two-color printing.
The values could be side by side or as a sum.

Cura generates:

M118 P0 filament_data L: [0.58, 0.29] m
M118 P0 filament_data W: [1.92794625, 0.9545545239257811] g
M118 P0 filament_data C: [1.15676775, 0.5727327143554687] Kc

the display shows:

![image](https://user-images.githubusercontent.com/46979052/106820118-1a1c3980-667b-11eb-97bc-af27b4354d70.png)

Thank you

kisslorand:
Can you test please (when you have time) what happens when you do not include filament data in your gCode? I am curious what will the TFT calculate.

radek8:
If I delete the M118 information from the file.
The display calculates this

![image](https://user-images.githubusercontent.com/46979052/106955153-1fd65580-6735-11eb-8473-289ac44f601f.png)

kisslorand:
Thank you. I will (try to) make the data displayed if filament info present to show the sum of filament data.

radek8:
Note that the githubu counts a dot as a comma.

M118 P0 filament_data L: [0.58, 0.29] m
M118 P0 filament_data W: [1.92794625, 0.9545545239257811] g
M118 P0 filament_data C: [1.15676775, 0.5727327143554687] Kc

![image](https://user-images.githubusercontent.com/46979052/106956442-e272c780-6736-11eb-9387-2104914897a7.png)

radek8:
Thank you

kisslorand:
Is see no difference...

radek8:
I am sorry
punctuation marks are fine.
The translator confused the dot with a comma

kisslorand:
OK, I've done it, now the sum of  each data is shown. The character ";" cannot be a separator, as it means a comment but other than that anything (except "." of course).
The funny thing is that the code is now shorter with 8 bytes without any RAM addition. :)
A PR is imminent, I have to update the Readme and sleep... Actually the other way, sleep and update the Readme.

kisslorand:
Done: PR #1607

radek8:
Thank you

oldman4U:
Hi radek8.

[...]

PR #1607, opened 2021-02-05
Filament info for multi-filament print, possible fix for Load/Unload hanging popup, fix TFT freeze

### LINKED ISSUE #1571: [BUG] (screen freezes on filament load)
So I know that there are previous posts about how M0,M1 doesn't work on TFT screen. I have updated my screen firmware  recently, but I still have issues with loading filament. this time issue is worst. It used to be that screen fell into a error loop. now it freezes altogether.

1. heat up the nozzle.
2. press load.
3. wait for extruder to purge.
4. and then pop up apears. "printer paused" " waiting for user to respond"
5.press ok

 then screen freezes 

### Hardware Variant
using a SKR 1.4 turbo
TFT 43 connected through both marlin and serial port

### LINKED ISSUE #1629: [BUG] TFT35 keeps freezing
### Description
When randomly pressing buttons within the touchscreen, after a few seconds, the entire screen will freeze.

When the freeze occurs, the button will be 'highlighted' as if its some kind of touch effect. The only recovery is to reset the screen.

This screen has been functioning properly for months - so this is rather new.

### Steps to reproduce

1. Update to latest firmare
2. Randomly go through menus
3. Observe the freeze

### Hardware Variant
TFT35

TFT App: 5143b8162a927b303f865a149ca0363a47c2e347
Marlin: bugfix-2.0.x
M/B: SKR Mini E3 v2.0

In troubleshooting, I have:
* Reflashed firmware, icons and config.ini
* Reset TFT35 to defaults

kisslorand:
### Requirements

BTT TFT or MKS TFT

### Description

This PR adds support for multi-filament prints. It enhances the print summary so it shows the sum of each  filament data present in the gCode. (sum of length, sum of weight, sum of cost)

Later on added a possible fix for the Load/Unload popup hanging issue.
...and more later added a fix for TFT freeze.

### Benefits

It benefits those who use the filament data embedded into the gCode and use multi-filament print.

### Related Issues

Implements FR #1605
Probably fixes #1571
Fixes #1629

### Additional info

This PR doesn't bring any new popup, doesn't make the additions user configurable, it doesn't remove any existing features nor buttons. It doesn't change any other procedure than the filament data acquisition from gCode.

Extra RAM usage: ZERO (tested only on BTT TFT3.5 v3.0)
Extra Flash usage: ~~-8 bytes (8 bytes smaller compiled file) (tested only on BTT TFT3.5 v3.0)~~ 8 bytes


digant73:
Just as a further investigation. Some host actions are currently displayed twice. E.G. if you press on "purge more" button the same dialog is displayed immediately and again when the action is completed by marlin. Is it possible the response needs to be modified in some parameters just to avoid this annoying dialog repetition?

kisslorand:
Upon investigating the popup hanging, I saw (by monitoring the serial on both sides, TFT and motherboard) that the popup appears only when Marlin asks for it. I think I went over it at least 100 times

digant73:
Yes I saw the same. But probably for purge more, a different answer must be provided to avoid the immediate reply from marlin. When I have some time I will do some tests

kisslorand:
For "Purge more" the answer should be `M876 S0', for "Continue" the answer should be `M876 S1`.
Check what Octoprint does, it has no problem sending the same.

digant73:
I had a look to marlin source code. The double prompt is always provided when pressing on purge more button. So the handling on TFT is correct. It is a stupid implementation on marln side and is related to "filament runout" reason that you see with octopprint immediately after the prompt action info

oldman4U:
@kisslorand 

This would fix an ever existing bug, so this would be great!

Do you see, why this only fails on STM32 mainboards buts not on LPC based?

I will check asap and let you know.

Thank you

oldman4U:
Does not fix the problem with STM32 boards;-(

digant73:
it seems to me that this PR also fixes #1497.
I had also the freze of the display when starting a print with no filament. The runout error message has been shown but then I got a freeze (not possible to click on the continue button to restore the print)

kisslorand:
I have to add a little clarification.
The bug of Purge/Continue was something I never saw on my machine (SKR1.4T). Once it happened on my machine I knew I would be able to replicate it. So I found a way to make it hang whenever I wanted. That's how I was able to fix it. It seems I discovered another bug that had the same outcome, the Purge/Continue hang.
So this PR seems to fix just one of the bugs, or just another bug that leads to Purge/Continue popup to hang.

sarvenn:
@kisslorand hello,
Will it be merged soon to prebuild firmwares?

kisslorand:
Merging is not up to me, it's the BTT staff who is in charge around here.
In China it is the Spring Festival time so I guess it will take some time until they will get back to work.

bigtreetech:
> Does not fix the problem with STM32 boards;-(

It's confirmed that it's Marlin's problem not TFT,  I use printrun to send "M701 + M876 S0" through the computer and the motherboard also will reset. I guess it's the problem of Marlin STM32 Serial buffer overflow or something else.

oldman4U:
I know it is not the TFT but something else. We all didn't know at the time I wrote the above.

bigtreetech:
image
@oldman4U  It should be the problem of Serial buffer overflow. In Marlin, you can solve this problem by increasing the Serial buffer or increasing the priority of Serial interrupt

oldman4U:
Thank you

How can I increase the priority of serial interrupt?

BigTreeTech <notifications@github.com> schrieb am Fr. 26. Feb. 2021 um
06:40:

oldman4U:
The issue is likely fixed

https://github.com/MarlinFirmware/Marlin/pull/21210


Georg Deisting <georgdeisting@gmail.com> schrieb am Fr. 26. Feb. 2021 um
07:00:

PR #1609, opened 2021-02-06
Adding the ability to turn StealthChop on and off (M569)

radek8:
Adding the ability to turn StealthChop on and off (M569)


rebel1 (approved):

Guilouz:
What do you think about enable/disable features on TFT screen ? To show or not show it, configurable in config.ini.

kisslorand:
What would be the benefit of it?

radek8:
I understand your questions.
TMC drivers have two operating modes, stealthChop and spreadCycle.
This PR allows you to switch the limits of these modes from the touch screen menu without having to switch to Mmarli mode.

radek8:
The Config.ini file is used to configure the TFT display parameters, not to configure Marlin.

Guilouz:
> The Config.ini file is used to configure the TFT display parameters, not to configure Marlin.

Yes I know, but some features disabled or not used in Marlin like Linear Advance, Jerk (because use junction deviation), firmware retraction etc... still displays on TFT (with value 0 and can't be edited if are not enabled in Marlin). I think it's useful to choose which parameters you want or not to display on the TFT. Show which ones are actually editable.

radek8:
I'm not bothered by the options in the settings I don't have configured. I just don't click on them.
If someone clicks on the settings of a function that is not active, the printer returns a message - unknown command ...
Nothing serious will happen.
I'm afraid that disabling the standing settings options in Config.ini would be too much of an intervention in the code that no one will want to access.

Guilouz:
> I'm not bothered by the options in the settings I don't have configured. I just don't click on them.
> If someone clicks on the settings of a function that is not active, the printer returns a message - unknown command ...
> Nothing serious will happen.
> I'm afraid that disabling the standing settings options in Config.ini would be too much of an intervention in the code that no one will want to access.

Me neither but I configure machines for people and they do not necessarily understand for certain parameters appear when the machine cannot manage it (like Linear Advance, Z Align etc). Especially since one of the last PR added all the parameters

That's why I made this request, in the sense that only show the settings that are really modifiable.

radek8:
The PR you are talking about has added options from the ini file to the display parameters menu so that they can be edited from the display.
You are talking about options that set Marlin parameters and you want to optionally show and hide them.
I'm not a programmer and I don't want to express myself, but I can imagine that it won't be easy.
This would be a question about Bigtreetech, Guruathwal, Digant73, Kisslorand, Mehmetsutas ....
Kisslorand has already written his opinion :-)

radek8:
Are you making printers for your customers, and are you afraid that their own stupidity will change the parameters and they will want you to repair it?
If you have the correct default values for Marlin in Marlin and in the Config file. ini default values for TFT, then the remedy is very simple.

Guilouz:
I mean to add possibility to hide some parameters in Machine settings screen. Parameters which are not used by the Machine used.

In exemple on a CR-10s Pro in Machine settings, I see the possibility to enable/disable Z Align but with this motherboard this function is not possible, same for Linear Advance, I see it, can enter a K-value. Which doesn't make sense to me because these functions are not enabled in Marlin.

Another exemple, in your PR you add StealthChop settings, I disable it in my Marlin firmwares but I would still see this parameter which will be useless on the machines I use.

It would be better to activate the parameters to display that one wishes or even better to activate these functions only when possible by Marlin.
Hope you understand what i mean.

radek8:
So open a new ticket with the Feature request

rebel1:
Why do you closed this pr ?

radek8:
The parameter setting works great, but if I merge the PR and load it into the TFT, the display will freeze when I start printing and executing the G28 command.
I do not understand why. Printing is not affected by the adjustment. Execution of the G28 Command from the display will not freeze.
Because other users object to adding features they don't use, I won't address this further.

radek8:
Fixed.
Now PR is fully functional and ready to merge.

PR #1672, opened 2021-03-06
G26 test leveling with babystep menu

labotecno:
### Requirements
Add  G26 Gcod test leveling

### Description
i think G26 gcod is very important to testing if the bed is leveled or not , or so i put on the abl menu one icon for get it

### Benefits
more funcionality

### Related Issues
i don t  know yet , get it long time , i never get issue


labotecno:
i will make it good , i will to restart code

Issue #1670, opened 2021-03-05
[FR] Option to drop to a safe Z-height after Homing during bed leveling


lightmaster:
**Is your feature request related to a problem? Please describe.**
Unable bed level from touch screen when using a Delta printer, since it tries to move along X/Y before dropping down to where X/Y can move freely. This causes the belts to grind against the stepper gear repeated and afterwards the printer thinks that it is at a location that its not actually at.

**Describe the solution you'd like**
There needs to be an optional z-height to drop by before attempting to move laterally to a X/Y location.

**Describe alternatives you've considered**

**Additional context**

radek8:
Do you also have this problem in Marlin emulated mode?

lightmaster:
Tried to test that and stumbled upon #1657 bug. Just found that report and how to get the screen working again, so I'm gonna test it in a minute.

As a temp fix, I changed this line from https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/c367983689afc3cd4fe1f7b230f8a02f32b6b47d/TFT/src/User/Menu/MeshEditor.c#L893

to:
```c
mustStoreCmd("G28\nG1 Z150 F1500\n");                         // only the first time, home the printer
```

Not an eloquent solution by a long shot, but it does work to keep it from messing up the belts at least.

lightmaster:
In Marlin Mode, UBL Mesh Edit works great. It drops maybe 30 in height so it gets down to a safe area, then moves out to the point you selected

radek8:
Maybe @digant73 could help.
But I don't think it currently has a working display and I don't know if it has a Delta printer available to simulate the problem.

lightmaster:
> But I don't think it currently has a working display

Don't know if what has a working display? All of the display screens dealing with UBL seem to be working when I compiled the latest from the repo.

oldman4U:
digant has a broken display and is waiting for a replacement.

radek8:
digant73 is a programmer who created the Mash Editor

digant73:
As far as I understand the problem is when you move to mesh tuner and the nozzle is moved on XY coord.. The tuner uses G42. It is strange that the problem is not avoided by marlin. No I don't have a delta printer but I understood the problem.If an indication of delta printer is returned by marlin, the fix is easy. Otherwise a new configuration param is needed

lightmaster:
[...]

PR #1677, opened 2021-03-06
Feature delta drop before mesh edit

lightmaster:
- Add `IS_DELTA` to configuration.h
- Prevent `probeHeightStop()` on Delta printers
- Drop Delta printers by 50mm after homing during a Mesh Edit. This amount **should** be enough on most Delta printers to prevent a tower from crashing if the first point used is directly under a tower.

### Requirements

### Description
This PR adds a Configuration.h setting for whether or not a printer is a Delta printer. If set to true, then when starting a mesh edit, after homing, the hotend will drop 50mm below the homing point to allow the printer to have free motion to move along X/Y without crashing into the top of the towers. This addresses the issue I brought forward in #1670.

[Video of the hotend crashing prior to this PR](https://drive.google.com/file/d/1ytqReGYUiGMMHlO3ekQIlyB_EiudZByH/view?usp=sharing)
[Video of the hotend successfully homing and moving between points after this PR](https://drive.google.com/file/d/1yvXUsVR_KRnNp51VW3JrHbhV5kFRbQRg/view?usp=sharing)

This is my first time coding in this language, so I'd appreciate if someone could make sure this PR doesn't have any unwanted side effects.

### Benefits

Allows a Delta printer to safely home during a Mesh edit, without effecting other printers. Setting defaults to `false` so as to not effect anyone unless they specifically set it to true.

### Related Issues

#1670

### DIFF
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -195,7 +195,7 @@
 #define Z_MAX_POS 250
 
 // Is this a Delta printer
-#define IS_DELTA true
+#define IS_DELTA false
 
 // Pause Settings
 #define NOZZLE_PAUSE_RETRACT_LENGTH               15  // (mm)
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -195,7 +195,8 @@
 #define Z_MAX_POS 250
 
 // Is this a Delta printer
-#define IS_DELTA false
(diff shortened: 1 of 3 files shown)


lightmaster:
Didn't notice before that the homing button within the Mesh Edit page uses its own code and doesn't reuse the same code as the initial homing when you try to move to the first mesh point. Added the same drop down in that code too.

Issue #1676, opened 2021-03-06
[BUG] Popup error.


radek8:
Pop-up error when printing was not completed.
If my primary language is set to English, a popup window will appear normally.
If the secondary language is set, both the window name and the dialog buttons have the wrong description.

EN
![image](https://user-images.githubusercontent.com/46979052/110221745-60f58d00-7ece-11eb-9ef4-b72954a94943.png)

PL
![image](https://user-images.githubusercontent.com/46979052/110221757-6fdc3f80-7ece-11eb-8302-a567a7d8a5b6.png)

CZ
![image](https://user-images.githubusercontent.com/46979052/110221844-be89d980-7ece-11eb-9701-daaf2e6f035b.png)

kisslorand:
Can you check by using second language also English (or any other)?

radek8:
Secondary English

![image](https://user-images.githubusercontent.com/46979052/110222282-5a1c4980-7ed1-11eb-913b-788bfea0b08f.png)

kisslorand:
Thx for the test, the result makes it a general bug.

Issue #1657, opened 2021-03-02
[BUG] Marlin does not boot if TFT is in LCD Emulator Mode


DaStivi:
### Description

If TFT is set to Marlin Mode (LCD Emulator/Simulator) and printer is reseted or power cycled, marlin does not boot, as soon as TFT is switched to touch mode marlin does boot up correctly!

### Steps to reproduce

1. switch do Marlin-LCD Emulator Mode
2. Reset/Power-Cycle

**Expected behavior**
Board/Marlin should boot up normal and just connect in Marlin/LCD-Emulator Mode (subsequently OctoPrint is able to connect)

**Actual behavior**
Board does not boot

### Hardware Variant
B1-TFT35 
### TFT Firmware Version & Main Board Firmware details

Marlin: 2.0.7.2
TFT: prebuilt from 24.2.2021 (d3453505ab3fae7e758d3adbbb402e78b8b66a12)

gordo3di:
Try turning the knob after it boots. My screen is dark unless I turn the knob so it looks like it didn't boot.

DaStivi:
> Try turning the knob after it boots. My screen is dark unless I turn the knob so it looks like it didn't boot.

nope, thanks for the hint.. but also don't work or change anything on this issue....
i've made ah short video to show you... but nothing special too see marlin just not initializing when TFT boots in LCD-Mode...

https://1drv.ms/v/s!AiIZqPOtqD4XqNBYvE9c6s-JwWAtrg?e=Gn3zkB

sarvenn:
I have the same issue here. When I switch to marlin mode and then shut down and power on my custom bootscreen stays on the screen. When I push and wait the encoder mode switch page does not appear. I tried to click and turn the knob from my memory to do a homing. It does. The only way to switch to tft mode is to copy my config file to an sd card and reboot my printer. Then it opens wirh tft mode.
![20210228_204532](https://user-images.githubusercontent.com/66178082/109841401-110e9000-7c5a-11eb-8541-01cc1a2588fa.jpg)

sarvenn:
I have tft35 e3 v2 26 feb firmware. Now I will install latest firmware 3 March.

sarvenn:
I also did a test with my stock ender 3 display. Everything is perfect. So I'm sure that it is 26 feb firmware related.
[...]

PR #1655, opened 2021-02-26
Clean-up and reduce RAM usage, rewrite Terminal Menu, display machine compatible parameters only

guruathwal:
- Clean-up and re-write the Terminal menu to remove unnecessary RAM usage.
- RAM for the terminal cache is allocated only when the Terminal menu is opened. The terminal menu will display only one page while printing and 20 pages (max) when idle.
- Remove global variables from the custom G-code menu.
- Remove global variables from the Parameter Menu.
- Remove global variables from the Move Menu.
- Remove global variables from the Print Menu.
- Parameter menu will now display only the parameters reported by the machine in the `M503` response.
- Fix long title display in Numpad.
- Fix wrong Extruder minimum and maximum values in Numpad.
- Add Material theme option for keyboard and Numpad in the configuration.h file.

Build Report:
Device | Old RAM Usage | New RAM usage | RAM Reduced | Old Flash Size | New Flash Size | Flash Size Changed
-- | -- | -- | -- | -- | -- | --
TFT35 V1.0 | 20620 | 17660 | -2960 | 143172 | 143460 | 288
TFT35 V1.1 | 20620 | 17660 | -2960 | 143172 | 143460 | 288
TFT35 V1.2 | 20620 | 17660 | -2960 | 143052 | 143340 | 288
TFT35 V2 | 21388 | 18428 | -2960 | 152632 | 152920 | 288
TFT35 V3.0 | 21824 | 18860 | -2964 | 171036 | 171332 | 296
TFT35 E3 V3.0 | 21848 | 18884 | -2964 | 171700 | 171980 | 280
TFT35 B1 V3.0 | 21848 | 18884 | -2964 | 171708 | 171988 | 280
TFT43 V3.0 | 21848 | 18884 | -2964 | 171692 | 171988 | 296
TFT50 V3.0 | 21848 | 18884 | -2964 | 171692 | 171988 | 296
TFT70 V3.0 | 21976 | 19012 | -2964 | 172752 | 173116 | 364
TFT28 V1.0 | 20620 | 17660 | -2960 | 143196 | 143484 | 288
TFT28 V3.0 | 21848 | 18884 | -2964 | 171512 | 171800 | 288
TFT24 V1.1 | 21772 | 18808 | -2964 | 166988 | 167276 | 288
MKS TFT32 V1.4 | 22536 | 18772 | -3764 | 166864 | 167160 | 296
MKS TFT28 V1.0 | 22536 | 18772 | -3764 | 166864 | 167160 | 296

[...]

PR #1562, opened 2021-01-25
update framework

Msq001:
### Requirements

* Filling out this template is required. Pull Requests without a clear description may be closed at the maintainers' discretion.

### Description
* modify framework from `stm32cube` to `cmsis`
* update current mcu type, sram size, mcu document url

### Benefits
* STM32F105RC / STM32F207RC / STM32F407VG has large SRAM size
* upload and debug mcu by upload tools, such as cmsis-dap, st-link, jlink, etc...(all the tools supported by openocd)

PR #1680, opened 2021-03-07
Bugfix: boot issue in marlin mode and prevent mode switching while printing, wrong keywords in Popup dialog

guruathwal:
- Fix freeze/hang/boot issue if booting in marlin mode with serial_alway_on active. (#1657)
- Fix mode switching issue if printing from onboard SD with Marlin Mode active due to changing menu index.
- Fix wrong keywords in Popup dialog. (#1676)

Resolves: #1657, Resolves: #1676


radek8:
Problem #1676 is fixed.
The description of the buttons is already displayed correctly, but after clicking the cancel button, the display freezes.

guruathwal:
@Msq001 There seems to be a problem with `Printing.sys` file after PR #1562 (I checked on TFT43, TFT35v2 and TFT70). @radek8 can you check with #1655 to confirm?

radek8:
When using # 1655 there is a bad description on the confirmation buttons.
When the button is pressed, the display responds correctly

Issue #1673, opened 2021-03-06
[BUG] Printing from OctoPrint does not show up in touch mode


lightmaster:
### Description

When I initiate a print from OctoPrint, the touch mode does not show any printing information and doesn't seem to notice that a print has been initiated. I can still control babystepping from the touch screen and I can see the hotend and bed temps during a print, so clearly there is still communication between the screen and board.

### Steps to reproduce

1. Connect OctoPrint to main board
2. Initiate print over USB serial
3. Observe that the TFT does not show a printing screen

**Expected behavior**
When a print is started from any source, including TFT SD card, main board MicroSD card, and main board USB, the TFT touch screen should show the printing screen with various bits of information about the ongoing print.

**Actual behavior**
TFT touch screen fails to detect that a print has been initiated.

### Hardware Variant
Using TFT43 and SKR 1.4 Turbo
### TFT Firmware Version & Main Board Firmware details
TFT Firmware compiled from `master`, branch on 2021-03-05
SKR 1.4 Turbo firmware compiled from `bugfix-2.0.x` branch on 2021-03-05

### Additional Information
[TFT43 Configs 2021-03-05.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/6094760/TFT43.Configs.2021-03-05.zip)
[SKR 1.4 Turbo Marlin Configs 2021-03-05.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/6094762/SKR.1.4.Turbo.Marlin.Configs.2021-03-05.zip)

Foxcor:
This is not a screen or firmware issue.
Install additional plugins in Octoprint.
https://github.com/OllisGit/OctoPrint-DisplayLayerProgress for example

oldman4U:
@Foxcor

Would you be able to provide some basic information regarding the things necessary to make Octoprint work with the TFT? I would add them to the ReadMe then. Thank you

lightmaster:
Ok, that makes sense then since the print I was trying was a bed leveling test print and has no layer information, therefor DisplayLayerHeight doesn't do anything for this print. iirc, the MKS Robin Nano I was using before would still show the print screen regardless of whether or not DisplayLayerHeight was going. Could be wrong though.

lightmaster:
[...]

PR #1681, opened 2021-03-07
Bugfix for PR #1680

kisslorand:
### Description

PR #1680 introduced a bug where if printing from SD remotely (Octoprint. Pronterface, etc) the printing menu would not activate, instead the TFT would stay at the menu where it was at the time the print started. 
This PR activates the printing menu and makes it next in line after the MainScreen. That means that after the print is finished and the user presses "Back" it will bring him to the MainScreen.

### Benefits

Everyone.

### Related Issues

No issue reported as it was discovered early in time.


DaStivi:
@kisslorand is there ah reason you reverted #1680 back? i've justed tested this, and now neither if serial_always_on is enabled or disabled, i can't get into marlin mode anymore now?!

DaStivi:
sorry for might generate some confusement... @guruathwal do i see that correctly you also allready reverted the #1691 bugfix of bugfix1680?  I don't see through it anymore 🙈 sorry... but anyhow i'am on the latest build now, and marlin mode doesn't work anymore, no matter if serial_always_on is enabled or disabled...

guruathwal:
@DaStivi this is not related to #1673

DaStivi:
> @DaStivi this is not related to #1673.
> P.S. I tested the latest commit and Marlin mode is booting with LPC based mainboard with TFT set to marlin mode with and without serial_alway_on active.

yeah, i have gone back to your initial change to fix it... its still not working.. so i don't know what happend, i gonna reset.txt the tft now... maybe it was ah issue while i gona way back to the bug version to test the octoprint-print screen i did in 1673  just had the 1681 now in mind as this reverted it... but when i have got everything correct you again reverted 1681 so it should be fixed again right now in latest builds, isn't it?

guruathwal:
you need to ignore this PR(#1680) and check PR #1686.

DaStivi:
> > @DaStivi this is not related to #1673.
> > P.S. I tested the latest commit and Marlin mode is booting with LPC based mainboard with TFT set to marlin mode with and without serial_alway_on active.
> 
> yeah, i have gone back to your initial change to fix it... its still not working.. so i don't know what happend, i gonna reset.txt the tft now... maybe it was ah issue while i gona way back to the bug version to test the octoprint-print screen i did in 1673 just had the 1681 now in mind as this reverted it... but when i have got everything correct you again reverted 1681 so it should be fixed again right now in latest builds, isn't it?

ok, it was my fault, sorry for the double confustion... marlin mode didn't worked because i used an old config.ini file where the new #### Marlin Mode Type
# Select Marlin Mode type.
#   Options: [128x64 Full Graphic LCD: 0, 20x4 Character LCD: 1]
marlin_type:0

options was not init... @guruathwal might be a good idea btw. to change the default behavior to the 128x64 mode, even without having it in config.ini to prevent DAUs like me :)

PR #1682, opened 2021-03-07
Revert "BTT TFT70 display icon repair (#1669)"

Acenotass:
As an author, I am very indignant why they changed the style of the Tre Round Maracle theme! Why did you replace the long-made adjustments for the TFT70 in my theme from another theme?


Acenotass:
misstake

oldman4U:
Nobody is allowed to touch your theme!

😁

AntoszHUN:
Only the incorrectly small logo.bmp file was replaced with the normal size. This is a problem? I did not know. Sorry.

PR #1526, opened 2021-01-17
This PR is fixing the bugs introduced by merging PR #1487

kisslorand:
### Requirements

MKS or BTT TFT

### Description

This is a bugfix for the bugs introduced by PR #1487.
The fixed bugs are:

- filament length not calculated when printing from remote host and from onboard SD
NoLength

- in some cases, prinintg from remote host and onboard SD, the file name is missing from the print summary
NoName

- other languages than English are not taken into account properly after print finished
PrintFinishLanguage

- in some cases, printing from remote host from onboard SD, after print finished cannot exit from the menu, screen freezes

https://user-images.githubusercontent.com/50251547/104843268-62dd9f80-58d2-11eb-8b01-749f4a45a2fb.mp4

- PR #1487 wrongly assumed that "File opened:" is received from Marlin only if printing from remote host and onboard SD, hence even if TFT initiated the print from onboard SD, TFT was informed that it's a remote that started the print (`infoFile.source == BOARD_SD_REMOTE`). This PR makes the TFT aware again if print was started from remote host or not.
- if the user embedded some filament data into the gCode and didn't wanted to have the filament length, its was still initiated and calculated (only if printed from TFT and onboard SD because of some before mentioned bug)
- duplicate code removed

### Benefits

Everyone

### Additional info

This PR does not include any extra popup, there's not user configurable parameter added.
Extra RAM usage: 0 (zero)
Extra flash usage: ~72 bytes

PR #1686, opened 2021-03-07
Bugfix: prevent switching to touch mode while printing from Marlin mode, Revert #1681 and remove Printing menu handling from ParseACK.c

guruathwal:
- Reverts commit #1681 and remove printing menu handling from `ParseACK.c`.  
  - Printing menu handling for Onboard SD printing is already handled via `loopCheckPrinting()` in `Printing.c`. 
  - In PR #1526 a duplicate printing menu handling was added in `ParseACK.c` which introduces a bug that causes unintentional switching to the Touch mode without proper hardware initializations when a print is started in Marlin mode, resulting in the buzzer being disabled and a non-functional stop button.

### DIFF
--- a/TFT/src/User/API/Printing.c
+++ b/TFT/src/User/API/Printing.c
@@ -493,7 +493,8 @@ void loopCheckPrinting(void)
     infoPrinting.printing = true;
     if (!hasPrintingMenu())
     {
-      infoMenu.menu[++infoMenu.cur] = menuPrinting;
+      infoMenu.cur = 1;
+      infoMenu.menu[infoMenu.cur] = menuPrinting;
     }
   }
 
--- a/TFT/src/User/API/parseACK.c
+++ b/TFT/src/User/API/parseACK.c
@@ -503,8 +503,6 @@ void parseACK(void)
 
         infoFile.source = BOARD_SD_REMOTE;
         initPrintSummary();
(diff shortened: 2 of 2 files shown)


kisslorand:
Great catch!

digant73:
@guruathwal, kisslorand
missing "break;" after line 120 in Buzzer.c (for case sound_deny:).
Desired behavior or bug? In my PR I added the "break;"

kisslorand:
I do not know if it is a bug or not, but the lack of that "break" makes the `sound_deny` to sound nice. Without "break" it also triggers the `default:` case and you'll have also the `Buzzer_TurnOn(BUZZER_FREQUENCY_HZ, BUZZER_FREQUENCY_DURATION_MS)` executed.
With "break" it sounds boring.

Now, I would put that "break" but also the second `Buzzer_TurnOn`. Something like this:
`case sound_deny:`
`___if (infoSettings.alertSound != 1)`
`______break;`
`___Buzzer_TurnOn(500, 10);`
`___Buzzer_TurnOn(1000, 20);`
`___break;`

guruathwal:
@digant73 there should be a break;

digant73:
yes, I already added the break and then followed the indication from kisslorand to a have a better sound

sarvenn:
Does it matter if I enable the speaker in my marlin config?

PR #1690, opened 2021-03-07
Addition of parameter M200 - Set Filament Diameter

radek8:
Added the ability to turn on volumetric  extrusions and filament diameter settings


digant73:
@radek too late unfortunately (as it happened also to me). The parameters should be placed (in all files) ordered by their gcode while instead M200 was put after M900. The same happened for M569 (it is after M913)

radek8:
> @radek too late unfortunately (as it happened also to me). The parameters should be placed (in all files) ordered by their gcode while instead M200 was put after M900. The same happened for M569 (it is after M913)

I am sorry.
If you mean the sequence of commands in parseACK.c and interfaceCmd.c, I can fix it

digant73:
everywhere, so also in machineParameter etc... Noh, no problem. Just a suggestion to follow for your next PR

radek8:
In Parametersetting.c I have it arranged intentionally.
CURRENT_SETTING, BUMP_SENSITIVITY, HYBRID_THRESHOLD, STEALTH_CHOP
I wanted to leave the TMC driver items together.

I placed the fiber diameter setting behind the linear edvance, because these are the parameters that relate to the filament extrusion.

digant73:
I know the logic you used. I made the same with offset (home, probe, MBL) and guruatwhal explained to me the logic he originally followed.

radek8:
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1697

radek8:
Guruatwhal is preparing for a bigger change, so let us be surprised. :-)

PR #1691, opened 2021-03-07
Followup to #1677 Delta drop before mesh edit

lightmaster:
DELTA_MBL_Z_DROP_MM *must* be definied as a float instead of an int or double if it's going to be called as a float.

### Requirements

### Description

Defining `DELTA_MBL_Z_DROP_MM as an int prevents it from being called since it's being called as a float.

### Benefits

Allows #1677 to function as expected.

### DIFF
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -196,7 +196,7 @@
 
 // Is this a Delta printer
 #define IS_DELTA             false
-#define DELTA_MBL_Z_DROP_MM  50 // MBL Drop 50mm first after home avoid crashing into the top of the towers.
+#define DELTA_MBL_Z_DROP_MM  50.0f // MBL Drop 50mm first after home avoid crashing into the top of the towers.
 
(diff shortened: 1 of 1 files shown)


digant73:
@lightmaster based on your PR I suspect you should have the same problem also in the Probe Offset menu. Did you try it?

lightmaster:
I'll try it and see. Already knew my probe's offset, so never messed with that menu.

digant73:
let me know so I will eventually fix it. It could be present also with MBL process

lightmaster:
Initial homing is fine, however when you click the button in the bottom left a second time (click once to home and drop down, click second time after setting probe offset) or click the back button it homes and then tries to rise up again. As a rule of thumb, a Delta should NEVER try to rise up higher after a homing.

digant73:
it homes when you click on the bottom left button and when you exit the menu.
Ok, as expected the problem is present. I will fix it in a next PR.
Thanks for the testing!

PR #1694, opened 2021-03-08
Bugfix: Wrong character width in Terminal

guruathwal:
- Fix: Wrong character width in the Terminal when the first character received has a larger width (originated from #1171). The character info was being fetched only for the first character on a page, this resulted in all the character on the page being rendered with spacing based on the first character's width.

### DIFF
--- a/TFT/src/User/Menu/SendGcode.c
+++ b/TFT/src/User/Menu/SendGcode.c
@@ -748,10 +748,10 @@ void menuTerminalWindow(void)
       terminalDrawPageNumber();
     }
 
-    getCharacterInfo((uint8_t *)(terminalBuf + lastTerminalIndex), &info);
-
     while (terminalBuf + lastTerminalIndex && (lastTerminalIndex != terminalData->terminalBufTail))
     {
+      getCharacterInfo((uint8_t *)(terminalBuf + lastTerminalIndex), &info);
+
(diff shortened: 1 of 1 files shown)


digant73:
not related to the bug, but I would also consider a further improvement consisting of using different colors for the requests and responses in the terminal. E.g. I would configure a second color just slighly less/more opcaque/brighter than the other one. That should make the readability even better than now. It could be tested first with an hard coded color

guruathwal:
Did you plant any spyware on my PC? I am actually working on this.
image

oldman4U:
We get a live coverage of your screen here in Europe. Free TV.

You didn’t know this!!??

guruathwal:
Ohh crap. I need to realign my satellite again.

digant73:
> Did you plant any spyware on my PC? I am actually working on this.
> image

I suspect the same in my home :-).
What about to test also orange and MAT orange (if it is present). the contrast should be less drastic and nice to see. In general I was thinking to use the MAT version for one of the two colors. Although I suppose both colors will be made configurable in config.ini

guruathwal:
There are two alternative to setting color through config.ini
1. option to select from preset color combo. (1byte RAM)
2. option to select any color for both.(4bytes RAM)

PR #1696, opened 2021-03-08
Update italian language

BsCmOD:
Update italian language


radek8:
Also add an ini file to PR.

BsCmOD:
Off course but I don't know how to do it

radek8:
If you are using VS Code, just tap on the sprout in the crew you made the adjustment.
Close VS Cde and it will be generated automatically when restarted. Then just publish.

radek8:
added to
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1698

Issue #1158, opened 2020-10-06
[BUG] Gcode Filename letter followed by number shows ???? when printing from Onboard SD


blueeagle69:
Hi.

I've noticed a weird issue when browsing GCode on my onboard SD card.
I think it's fine when files are on TFT media.
On some files, the names show as "????"
![20201006_234538](https://user-images.githubusercontent.com/9963876/95268262-e0533780-082e-11eb-8f80-bb6640f8f008.jpg)

The file name shows as B1SENS~1.GCO when I select it.
The long filename is "B1 sensor_bracketWIP V3_1_0.2mm_PLA_29m.gcode".
The filename is a tad long, but the question marks are weird.

I do wish the TFT would support long filenames when printing from onboard media.

zonalimitatore:
maybe dont like .GCO
Try rename to .GCODE

blueeagle69:
It has a .GCODE extensuin, but the TFT shortens filenames when printing from the Onboard SD  😒 All the rest are fine, but I always seem to get one that shows "????"

zonalimitatore:
Try to rename it,more short and without any point or symbol

blueeagle69:
Was about to try the same thing. Will  let you know

blueeagle69:
Nope I shortened the filename right down.
I also removed all underscores although the rest have them.

It still shows as "????"

zonalimitatore:
there is again 0.2 in filename?

blueeagle69:
Okay then.

I shortened the filename to B1 Sensor Bracket.gcode.
Same thing.
I then removed the B1 and left it as Sensor Bracket.
That worked.
I then put it back minus the space after B1, it came back as ???? on the TFT.
I then put the B back and left out the "1". It worked.
So for some reason the TFT doesn't like a number after the first character of the filename?

zonalimitatore:
Ahahahahahaha.....omg
One day...one bug

blueeagle69:
Confirmed. I created a dummy Gcode file and called it A3 Test.gcode.
It showed as ???? on the TFT.
I think I may have stumbled across a bug.
Only in TFT

zonalimitatore:
Ive name this without problem.....Now i try like you do

![Annotazione 2020-10-07 163936](https://user-images.githubusercontent.com/43097238/95346244-e6432a00-08bb-11eb-8ea4-71bb7df41d1f.jpg)

[...]

PR #1699, opened 2021-03-09
Bugfix: '???' instead of Long file name for Onboard SD

guruathwal:
- Fix: long file name not updated correctly causing '???' displayed in place of file name. (#1158)

resolves: #1158


digant73:
@guruathwal
there is a missing "break" for "case 218:  //M218 Hotend Offset" in interfaceCmd.c

PR #1706, opened 2021-03-10
Enable RRF Macros menu support

pfn:
### Requirements

* Add menu to display macros from RRF
* Execute macros
* Config file setting to show or hide menu
 
### Description

Enables a Macros menu when `rrf_macros_on: 1` is set in config.ini

Can browse and execute macros in the top-level. This support was rebased onto master from https://github.com/Thro42/BTT-SOVOL-TouchScreenFirmware-Duet. However it is more robust than the original version.

Limitations:
* macros cannot be interactive, no prompts or confirmations
* the TFT will block until the macro completes running (intentional)

### Benefits

This is a feature enhancement to allow users on BTT TFT on RepRapFirmware to execute macros stored on the printer.

### Related Issues

#1400 this improves RRF support.

PR #1716, opened 2021-03-11
Fixed Mesh Editor + other fixes and cleanup

### LINKED ISSUE #1635: [BUG] Manual Bed Mesh Edit saves/displays the wrong position
### Description
I try to edit the bed mesh. But I cannot save my edits.

### Steps to reproduce

1. I leveled the bed by hand with the 4 screws and a paper. 
2. To compensate my extremely uneven, wrapped bed, I started the manual bed leveling (I don't have a BLTouch).
3. I did some test prints and now I want to finetune the mesh. But every time I change any value, in the _Mesh Bed Leveling_ menu and saves it to the EPROM, it gets lost in the 8Bit digital nirvana. 

You can follow the process in the video:
### 🎞 https://youtu.be/_9Il-2r0p7E

I got the hint, that the **I2 J2** seems be the wrong index. And maybe it displays the wrong coordinate.
So I tested:
1. Open  _Mesh Bed Leveling_ 
2. Change I2 J2 from `600` -> `540`
3. Check the value of **I2 J2** -> It's still `600`
4. Check the value of **I2 J1** -> Changed to `540`

> So it stores the value. But not to the position displayed in the mesh preview.
5. If I check the value with `M420 V`, the `540` is present on **I2 J2** (0 based index).

**Expected behavior**
After hitting the save icon, the value should be present in the mesh.

**Actual behavior**
The value resets every time to the previous value.

### Hardware Variant
Artillery Sidewinder X1 4 default TFT. STM32F 107
**TFT**: `0.26.1_patch_9` **Board**: `2.0.7.2_mbl`

digant73:
* fixed wrong mesh grid visulaization for MBL (#1706) in  Mesh Editor (e.g. a 5x5 grid was displayed as 5x4)
* fixed missing labels for many languages for last merged PR  #1706
* reviewed string parsing order in parseAck.c to improve general performance
* Added new param "z_raise_probing" on config.ini to support also Delta printers on menus Probe Offset / Mesh Editor for 
  the Z offset / Z height tuning process. It improves PR #1670
  **NOTE:** It MUST be negative (e.g. -50mm) for a Delta printer to avoid crashing into the top of the tower.
* some cleanup

fixes #1635
fixes #1706 (PR)
improves #1670 (PR)


TFT/src/User/API/config.c
```diff
@@ -811,6 +811,10 @@ void parseConfigKey(uint16_t index)
       infoSettings.xy_offset_probing = getOnOff();
       break;
 
+    case C_INDEX_Z_RAISE_PROBING:
+      SET_VALID_FLOAT_VALUE(infoSettings.z_raise_probing, MIN_Z_POS_LIMIT, MAX_SIZE_LIMIT);
```
lightmaster, line 815: Using `MIN_Z_POS_LIMIT` as the minimum will prevent Delta's from being able to set it to a negative value, since `MIN_Z_POS_LIMIT` is currently set to 0.

lightmaster (changes requested):
`MIN_Z_POS_LIMIT` is set to `0`, preventing Delta's from setting `z_raise_probing` to a negative value

TFT/src/User/API/config.c
digant73: thanks for the feedback. PR  #1718 fixes the XYZ min limit to -2000 to support also Delta printers

Issue #1400, opened 2020-12-16
[FR] RRF compatibility 


bttguy:
**Describe the solution you'd like**
Some (and a growing number) of us use the BTT TFT for basic monitoring and controlling their SKR LPC/STM port based RRF. It would be nice to be able to monitor the print status, modify some of the parameters during printing, accessing macros, and use some of the extra features if possible like LED changing, and probing...

**Describe alternatives you've considered**
There are some forks and forked forks that try to modify the base firmware to enable as much functionality as possible, but lack the constant updates, and work as is they were made.

**Additional context** 
The RRF project bases mainly on BTT boards, and there are already sollutions to make them work with BTT TFTs, so it would be nice to support it “officially”. I dont have to skill to do this and dont know how hard it would be to make and manage, but this TFT github community is quite nice as I followed during my Marlin use, so maybe some of you have the skill and time to modify and keep updating the Gcode and communication related differences in the firmware as an alternative for the users, so more of use could use this TFT in our printers. 

The project: https://github.com/gloomyandy/RepRapFirmware/wiki

Does this idea has any potential to be implemented in the future?
Thank you in advance.

radek8:
look at #1399

bttguy:
A saw it, thats why I try to push RRF compatibility even further. If you add that Gcode line, that only allow to communicate to the board at te TFT port. Would like to resolve if possible his/our issue "Due to the lack of \n when the Reprapfirmware responds to M20, it cannot read the onboard sd card normally.
Some advanced settings of marlin different from reprap are not available."
I am thinking of a paralell "official" branch that keep up with the changes and modify the code to be able to properly communicate with RRF.
This is a fork which is try to modify the FW for RRF: https://github.com/Thro42/BTT-SOVOL-TouchScreenFirmware-Duet

oldman4U:
Add @Subcircuit because he is the one who contributed related to RRF.

oldman4U:
The interest seems to be somehow limited. bttguy, are you willing to take over all the administration part and what will you do to attract developers to do the development? Or are you a developer and how can this community help?

Thro42:
If you like to bring in my expirients from https://github.com/Thro42/BTT-SOVOL-TouchScreenFirmware-Duet. For me it is better to start here from scratch. The fork from JimmymaDdness and bcmob is too far away from the actuallity one. I have to start over anyway.

bttguy:
[...]

PR #1718, opened 2021-03-11
Configurable emulated M600 / real M600 + bugfixes + cleanup

### LINKED ISSUE #1634: [BUG] The runout sensor doesn't work correctly
I have bough the btt skr mini E3 v2.0 whit the tft35-e3 v3.0 few weeks ago. When I start to configure them I have an issue with a basic runout sensor. This sensor is connected to the mainboard and when the sensor is triggered works well until you want to restart the print. When you want it, the screen in graphycal mode show a message that sais "The printing was stopped by user" with an "ok" button. I spect that when I press the button it will restart, but the message continuously are showing and the print doesn't restrart. In marlin mode, this isssue doesn't happend but if I bougth the tft is for use them whit the graphycal mode.

So, the question is, how I can restart the printing after this message? how I can fix that bug?

For more details I use the version of Marlin 2.0.7.2 that appears in the BTT github and the firmware of the TFT is the version 0.26.x of Feb 4 2021 of that model.

NOTE: If you come to the bug that happens with this sensor that even if it was triggered doesn't work as I sais (like happen in this bug report: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/1612) you must know that the solution is put as G-code start M75 in your slicer. This is another bug report for another bug.

### LINKED ISSUE #1719: [BUG] invalid entry in Config.ini
there is a problem with the rrf_macros_on entry in the current version

image

digant73:
* Added parameter "emulate_m600" on config.ini to select between emulated M600 / real M600 handling.
With emulated M600, the TFT intercepts the M600 gcode (filament change) and emulates the logic instead of demanding it to Marlin firmware.
Enable it, in case Marlin firmware does not properly support M600 on the mainboard
* New param emulate_m600 is also selectable at runtime from Features menu
* fixed missing "break;" for new configurable RRF Macro param
* fixed XYZ min limit to support also Delta printers
* fixed a bug on Marlin prompt action process execution (such as filament runout, filament change etc) not allowing to move on main menu after print completion or cancellation. Continous beeps and "filament runout" error message was displayed. Furthermore, the print menu looped on pause/resume instead of moving to main menu
* on print resume, the filament purge is now performed before the nozzle is moved on top of the print. So it avoids extra filament on top of the print also in case a filament change is done  (e.g. using the extrusion menu) on pause
* fixed a possible out of buffer error on function hostActionCommands() in parseAck.c
* minor fixes on some UI menus (e.g. scrambled °C on PID menu)

fixes #1719
fixes #1634

**PR status: READY TO BE MERGED**


Guilouz:
Good job !

TFT/src/User/API/config.c
```diff
@@ -659,6 +663,7 @@ void parseConfigKey(uint16_t index)
     //----------------------------RRF Mode Settings
     case C_INDEX_RRF_MACROS_ON:
       infoSettings.rrf_macros_enable = getOnOff();
+      break;
```
pfn, line 666: woops! good catch, thanks

bttguy:
Thank you for helping #1400 as well!!

TFT/src/User/API/config.h
```diff
@@ -173,12 +174,8 @@ extern "C" {
...
 #define NAME_MIN_LENGTH           3         // minimum name length
 #define GCODE_MIN_LENGTH          3         // gcode length less than this will not pe parsed.
-#if (IS_DELTA)
-  #define MIN_XY_POS_LIMIT        -2000     // Set a negative minimum position for Deltas
-#else
-  #define MIN_XY_POS_LIMIT        0         // position value less than this will not be parsed.
-#endif
-#define MIN_Z_POS_LIMIT           0
+#define MIN_XY_POS_LIMIT          -2000     // Set a negative minimum position for Delta printers
+#define MIN_Z_POS_LIMIT           -2000     // Set a negative minimum position for Delta printers
```
lightmaster, line 178: Shouldn't we have 2 separate minimum z limits? This variable is used as a sanity check for the amount that the nozzle should be raised by when pausing, which should never be negative. It's also used with this stuff, which I'm not exactly sure how that works, but seems like it shouldn't ever be negative either.

https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/bff5d814241e4f282bfb7fd14d7ba2a49687a348/TFT/src/User/API/config.c#L798-L808
&nbsp;
&nbsp;
Also, with `MIN_XY_POS_LIMIT`, for Cartesian printers this shouldn't be negative, so if the sanity check allows it to be negative for those printers, what's the point of the sanity check in the first place? Due to how different Deltas and Cartesian printers handle the X/Y values and where 0,0 is on both, I think it might be best to branch the sanity checks to have separate limits for these 2 different types of printers.

digant73: we can do it easy without too much constraints (compile directives or new conf. param to distinguish cartesian or delta). It could be simply put a note in config.ini reporting that negative values can be used for delta as I did in the new param "x_raise_probing" I provided

TFT/src/User/API/parseACK.c
```diff
@@ -1093,6 +1093,38 @@ void parseACK(void)
           ackPopupInfo(echomagic);
         }
       }
+
+      // keep here and parse at lastest
```
kisslorand: *latest

TFT/src/User/API/interfaceCmd.c
```diff
@@ -224,12 +225,12 @@ void sendQueueCmd(void)
...
-          if (isPrinting())
+          if (isPrinting() && infoMachineSettings.firmwareType != FW_REPRAPFW)  // Abort printing by "M0" in RepRapFirmware 
             setPrintPause(true, true);
           break;
 
         case 1:
-          if (isPrinting())
+          if (isPrinting() && infoMachineSettings.firmwareType != FW_REPRAPFW)
             setPrintPause(true, true);
           break;
```
kisslorand: Maybe:
`case 0:
case 1:
___if (isPrinting() && infoMachineSettings.firmwareType != FW_REPRAPFW)
______setPrintPause(true, true);
___break;`

kisslorand: Maybe:
`case 0:`
`case 1:`
`___if (isPrinting() && infoMachineSettings.firmwareType != FW_REPRAPFW)`
`______setPrintPause(true, true);`
`___break;`

lightmaster:
Didn't you said you were gonna fix the probe offset such that it doesn't crash the towers after homing? Think this is the line that's causing the trouble, specifically Line 46

https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/3a5c5a8198ce1cc435abf40314ee325030c86fd2/TFT/src/User/API/ProbeOffsetControl.c#L40-L49

digant73:
@lightmaster yes correct thanks, I forgot to replace the old parameter level_z_raise with the new one z_raise_probing as it is in function probeOffsetEnable()

TFT/src/User/config.ini
```diff
@@ -30,6 +30,7 @@
 # ensure that the following options are enabled in Marlin firmware.
 #
 # General options:
+#   G26_MESH_VALIDATION (in Configuration.h)
```
Guilouz, line 33: PR for G26 is closed. is it really useful to put it in the config.ini?

digant73: I think the author of the PR G26 is trying to reopen it without the problem he had.
However, yes it's better to remove the text from config.ini at the moment

lightmaster:
Jeez this has become a busy PR, lol.

I see you now added "ready to merge" to the description; next time, why not mark your PR as draft do it can't be merged until you're actually ready for it to be?

kisslorand:
I guess it was ready to be merged but newer and newer fixes were added. Instead of opening another PR he added it to this one.

lightmaster:
Yea, unrelated fixes and features should probably be done as separate PRs. He's gonna have fun if the repo owner only wants part of it merged.

kisslorand:
I wouldn't worry that much.

TFT/src/User/Configuration.h
```diff
@@ -299,11 +299,19 @@
 #define PID_PROCESS_TIMEOUT (15 * 60000)  // (MilliSeconds, 1 minute = 60000 MilliSeconds)
 
 /**
- * M600, M601 ; pause print
+ * M600 ; emulate M600
+ * The TFT intercepts the M600 gcode (filament change) and emulates the logic instead of demanding it to Marlin firmware.
+ *
+ * NOTE: Enable it, in case Marlin firmware does not properly support M600 on the mainboard.
+ */
+#define EMULATE_M600 true  // To enabled: true | To disabled: false (Default: true)
```
lightmaster, line 307: Shouldn't this be set to `false` by default, so that the introduction of this doesn't unexpectedly chance things for users?

lightmaster (changes requested):

TFT/src/User/Configuration.h
digant73: that was the default also in the past (when it was hard coded). Not all people have Marlin configured to support real M600.
For the status of the PR, don't worry BTT is awares of what some people is working on

sarvenn:
Hi.
After this PR, does it matter which one I use emulated or marlin M600?
Do they act the same now? If not, which one is better to choose then?
My Marlin M600 is fully configured properly. So as I understood from the description, it is better to use Marlin M600 if configured properly, am I right?

digant73:
@sarvenn Marlin M600 is more complete. So if you have Marlin properly configured, you can use it.
This means "emulate_m600" set to 0 in config.ini file.

Jhyrachy:
Hello

I compiled the firmware and installed it, but M600 does not work in TFT35 touchscreen mod

When the monitor is in emulation marlin mode (no touchscreen) M600 works without an hitch, when is not it's stuck in pause and can't recover the printing.

Issue #732, opened 2020-05-30
[FR] (Simplified Status Screen while Printing)


gordo3di:
So far we love the Unified Menu status screen while printing, however would it be possible to implement a very simple screen that appears after "x" seconds. Basically just showing name of file, percentage complete and maybe time all with very big text on a black screen. This way you can see basic info from far away. This would be super useful for people with multiple printers. Also the persistent hotend and bed temp could still be on top.

Things like FR, Fan, Temp, Layer height, Pause, Babystep, More and Stop aren't really important while printing unless something goes wrong.

Then if you touch the screen it would immediately go back to the normal status screen and wait for the new countdown...

Just a thought. I'd love to program it however I think you would spend more time fixing my code than anything... How can I help?

![status-screen_idea_200530](https://user-images.githubusercontent.com/11464057/83334415-63a6c000-a274-11ea-9d63-14abbe909542.jpg)

oldman4U:
Hello gordo3di.

I believe this would be a good idea, but based on the current font used for the UI this could look ugly: But you would like to see the attributes shown in your screenshot, where the Progress is available twice and therefore is redundant.
![Font](https://user-images.githubusercontent.com/54359396/84637482-08243580-aef6-11ea-9d43-6369b7594ce5.png)

gordo3di:
That makes sense. We were just looking for something we could easily see from far away and take away stuff that doesn't matter for general printing. I take it scaling fonts doesn't really work?

oldman4U:
What if it would be possible to improve the readability of the whole UI?

Such a simple status screen, how would you activate it? Automatically after xsec the print started or manually by pressing a button on the interface?

oldman4U:
Sorry....

As far as i know you need an additional, bigger font. And space is limited - as usual.

gordo3di:
> What if it would be possible to improve the readability of the whole UI?
> 
> Such a simple status screen, how would you activate it? Automatically after xsec the print started or manually by pressing a button on the interface?

Our thought was to have it appear after x seconds kind of like a screen saver. Currently our printing screen looks like the attached image.

![IMG_20200617_114654](https://user-images.githubusercontent.com/11464057/84946489-d0b0b700-b0b6-11ea-908a-811a7f56f008.jpg)

It works but sometimes it hard to see the progress at a fast glance especially if you have many printers running.

Thanks for the input

oldman4U:
[...]

PR #1728, opened 2021-03-13
ProgressBar during print

kisslorand:
This PR draws a progress bar during print.
The color chosen is Orange because it is easy on the eye if you like to stare at your TFT yet easily seen from a distance. It's not user configurable (yet).

It might affect the print quality so it will be kept as a draft until it gets the blessing from you.
I could only test it on 32bit system with large enough buffers and had no print quality issues.


thisiskeithb:
> It's not user configurable.

With the menus, icons, and various other TFT color options being configurable, why not?

Guilouz:
Do you have a screenshot of that ?

Maybe progress bar can be enabled/disabled by user in config.ini ?

kisslorand:
> With the menus, icons, and various other TFT color options being configurable, why not?

Because it is a draft, Orange should do it just fine for testing purposes. If it turns out it really doesn't hurt the print quality, than user config will be introduced.

> Do you have a screenshot of that ?

ProgressBar

> Maybe progress bar can be enabled/disabled by user in config.ini ?

Possible but at the cost of wasted codespace. If progress bar is disabled, the upper icons should be moved back to their original space. For this to be possible both positions should be defined and kept in the flash memory.
If really needed, yes, possible.

kisslorand:
Commit [07a92b7](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1728/commits/07a92b7c20501fac651f877905f598cd96a99015) brings some GUI polishing and a very fast progress bar drawing. It makes progress bar drawing faster than any other item refresh during print. 
It makes it the least invasive GUI update amongst all.

gordo3di:
Ooh this is another part of my wish list from a while back https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/732

kisslorand:
I started working on the progressbar enabled/disabled and color customization and it adds RAM usage and extra code to the printing process.
I am not fond of neither of these side effects so I just stopped.
There will be no customization implemented just for the sake of it.

oldman4U:
Which TFT's should I test it with and what exactly? Just the drawing of the bar on the TFT or also the print quality (which is a bit hard to test...).

Mainboard matters or can I use any mainboard?

kisslorand:
@bigtreetech 
It can be merged.

If there will be a real need for customization, it can be done. Until than an orange progress bar is better than none.

bttguy:
Like this feature @kisslorand ! Is this also working with RRF? We just got print status monitoring, in #1743 that needed some special schanes in our fw for be useful? Can you check it on the code I should expect is to work?

kisslorand:
Sorry, I cannot test it other than Marlin.

oldman4U:
bttguy

The PRs are available here so users can test them before they are merged. Once they are merged it is too late. Please test now and let us know. Thank you

bttguy:
Oh, I see @oldman4U . If I wanna test PRs I need to use kissroland's fork where he did the changes and pushed the changes here or is there a button where I can download the fw including this PR? Still leraning github, but willing to learn :)

kisslorand:
It was a draft for 8 days for this very reason, to be tested BEFORE merge.

bttguy:
@kisslorand Since print progress just been added recently, I was busy testing that first. But now I compiled your PR for my TFT 24, and it works as you intended at a first glance also for us RRF users. Both in simulation and in an actual print. So thanks you for your work, next time I will be more careful to test these kind of changes.

oldman4U:
Hi. I always use the repository and must say that I do not know if there is
another (better) way😁

Good to hear it works👍🏻

bttguy ***@***.***> schrieb am Mo. 22. März 2021 um 17:50:

labotecno:
Seem this PR work with tft 35 e3 

You do a great  work with this progress bar .

kisslorand:
@bigtreetech 
Any reason holding this back?

PR #1735, opened 2021-03-15
Reviewed Leveling menus

digant73:
Reviewed Leveling menus in order to allow further feature menus (e.g. M26 etc...).
* Manual Leveling menu:
  1) Reviewed point buttons location. The layout better reflects the target points
* Bed Leveling menu:
  1) Home Offset menu moved from Bed Leveling menu to Tuning menu.
  2) Bed Leveling Layer 2 menu is now available also for MBL. It was previously available only for ABL (BBL, UBL, other ABL).
  3) Z Align button moved to Bed Leveling Layer 2 menu.
  4) shortcut to heat/pre heat menus moved from Bed Leveling Layer 2 in order to be available for the leveling, Z offset, mesh edit and mesh print testing processes.
* Added shortcut to heat/pre heat menus (clicking on the value area) to Home Offset, Probe Offset and MBL menus. 
* minon cleanup

**PR STATUS: READY FOR MERGING**

IMG_20210314_170102

IMG_20210315_215123

IMG_20210315_215134

IMG_20210315_221027

IMG_20210315_215246


DaStivi:
hey there, not sure if i can put this "request" here, even the PR has been commited allready.. but wouldn't it be possible to have 15x15 Mesheditor available? right now its only 10by10...

lightmaster:
@DaStivi You realize how massively overkill a 15x15 leveling grid would be, right?

DaStivi:
> @DaStivi You realize how massively overkill a 15x15 leveling grid would be, right?

depends on how you would see it.. because you get perfect results for bedleveling.... and with UBL it just take the time once!
i could send you plenty of UBL Meshed i did allready, after different maintenance works, ands its strange how different they are each time...

sreichholf:
A build surface should be somewhat flat. If your's requires 225 probing points to work out you should improve your build surface instead of trying to compensate a messy build-surface via levelling.
I'd consider 15x15 complete overkill, too.
Well at least for common size printer.
On a 500x500 Build surface it may be very valid, though.

digant73:
@DaStivi Just increased the max mesh grid size in my PR #1748. The reason for that is just to be compliant with the max size suggested by Marlin fw (15x15 for UBL).
As reported by other users, that limit makes sense only on a big build volume, otherwise it's an overkill.

DaStivi:
> @DaStivi Just increased the max mesh grid size in my PR #1748. The reason for that is just to be compliant with the max size suggested by Marlin fw (15x15 for UBL).
> As reported by other users, that limit makes sense only on a big build volume, otherwise it's an overkill.

thx for just adding it... doesn't matter i think if it makes sense or not can everybody choose for it selfe... why not using it when its possible.. the initial mesh creation take around 15 or 20min, ok but still its the best you can get there.. and before each print its only ah 3 point probe so, its the best you can get, perfect mesh for the bed leveling thats saved and adapted while printing and then also fast before print probing... i don't see why not using it

PR #1743, opened 2021-03-18
rrf auto detect remote print

### LINKED ISSUE #1723: [FR] Separate config.ini for RRF
RRF compatibility is starting to became an active project #1400 . As RRF is a different firmware, it requires different settings in the current config.ini to be able to used. I think the current situation could be improved.

I could imagine two different config files like config.ini_Malrin and config.ini_RRF. This could be beneficial both for easier testing and for new users that would want to set up their TFT's for RRF. This could also give the possibility of some restructureing of the config.ini_RRF as some of the lines are not affect our usecase, only Marlin mode. It could lead to less user error as well that came from a wrong config.

Edit config.ini each time I test a new commit. Or save a tuned config in a hope that no new line added to it, that could cause a possible error.

**Additional context**
Since Bigtreetech released their RRF compatible board, and there are a lot of users who already use BTT board with BTT TFT's with RRF it would beneficial for new users as well as for BTT if some quality of life improvements could be made for RRF users.
In exchange of a simple rename (or somebody could come up with a coder sollution not needed as with the different fw.bins) and some extra caution (but with so many fw.bins it is a proven doable thing) both testing and user experience would be better.
It would also require a bit getting used to from long time users, and in case of a bigger code change two config.ini-s need to be maintained, but this is why I ask it, to discuss the possible pitfalls I dont thought of. At my level of knowledge config.ini is just an overwrite file not a hard configuration.h change, so it works in my head, but open for any thoughts.

In case this is a welcomed change I would be happy to help to set up and maintain a new config.ini_RRF as I am also a RRF user and with a BTT TFT, and in case some codeing is needed I can seek for some help as well as help in the documentation changes this FR makes necessary.

Msq001:
### Requirements
Need to add the following command to simulate the marlin response in the sys file
* `M409 K"job.file.fileName"` in `start.g`
* `M118 P2 S"//action:cancel"` in `stop.g`
* `M118 P2 S"//action::paused"` in `pause.g`
* `M118 P2 S"//action::prompt_begin Resuming"` in `resume.g`

* `M575 P1 S2 B57600` in `config.g` of RRF to set baudrate to 57600(The baudrate of the TFT and motherboard must be the same)
### Description
The TFT will auto goto printing menu when motherboard start printing by DWC or USB

### Related Issues
fix #1723


bttguy:
Too good to be true @Msq001 Huge thanks, can't wait to test it out! Nice addition to #1400

jaysuk:
M0 requires adding to gcode end code if stop.g is required to be ran

PaulG4H:
Tested with an 3.5 V3 and SKR 1.4T board and BTT WiFi Module ESP with latest @Teamgloomy RRF 3:
 - Starting the print from DWC and the display shows the printing dialog, layer, file name...
 - When I pause the job by DWC the TFT shows the Resume and when select on the TFT the print job resumes
 - When I pause the job by DWC it can be resumed by TFT...
 - When I stop the job in DWC the TFT shows the summary

Msq001:
> M0 requires adding to gcode end code if stop.g is required to be ran

M0 will be executed If printing is aborted by DWC or TFT

bttguy:
@Msq001 any cause of #1746

PR #1748, opened 2021-03-19
resolves FR #1741 + increased max mesh grid size + MBL relative initial height

### LINKED ISSUE #1741: [FR] Manual bed leveling Ender 5
Hello

I fix position in script for bed leveling from TFT for ender 5 cause its pointing to wrong position. Now its read correctly MIN MAX Y X axis and point head to right position regarding to icons on LCD and positions. 

image

#include "Leveling.h"
#include "includes.h"

const MENUITEMS manualLevelingItems = {
  // title
  LABEL_LEVELING,
  // icon                          label
  {
    {ICON_POINT_4,                 LABEL_POINT_4},
    {ICON_POINT_3,                 LABEL_POINT_3},
    {ICON_LEVEL_EDGE_DISTANCE,     LABEL_DISTANCE},
    {ICON_DISABLE_STEPPERS,        LABEL_XY_UNLOCK},
    {ICON_POINT_1,                 LABEL_POINT_1},
    {ICON_POINT_2,                 LABEL_POINT_2},
    {ICON_POINT_5,                 LABEL_POINT_5},
    {ICON_BACK,                    LABEL_BACK},
  }
};

void moveToLevelingPoint(uint8_t point)
{
  int16_t pointPosition[5][2] = {
    {infoSettings.machine_size_max[X_AXIS] - infoSettings.level_edge, infoSettings.machine_size_max[Y_AXIS] - infoSettings.level_edge},
    {infoSettings.machine_size_min[X_AXIS] + infoSettings.level_edge, infoSettings.machine_size_max[Y_AXIS] - infoSettings.level_edge},
    {infoSettings.machine_size_min[X_AXIS] + infoSettings.level_edge, infoSettings.machine_size_min[Y_AXIS] + infoSettings.level_edge},
    {infoSettings.machine_size_max[X_AXIS] - infoSettings.level_edge, infoSettings.machine_size_min[Y_AXIS] + infoSettings.level_edge},
    {(infoSettings.machine_size_min[X_AXIS] + infoSettings.machine_size_max[X_AXIS]) / 2, (infoSettings.machine_size_min[Y_AXIS] + infoSettings.machine_size_max[Y_AXIS]) / 2},
  };

  if (coordinateIsKnown() == false)
  {
    storeCmd("G28\n");
  }

  storeCmd("G0 Z%.3f F%d\n", infoSettings.level_z_raise, infoSettings.level_feedrate[FEEDRATE_Z]);
  storeCmd("G0 X%d Y%d F%d\n", pointPosition[point][0], pointPosition[point][1], infoSettings.level_feedrate[FEEDRATE_XY]);
  storeCmd("G0 Z%.3f F%d\n", infoSettings.level_z_pos, infoSettings.level_feedrate[FEEDRATE_Z]);
}

void menuManualLeveling(void)
{
  KEY_VALUES key_num = KEY_IDLE;

  menuDrawPage(&manualLevelingItems);

  while (infoMenu.menu[infoMenu.cur] == menuManualLeveling)
  {
    key_num = menuKeyGetValue();
    switch (key_num)
    {
      case KEY_ICON_0:
        moveToLevelingPoint(3);
        break;

      case KEY_ICON_1:
        moveToLevelingPoint(2);
        break;

      case KEY_ICON_2:
        {
          char tempstr[30];
          sprintf(tempstr, "Min:%d | Max:%d", LEVELING_EDGE_DISTANCE_MIN, LEVELING_EDGE_DISTANCE_MAX);

          int val = numPadInt((uint8_t *)tempstr, infoSettings.level_edge, LEVELING_EDGE_DISTANCE_DEFAULT, false);
          infoSettings.level_edge = NOBEYOND(LEVELING_EDGE_DISTANCE_MIN, val, LEVELING_EDGE_DISTANCE_MAX);

          menuDrawPage(&manualLevelingItems);
        }
        break;

      case KEY_ICON_3:
        storeCmd("M84 X Y E\n");
        break;

      case KEY_ICON_4:
        moveToLevelingPoint(0);
        break;

      case KEY_ICON_5:
        moveToLevelingPoint(1);
        break;

      case KEY_ICON_6:
        moveToLevelingPoint(4);
        break;

      case KEY_ICON_7:
        infoMenu.cur--;
        break;

      default:
        break;
    }

    loopProcess();
  }
}

digant73:
* max mesh grid size is now set to 15x15 (it was 10x10) just to be compliant to the limit suggested on Marlin fw (a 15x15 for UBL). The mesh data are already dinamically allocated entering on the Mesh Editor and released on exit, so no global memory is statically increased. That max limit makes sense to be reached only on big build volume (big printers).
* resolves FR #1741
* MBL: when nozzle is moved on next point, the nozzle is raised (relative coord) by the configurable setting "level_z_raise" (already used for manual bed leveling). It was previously hard coded to 0.2mm and it was absolute.
* Marlin notifiations (with the exception of the useless/frequent "...Ready." notification) are now dispalyed also on the Notification bar so they are always visible when received.

resolves #1741


Cal1sto:
Is not better to set a compilation conditional parameter in Configuration.h this is a printer specification not a printer configuration covered by "config.ini" where this could be changed on same machine ? This way should reduce application size by conditional compilation , same idea could be applied for other part like machine with or without probe, etc ...

bigtreetech:
>  This way should reduce application size by conditional compilation , same idea could be applied for other part like machine with or without probe, etc ...

Yeah, I agree with you on this point, But many users don't like (or don't have the ability) to modify the conditional compilation than recompile firmware, They prefer to change `config.ini ` to achieve different settings. This does lead to large program size and RAM usage.  So, there is a conflict between the convenience of use and the reduction of resource use

kisslorand:
There is an increased tendency of requests to have every new feature configurable, enable/disable-able.
We keep forgetting that the most important job of a printer is to print.

digant73:
> Is not better to set a compilation conditional parameter in Configuration.h this is a printer specification not a printer configuration covered by "config.ini" where this could be changed on same machine ? This way should reduce application size by conditional compilation , same idea could be applied for other part like machine with or without probe, etc ...

What are you refererring to in particular? no new parameters were added in config.ini by this PR

bigtreetech:
Hello, I've changed here https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1748/commits/8dab17e96b783dce933e745a463df59b57766788
Please check. Thank you

digant73:
@ciao, yes I saw the change on manual leveling. It's ok

kmarty:
> * Marlin notifiations (with the exception of the useless/frequent "...Ready." notification) are now dispalyed also on the Notification bar so they are always visible when received.

How can I disable this thing as a user? I'm not quite happy when display every layer and every percent flashes and beeps (I want to see output from DisplayLayerProgress and I saw it on the main screen in the middle windows. Without annoying flashes and beeps).

kmarty:
At least something like:
```
index c20f7ad4..3c2bca26 100644
@@ -227,7 +227,8 @@ void hostActionCommands(void)
     uint16_t index = ack_index;
 
     if (!ack_seen("Ready."))  // avoid to display unneeded/frequent useless notifications (e.g. "My printer Ready.")
-      addToast(DIALOG_TYPE_INFO, dmaL2Cache + index);
+        if (infoMenu.menu[infoMenu.cur] != menuStatus) // Don't show it when menuStatus
+            addToast(DIALOG_TYPE_INFO, dmaL2Cache + index);
   }
   {
```
would be nice (no idea whether it is correct way or not. It works for me).

Issue #1617, opened 2021-02-11
BTT TFT35 E3 V3 26.X June 15, 2020 - Filament runout only works in marlin screen mode not TFT mode. 


telco55:
### Steps to reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior**

**Actual behavior**

### Additional Information
* Include a ZIP file containing your `Configuration.h` or use [Pastebin](https://pastebin.com/) and paste a link in this issue.
* Provide pictures or links to videos that clearly demonstrate the issue.

telco55:
Reproduced several times. 
With the runout sensor installed and enabled in the Touch Screen mode will not allow a print to start. 
The mesage is "filament runout" and the print cancels and shuts down the temps.
Fortunately I am able to print in Marlin Mode just fine. 
I will have to use Marlin Mode till this is fixed. 
I need to tell everyone to not buy a BTT TFT screen till it this issue is resolved.
I have updated firmware on the SKR mini E3 V2 board several times with no effect.
   Please fix this issue so I can use the nice color screen. 
There several other issues with the screen like it hangs the screen in "Loading" after a print pause. This requires a print restart(power cycle). 
When switching back and forth between Marlin screen and TFT screen I found that the Probe Offset was not shared between screen modes. I am not sure if that is normal or a bug. I also had in a Marlin mode session during a print the that the screen became unresponsive  I was unable to get the screen menu to change.  
Thanks    Please Help

oldman4U:
Please update to the most recent TFT firmware.

Then read previous made tickets related to your issues. 

For those issues you see on your system but you can not find a ticket for make a new ticket per issue. Include all relevant information incl which filament sensor you use and where it is connected.

This is a friendly reminder, that this repository is about the TFT firmware and not a user group.

Thank you

scathappens:
I have the same issue, have had a ticket going for this, it was supposed to be fixed however i have updated the firmware on the TFT35 today to the 04/02/2021 release and it has not solved this issue 

when the filament sensor notices a runout (printing from SD) the head parks as it should, then when i press resume it does nothing....then after a pause of about 30-60 seconds it says heater timout, when i press reheat it the heater turns off.
if i hit stop print at this time i get a message saying 'busy processing, please wait'

so the runout sensor is still not functioning as it should in the latest firmware (dated 04/02/2021)

[...]

Issue #1612, opened 2021-02-07
[BUG] Filament sensor does not work after the update


Chris1231:
i use that BTT TFT 35 E3 V3.0
after the update to the version 3.0.26.x it is no longer recognized that there is a filament in the filament sensor.
no matter whether filament is inserted or not
I use a simple sensor (not a smart one) connected directly to the TFT.
with the old version 3.0.25.3 it worked fine.

do I have to use other settings compared to the old version?

I hope you can help me

YBeyin:
Thats about the marlin. Lol i just open a topic about that bug. Compile your motherboard's firmware from Marlin 2.0.6 thats works.

Chris1231:
Thanks for the quick reply, I've tried it, but it's still the same problem :(

YBeyin:
> Thanks for the quick reply, I've tried it, but it's still the same problem :(

Are you plugged your sensor on motherboard or Tft?

Chris1231:
connected directly to the TFT

YBeyin:
> connected directly to the TFT

I am pretty sure something wrong about pins. Can you check your pins with multimeter if there is voltage?

Chris1231:
i checked.

at the pin PA15 with filament 0V and without filament 3.3V

Apfelbrot34:
Good evening I have the same problem too. My hardware is: SKR 1.4, TFT35 V3.0 and BTT Smart Runout Sensor. The sensor is connected to the TFT. The signal output is from 0V to 2.6V. input voltage is at 3.3V at Fil_Det. Firmware is Marlin 2.0 bug fix 20008 according to config. TFT is V3.0.26.x. The sensor is simply not recognized by the TFT. If the plug is unplugged, there is no runout signal or even True or False, there is no reaction. M114 E we output from the Linear Advanced board is switched off.

Grex3D:
Same issue, more easy to use with the dedicated  fil-det pin on screen (tft70) but nothing work....

YBeyin:
[...]

Issue #1611, opened 2021-02-07
[BUG] Btt Smart Sensor Does Not Work on the Tft35 E3 V3 with Marlin 2.0.7 (Works with 2.0.6)


YBeyin:
I just bought brand new Btt Smart Filament Sensor for my Ender 3 Pro. Everything went perfect untill test it. Upgraded my motherboard's firmware, Tft35's firmware but sensor would not work. It was like dead. At first i had a lot of panic. I did pull off the cable of sensor even my printer while working, there is nothing happened. My printer keept working.

Digged in the github and searches on the internet not found a solution. So i decided to contact with the Rob from 3D Printscape youtube channel, asked for his help and he helped a lot. At final he shared his marlin files with me. The result was incredible. My sensor begin to work perfectly. Then we realise all that was issue about the marlin itself. My sensor did not work with 2.0.7 but worked perfectly with the 2.0.6

Now my sensor plugged on the Tft's Fil-Det slot, using Tft's SdCard input, using in touch mode (not in marlin) and everything perfect.

I just want to let you know if you are having issue about your sensor, try it with marlin 2.0.6

My Printer's Specs:
Ender 3 Pro
BL-Touch
Btt Skr Mini E3 V2.0
Btt Tft35 E3 V3.0
Btt Smart Filament Sensor (Plugged on the tft's Fil-Det slot)

### Steps to reproduce

1. Plugged the sonsor
2. Upgraded the firmware of motherboard
3. Upgraded the firmware of tft.

### Hardware Variant
Btt TFT35 E3 V3
### TFT Firmware Version & Main Board Firmware details
 Lastest Tft Firmware & Marlin 2.0.6

DaStivi:
look at the other SFS issue...   when you upgraded to 2.0.7 did you enabled linear advanced eventually? i think SFS does not work with LA enabled..

https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/1612#issuecomment-782926663

killrob:
i made this work, see here
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/907

oldman4U:
The information that it does not work is nonsense and could lead users into the wrong direction. Please close this ticket.

Issue #1538, opened 2021-01-19
A few issues with pause and filament runout TFT35 e3 v3 and SKR mini E3 V3 (04/02/2021 firmware update)


scathappens:
SO am new to Github but seems like the place to get some answers. 

I have installed a TFT35 E3 V3 and an SKR mini e3 V2 in my printer and there are a couple of bugs i cannot work out. 

I have the filament runout sensor attached to the TFT35 and i am printing from the SKR SD slot.

when the filament sensor notices a runout (printing from SD) the head parks as it should, then when i press resume it does nothing....then after a pause of about 30-60 seconds it says heater timout, when i press reheat it the heater turns off.
if i hit stop print at this time i get a message saying 'busy processing, please wait'

so the runout sensor is still not functioning as it should in the latest firmware (dated 04/02/2021)

I get the exact same reaction when i use the pause feature aswel, the head parks but then it will not resume in the same manner described as above

Any advice would be helpful

oldman4U:
Add a M75 to your start gCode

Which firmware version are you using, send a screenshot from Info window

scathappens:
Thanks for getting back to me, really appreciate it, i have had to take a crash course in marlin to get this printer working lol, its been a rough ride 

filament runout is connected to the TFT 
SKR mini Firmware 
![image](https://user-images.githubusercontent.com/77690628/105103001-b3f2bc80-5aa7-11eb-827a-bb3fbbd25a1d.png)

TFT 35 e3 V3 firmaware (i updated it to the latest version today) 
![image](https://user-images.githubusercontent.com/77690628/105103157-016f2980-5aa8-11eb-87b3-95ab2f1d7765.png)


adding M75 to my start gcode will solve the problem of the cancel g code not running? or is it something else?

oldman4U:
M75 helps to recognize the sensor.

Pausing the print from on board sd does not work like mentioned before.

scathappens:
thanks for that will do some more experimenting

oldman4U:
Any news?

scathappens:
@oldman4U i think this was also to do with the other issue of the runout sensor and the onboard sd slot...that issue is apparently sorted but i am not sure if the firmware has been updated...waiting on clarification and then i will install and see

oldman4U:
Any news?

Please help the community and close the ticket once you do not need it anymore. Thank you

scathappens:
how do i close a ticket?!?!

[...]

PR #1752, opened 2021-03-20
possible bugfix for filament sensor not working connected to TFT

digant73:
This PR is meant ONLY for testing puropose (it will be marked as draft) for all the users having
the filament sensor not properly working when connected to TFT.
I kindly ask all people affected by the problem with the sensor and SFL on TFT to test this PR and provide their results.

In order to use this PR the following setting MUST be enabled on Marlin (Configuration_adv.h):

M114_REALTIME

This PR could fix the following reported bugs:

#1617
#1612
#1611
#1538

**PR STATUS: NOT TO BE MERGED**


oldman4U:
It has been working perfectly when printing from TFT storage for me. Do you
expect improvements there and also when printing from on board sd card?

Will test in about 12 hours.

Thank you

digant73 ***@***.***> schrieb am Sa. 20. März 2021 um 22:33:

digant73:
@oldman4U 

I expect that from TFT it will continue to perfectly work for you and also for the users affected by the reported bugs.
Which are the problems from onboard?

oldman4U:
👍🏻

Onboard I see it interrupting the print perfectly every time with the
correct Filament runout information on screen. The problem is that it
resumes print correctly sometimes but most of the time it shows Filament
runout again one or two times before print finally resumes correctly. I
have been testing this around 100x and could not find a pattern. Printing
from TFT storage I have been testing around 30x and it never failed and
always resumed the print correctly. Retract (and purge at Pause position)
has been working 100% independent of the storage I used.

Maybe I will upload some movies tomorrow.

Thank you

digant73 ***@***.***> schrieb am Sa. 20. März 2021 um 22:43:

digant73:
could you provide the output from pronterface when you have the problems from onboard?
I will test it also from my side

oldman4U:
ESP32 via WIFI should also do the job. Right!?

digant73 ***@***.***> schrieb am Sa. 20. März 2021 um 23:00:

digant73:
Everything you can use to monitor the comunication with marlin is good

oldman4U:
Ready for a test and Marlin is already updated with M114 RealTime enabled.

Anything I have to keep in mind regarding the TFT firmware besides activating Filament sensor?

digant73:
@oldman4U nothing in particular. just verify if it works. From onboard it will be usefull to get the output to/from mainboard, so I could understand the cause for the multiple resume retries

Chris1231:
I just loaded the last TFT firmware.
M114 REALTIME is enabled in Marlin. but unfortunately it still doesn't work. I still get the message no filament.

Sensor connected to TFT
Last Marlin version
Last TFT version

Is there anything else I have to consider?

otherwise I would first switch back to TFT firmware 25.3 :(

digant73:
@Chris1231 stupid question, as "last TFT fw", do you mean this PR not the current BTT main branch, right?

Chris1231:
my mistake,
the last current BTT main branch.

I'll try the PR right away.

oldman4U:
OK, let's use this ticket for any further tests.

I will upload a movie in a few minutes which shows TFT35 E3 v3 with digant73's PR (even it does not matter because it works the same wit BTT FW) and a sensor connected to TFT printing from TFT. Just want to finish another test, which is not really related.

oldman4U:
https://user-images.githubusercontent.com/54359396/111909060-0591d600-8a5c-11eb-86bd-8584a36891c4.mov


https://user-images.githubusercontent.com/54359396/111909083-1e01f080-8a5c-11eb-8bc3-22eb68f32ccb.mov


https://user-images.githubusercontent.com/54359396/111909094-2fe39380-8a5c-11eb-882b-45ecf6bab363.mov


https://user-images.githubusercontent.com/54359396/111909126-4558bd80-8a5c-11eb-8761-ecf9ecfa2941.mov

Chris1231:
I have the PR firmware on the printer. Unfortunately without success.

i use the old z-endstop, could that be the problem?

oldman4U:
What do you mean with old z endstop?

Have you seen the movies which I uploaded?


Chris1231 ***@***.***> schrieb am So. 21. März 2021 um 16:26:

Chris1231:
@oldman4U yes have seen the movies, so I ask if the z-endstop from the ender 3 is no longer compatible with the tft firmware.

oldman4U:
I do not understand what you mean. Sorry.

How is the endstop related to filament runout? My filament sensor works
technically like an endstop or vice versa.  Both define a status to be on
or off.

Chris1231 ***@***.***> schrieb am So. 21. März 2021 um 16:45:

Chris1231:
that you don't understand me is more a result of my english.

but you have now clarified the question.
but why does it work for you, but not for me?

GrahamJB:
Do you mean you are reusing the microswitch from a limit switch as the microswitch in a runout sensor?

Chris1231:
@GrahamJB yes

oldman4U:
In which order are the 3 wires connected to your TFT?

Chris1231:
it's only 2, 3.5V and sig. when triggered, the 3.5v are switched to PA15.

digant73:
@Chris1231 Did you configure filament runout sensor to SMART or only ON?

Chris1231:
Only ON

digant73:
Extruder position is used for SFS not for ON/OFF sensors.

The polling is made based on your current tool:

if tool 0 -> FIL_RUNOUT_PIN is used
if tool 1 -> FIL_RUNOUT_PIN_1 is used
....

If you have a BTT TFT35 V3, by default the following pins are used:

FIL_RUNOUT_PIN PA15

The following pins are by default disabled, so if use one of that you must uncomment the the proper line

// #define FIL_RUNOUT_PIN_1 PC12  //Extruder T1  Attention: Is the same pin as PS_ON.
// #define FIL_RUNOUT_PIN_2 PB10  //Extruder T2             Is the same pin as USART3 TX.
// #define FIL_RUNOUT_PIN_3 PB11  //Extruder T3             Is the same pin as USART3 RX.
// #define FIL_RUNOUT_PIN_4 PA0   //Extruder T4             Is the same pin as USART4 TX.
// #define FIL_RUNOUT_PIN_5 PA1   //Extruder T5             Is the same pin as USART4 RX.

Furthermore, "runout_noise_ms" and "fil_runout_inverting" are used.

Try setting "runout_noise_ms" to 1000 (1 second).
Try setting "fil_runout_inverting" to true/false

If you connected the sensor to the correct pin, it must work

Chris1231:
short question, if I measure to GND with the PA15 multimeter, 3.5V are present, could that be the problem for me?

Chris1231:
i think i have the solution to my problem.

I changed the wiring.
instead of switching 3.3V to PA15,
I switch PA15 to GND.

after a short test, it seems to work

digant73:
@Chris1231 you should obtain a working runout changing the value for "fil_runout_inverting"

Chris1231:
@digant73 I had already tried that before, without success.

digant73:
@Chris1231 Ok. A usefull info for other similar problem as your

oldman4U:
Using only two wires you have to use gnd and signal but i am wondering how this makes a difference in firmware!?

But it works now with the "latest" firmware!?

Chris1231:
i am using the PR from digant73 and it works now.

oldman4U:
So it will also work with the current master firmware.

Please help the community and close the ticket once you do not need it anymore. Thank you

oldman4U:
Please close your ticket not this PR😆

digant73:
@Chris1231 please, in you bug report put and update on you first post (where you initially described your problem) and explain where the problem was and how you fixed it. It will help other users having the same problem to quickly solve the problem without openung a new bug report. Thanks

Chris1231:
thanks for your help, i closed my ticket and updated the first post.

oldman4U:
Should we expand the verified bugs and add known issues to it and mention la and that it does not work with smart filament sensor?

digant73:
I would say yes

oldman4U:
👍🏻

digant73:
closing the PR due to no other users having any problem with SFS not working when connected to TFT

PR #1755, opened 2021-03-21
fix extrude mm update not in realtime bug & RRF support improve

Msq001:
### Description

* fix extrude mm update not in realtime bug
* fix M221 D para to HIGH bug (`T0 P0` need in sys `config.g` avoid `No tool selected` error)
* fix RRF ABL staus M420 bad command


bttguy:
Save for #1400 for future documentation :)

PR #1769, opened 2021-03-25
Reduce RAM and memory usage & rename Macros.h to RRFMacros.h

guruathwal:
- Reduce RAM and memory usage. (Use already defined functions and variables).
- Rename **Macros.h** to **RRFMacros.h**
- Fix text display bug in Popup dialog in Macro menu when using secondary language.


oldman4U:
Gurmeet is back👍🏻

PR #1774, opened 2021-03-26
Move and ExtruderTune menu unit steps bugfix

### LINKED ISSUE #1772: TFT 34V2 manual move screen units
### Description

On the manual move screen, 10mm is selected for distance travel. Move axis 10mm. Change screens and come back to manual move screen. Increment is changed on screen, but 10mm is still selected. Go to press move button .1mm and it moves 10mm.

### Steps to reproduce

See above

**Expected behavior**
Increment to  update when changing screens

**Actual behavior**
previous increment loaded but icon not changed.

### Hardware Variant
TFT35V2
<BIQU_TFT35_APP1_V2.0.26.x>

### Additional Information

kisslorand:
### Description

It's a little bugfix for the Move and the ExtruderTuning menu.  Entering into one of these menu, the unit steps displayed the default steps, not the actual ones.

Fixes #1772


digant73:
No need to change the initial labels and icons. they are aligned with the initial index (used as place holder). It's the same in many other menus.

kisslorand:
I would agree with you but...
If the labels and icons would have been initialized like this in the first place, this bug wouldn't occur. The coder would have made the initialization later in the code for the labels and icons in question.
I might even consider to initialize at definition stage with `NULL` those icons and labels that are going to change during operation.

digant73:
@kisslorand no, I was referring to the replace you did like:
{ICON_RESUME,                  LABEL_START},

with

{ICON_BACKGROUND,       LABEL_BACKGROUND},

We used to provide a place holder for all the entries that will be present (even if the entry's values can be changed).
Only in case a entry is optional (so it can remain blank) we initialized the entry with XXX_BACKGROUND. See for example BedLevelingLayer2.c where entry number 1 and 2 can remain blank if BL is not UBL.

kisslorand:
@digant73 
I understood you the first time. I reverted the changes but I still maintain my point of view. If it would have been like I changed the initialization than the bug have been avoided.

digant73:
yes, some things could have been implemented in a different way for sure. I'm also making some cleanup in all the menus just to have the same structures.
I also found and fixed the bug with emulated M600 that was purging instead of retracting. It was a bug in the handling of infoCmd in interfaceCmd. That handling could also create a loop creating a freeze in some other function called by parseCmd in interfaceCmd.c. Just working on that now

PR #1785, opened 2021-04-01
Bug fixes + reviewed Printing API + some menu optimizations

### LINKED ISSUE #1736: Probe offet for mesh validation
Add Z offset for UBL mesh validation (business card mode?)
When validating the mesh data the Z position is always zero. 
Example: G29 performed, mesh data stored. Go to mesh editor, pick a point, start. Z position is always to low with shim under nozzle. Go back to babystep, increase babystep height by paper thickness, re-enter mesh editor, validate.

Option added for business card mode to set the business card thickness

Below is a copy and paste of the config.ini. Could this be applied to the Z position for mesh edit?
#   Unit: [position in mm]
level_z_pos:0.05

**Additional context**
None

### LINKED ISSUE #1787: [BUG] (After #1748 a notification bar flashes and beeps even when menuStatus is active thus it shows redundantly)
### Description

Since #1748 an annoying beeps and flashes of M117 are redundantly displayed on main screen with menuStatus enabled (very annoying in combination with DisplayLayerProgress plugin in Octoprint).

### Steps to reproduce

1. Just start to print

**Expected behavior**
When menuStatus is active, no further notification/popup is necessary. It is already there.

When menuStatus is active, just show it in status window. No notification bar, no beep.

**Actual behavior**
Displays in duplicate the same message (and beeps - which can be disabled, but when I disable it, what else did I mute?)

### Hardware Variant
TFT35 V3.0
TFT Firmware: from commit 201cd68e (Thu Apr 1 03:20:04 2021 +0000)
Marlin: bugfix-2.0.x, commit [600ef1e47c](https://github.com/MarlinFirmware/Marlin/commit/600ef1e47cf905bec1e76b35f3912863ade4fff0) (2021-04-01)

### Additional Information
IMG_0825
IMG_0827
IMG_0830

Something like:
```
index c20f7ad4..6d0520fe 100644
@@ -227,7 +227,8 @@ void hostActionCommands(void)
     uint16_t index = ack_index;
 
     if (!ack_seen("Ready."))  // avoid to display unneeded/frequent useless notifications (e.g. "My printer Ready.")
-      addToast(DIALOG_TYPE_INFO, dmaL2Cache + index);
+      if (infoMenu.menu[infoMenu.cur] != menuStatus) // Don't show it when menuStatus
+        addToast(DIALOG_TYPE_INFO, dmaL2Cache + index);
   }
   {
```
would be really nice (no idea whether it is correct way or not. It works for me).

digant73:
**IMPROVEMENTS:**
* improved Printing API: all the logic for printing is handled by the Printing API. Now, the logic is no more spreaded also in parseAck.c and interfaceCmd.c. That makes more readable the code and more easy to debug
* optimized different menus (e.g. heat, fan, extrude, speed etc..): different pieces of code were very similar in many menus. Now they were made shared to reduce about 1200 bytes of flash.
* implemented #1736: now the Mesh Tuner add a shim, for ABL types (not for MBL). The Mesh Tuner will raise the nozzle by the configurable value "level_z_pos" in config.ini to allow the use of a paper to fine tune a mesh.
* shim applied to Probe/Home Offset menus: as already happens for Manual Leveling menu, now also for those manual menus, the nozzle will be raised by the configurable value "level_z_pos" in config.ini to allow the use of a paper to fine tune Z offset.
* disabled ABL before performing Probe Offset and Mesh Editor: ABL is now always disabled before setting Z offset or tuning a mesh from Mesh Tuner. This avoids to apply any mesh that will wrongly influence the final result (e.g. Z offset set to -1.0 but applying also an hidden mesh +0.3 on the probing point).

**FIXES:**
* fixed a bug in function sendQueueCmd(): the lost of one gcode from the infoCmd queue was produced when functions printPasuse() and printAbort() were invoked.
  Now, firstly dequeue the gcode to avoid a possible reprocessing or infinite nested loop in case the function loopProcess() is invoked.
* added critical section to function printAbort(): as already present on printPause(), it avoids possible multiple invokations
* fixed a bug introduced by PR #1748: the MBL menu was wrongly moving to initial height "level_z_raise" instead of "level_z_pos" for each probing point
* fixes #1787

fixes #1787
resolves #1736

**PR STATUS:** ready to be merged


Guilouz:
Do you think is possible to display printing menu when printing from serial like with Octoprint ?

digant73:
@Guilouz that should be already possible. the TFT fw receives the acks from marlin and in case another host (not the TFT) is printing, the TFT fw will switch to the Printing menu

Guilouz:
@digant73 with your PR or with latest master ?
Because with lastest master it’s not the case.

digant73:
with the previous and current master it should already work.

Guilouz:
> with the previous and current master it should already work in case you have "onboard_sd_support" set to enabled in config.ini and Marlin sends the Ack with the string "File opened:" indicating the file beeing printed.
> Probably the monitoring of an external host was added recently and the check on "onboard_sd_support" was not yet properly changed in parseAck

Nope, not working, screen stay on main menu, only M117 messages are received.

digant73:
you should start a print from the onboard SD not controlled by Octoprint

bigtreetech:
> Do you think is possible to display printing menu when printing from serial like with Octoprint ?

Marlin will only inform the TFT that the printer in the printing state  when the printing starts from the media of Marlin. 
When octoprint controls to print the SD card files in Marlin(motherboard), the screen will automatically go to the printing menu. If octoprint prints files on octoprint side. TFT can't go to printing menu automatically

digant73:
@bigtreetech yes, the only problem is that the switch to printing menu occurs only if onboard_sd_support was set to ENABLED in config.ini.

Guilouz:
@digant73 I have an issue with your fix #1787, when I set temp manually, I have always toast notification "Printer ready" with sound.

Guilouz:
> > Do you think is possible to display printing menu when printing from serial like with Octoprint ?
> 
> Marlin will only inform the TFT that the printer in the printing state when the printing starts from the media of Marlin.
> When octoprint controls to print the SD card files in Marlin(motherboard), the screen will automatically go to the printing menu. If octoprint prints files on octoprint side. TFT can't go to printing menu automatically

@bigtreetech @digant73 With DisplayLayerProgress plugin for Octoprint it's possible to obtain the status of printer, see : https://github.com/OllisGit/OctoPrint-DisplayLayerProgress/issues/177#issuecomment-771976367

Do you think is it possible to use it in ParceACK to switch to printing menu when status is "PRINTING" and "PAUSED" ?

lightmaster:
>     * to allow the use of a paper to fine tune a mesh.

Fine tuning with a piece of paper using the UBL menu was working perfectly fine before. Stick paper on bed, go to point, adjust z-height until paper was a little snug under nozzle. Go back to the same point to verify the offset was correct.

Now that all goes fine, however the z-offset that it sets it to is roughly 0.180mm too low, meaning that when I tell my printer to go to Z0, it hits the bed far enough to jump a tooth on the belts. If I go to Z0.180, then that's where Z0 should be.

Since fine tuning with paper was working correctly before this commit, is it safe to just revert this whole commit to get it working again? I assume this PR is centered around a single fix/issue like pretty much all repos do, so removal of this PR should just remove the code that messes with UBL fine tuning, right?

lightmaster:
This is definitely a collection of multiple PRs piled into one, not just a single PR to fix a single thing, so can't easily remove the code messing up mesh editing without screwing up the commits after this one. Ended up having to rebase all of my config changes onto the commit before this one, removing all the commits after this one entirely. At least mesh fine tuning isn't broken though.

sarvenn:
In the past this setting below only applied to corner leveling (z position). But after this PR this Z value also effects the Z height initial position while doing manual mesh bed leveling. This causes an issue.

Z Position (Manual Leveling)
Lower z axis to this position after reaching a leveling point.
Format: [level_z_pos: Z]
Unit: [distance in mm]
level_z_pos:0.2

New logic overrides my marlin config setting. (configuration.h)

The "Manual Probe" provides a means to do "Auto" Bed Leveling without a probe.
Use G29 repeatedly, adjusting the Z height at each point with movement commands
or (with LCD_BED_LEVELING) the LCD controller.
*/
#define PROBE_MANUALLY
#define MANUAL_PROBE_START_Z 0.2

This parameter allows me to set my desired Z height while doing manual mesh bed leveling. But TFT config file overrides it. It should only effect the cornel leveling Z height.

I prefer to do corner leveling while my nozzle is at Z0. I don't use a paper. I set my nozzle to bed distance at 0 at corner leveling.
Then I do the manual mesh bed leveling. I expect the nozzle to stay at Z0.2 position and adjust my nozzle to slightly touching to the bed at each MBL points.
With this method I almost do not need to adjust a nozzle offset value at all (may be +/-0.01-0.02). I got perfect first layers.
But after this PR, I set this parameter to level_z_pos:0, corner leveing is OK, but at manual MBL, nozzle first stops at Z0.2 for a very small time, then goes to Z0 which causes my nozzle to hit my bed( this is risky, I already damaged my bed due to this yesterday). This is an unwanted behavior.

Can we go back to previous logic please? I may beg for it :) Please exclude mesh leveling Z position from this logic. It can be set in marlin config, or just seperate corner leveling and mesh leveling initial Z positions as different parameters. Please, please, please.

Z Position (Manual, Mesh Leveling, Probe/Home Offset and Mesh Tuner)
For manual leveling and MBL, lower z axis to this absolute position after reaching a leveling point.
For Probe/Home Offset and ABL in Mesh Tuner, raise z axis by this relative position after reaching a leveling point.
Unit: [position in mm]
Value range: [min: 0, max: 2000]
level_z_pos:0.2

lightmaster:
@sarvenn I'm keeping my fork at the commit before this was applied, since this has absolutely destroyed my ability to do any mesh editing with the LCD, regardless of whatever I set `level_z_pos` to. Its way over my head to be able to undo just these changes while retaining the future commits or fix this to actually work. Feel free to grab my fork and set configuration.h and config.ini to your settings if you want.

sarvenn:
Thnx @lightmaster. I hope our developers will solve the issue. Actually there wasn't any issue this PR. I know this PR also solved another user complaint about the need for mesh tuner Z offset, but cause another one, a bigger one.
I also know that problem. When I select a point in the mesh edit menu it doesn't stays at the right height but It wasn't an issue at all, because whenever I need to edit a point in my mesh, I don't use the mesh editor, I do the MBL from the beginning. It was more healthy, convenient and reliable.

@digant73 hello,
I may help you to understand the issue if you like. Unfortunately I don't understand coding so I can't review the codes and give any suggestions. But I can share you videos if you like. I may show the problem if you can not replicate it.

My steps:
set level z to 0.
heatup the bed to 60C and nozzle to 140C.
Do the corner leveling without using a paper shim. I set my nozzle to slightly touching to the bed. (Z0 is really Z0).
Then I do 5x5 manual mesh bed leveling. In this case I want my nozzle to park at Z0.2 and wait for my adjustment. But it goes to Z0. On some points my nozzle hits the bed. My marlin config is Z0.2 for bed leveling park position `#define MANUAL_PROBE_START_Z 0.2` but TFT config setting overrides it. I set my nozzle to slightly touches to the bed. With this method I do not require a Z offset at all becaus my corners are truely set to Z0 and my mesh points are offsetted to true Z0 position. This only applies to manual mesh bed leveling ofcourse.

digant73:
@lightmaster The Mesh Tuner should increase the height by the shim. E.g. if the probing z height is 0.1 and "level_z_pos" is 0.2 then the nozzle will be moved to Z 0.3. This is the height that will allow the use of a paper (having a 0.2 thickness). If you will lower the nozzle to 0.2 the resulting z height will be 0.0 (0.2 - 0.2 of the used shim). Can you provide an example of the z height (probing z height) you see in the mesh editor and the z height you see/set in the mesh tuner, and the value you will see coming back to mesh editor?

@sarvenn I understand your problem but you are making a wrong usage of the param configurable in config.ini. We cannot provide a configurable parameter for each menu and for each feature enabled/disabled in Marlin. if you used a shim of 0.2 for manual leveling (4 knobs), that value should be good also for all the other features

sarvenn:
> ... I can remove the use of "level_z_pos" but then many other people will complain of that the nozzle is moved to high from the bed (as it was already reported). "level_z_pos" should be used in all the kind of leveling instead of using 0 for that, 0.2 for this etc...

@digant73 I encountered the same, but never considered that as an issue since I never needed to re-adjust individual points. If I needed to remap my mesh for any reason, I did the MBL from the beginning (first of all checking my corner levels). So that, this never caused an issue. I don't like to use paper because it is an inconsistent method and requires to set a nozzle offset to nail the perfect first layers.
As I expained I set my nozzle to slightly touching to the bed I almost do not need a nozzle offset value for manual mesh bed leveling. I know this is not applicable for probing applications, but still adjusting the corners at real Z0 position is beneficial.

Since marlin has the nozzle initial height set feature, people should config their marlin accordingly, right? I learnt this from the developers here and learnt how to configure my own marlin firmware, thanks to you all. But this change upset me a lot. I wish I could noticed that before it was merged :( Old logic was perfect.
Can't we find another way for those people complaining about their nozzle parks to height while checking individual points at Mesh Editor, and revert this to old logic?

sarvenn:
🙏🏻

lightmaster:
> @lightmaster The Mesh Tuner should increase the height by the shim. E.g. if the probing z height is 0.1 and "level_z_pos" is 0.2 then the nozzle will be moved to Z 0.3. This is the height that will allow the use of a paper (having a 0.2 thickness). If you will lower the nozzle to 0.2 the resulting z height will be 0.0 (0.2 - 0.2 of the used shim). Can you provide an example of the z height (probing z height) you see in the mesh editor and the z height you see/set in the mesh tuner, and the value you will see coming back to mesh editor?

My procedure is to first find that exact z offset of the probe vs nozzle:
1. Heat up Bed to first layer temp (70C) and turn off bed leveling.
2. Set Probe offset to 0.
3. Use `G33 P1` to probe and find the exact height of the printer and also set Z0 to be the exact point at which the probe activates on the bed.
4. Heat up Nozzle to average first layer temps (220C as I use a lot of PLA+)
5. Very carefully lower the nozzle down onto a piece of paper so that it is snug but still possible to move the paper without it folding up when pushing forward.
6. Set the Probe offset to the current Z height (ie: if nozzle is at Z-16 when its snug on paper, set offset to Z-16)
7. Home again and send nozzle close to bed, then verify that Z0 is now the exact height of the paper.

To create a mesh, I do:
1. Heat bed to 70C
2. Use `G29 P1` or the LCD menu option to create a new mesh.
3. Once finished with initial probeable points, fine tune those points by (prior to this PR messing it all up):
  a. Heat up Nozzle to 220C
  b. Use menu to send the nozzle to the first point and adjusting its offset such that a paper is again snug but doesn't fold up when pushed forward.
  c. Repeat with all points located on the build plate.
  d. Use `G33 P3` several times to fill in points that lie outside of the probeable area.
4. Now turn on the mesh and verify that at all points, Z0 is the exact height that a piece of paper becomes snug but still moveable at.

Doing this allows me to print a perfect first layer every time as long as I am using a filament that uses the same first layer bed temp. I have a second mesh for ABS/ASA as those use 90C for the bed temp which causes different amounts of thermal expansion on the bed plate.

The problem with this PR is that when I try set a points offset to say Z0.050, it actually gets set to Z-0.150. This causes the nozzle to hit the bed hard when going to Z0 and and when the first layer tries to print at Z0.2, the nozzle is so close to the bed that the filament can not be extruded and trips the filament detection sensor. If you absolutely refuse to fix this, can you at least make a PR that sets all of this as an Opt-In feature that is enabled in configuration.h, so that with it set to `False` it behaves exactly as before? Although I do think the auto disabling and enabling of the bed mesh when probing is a good idea regardless of whether shims is enabled or disabled.

Note that when settings `level_z_pos` to 0, I noticed no significant change in how it actually functioned, which is why I ask for an option to revert back to the previously used code.

digant73:
@lightmaster Ok. what you did is clear but the main error you did is on setting Z offset (with Probe Offset). Z0 should be the height at which the nozzle starts to touch the bed (not a paper or any filler gauge you use). With the old code the mesh tuner was also not considering the shim of the paper. So, using the Mesh Tuner, if you fine tuned z 0.60 using a paper, that value was taking into account the paper while 0.60 should be the height at which the nozzle is touching the bed. The result for you was that all the meshes were too high from the bed by the paper thickness you used in the Mesh Tuner.
Try to do the following and let me know the results you get:
1) use Probe Offset menu to set Z0. Also this menu will allow you to use a paper. So if you configure 0.100 as shim (so you use also a paper having a thickness of 0.100) in config.ini, when the paper starts to snug with the nozzle save Z offset. The real z offset will be the height reported on the GUI - 0.100 (you will see it disabling probe offet menu or entering in the menu). That value will allow you to have a Z0 with the nozzle touching the bed.
2) After step 1, perform UBL process.
3) after step 2, use the mesh editor/tuner. If your probe is almost precise, there should not be to fine tune anything.
In any case, if you move to the tuner just to verify your meshes, you will see that the nozzle will be moved to the z mesh height + the shim. With the additional shim you have the possibility to insert the paper and verify if you need to lower the nozzle by a bit.

The old editor didn't give the possibility to use the paper because the nozzle was moved exactly to the probing point (by definition, to the point where the nozzle should touch the bed). That problem was reported by some users (they followed the leveling procedure reported above by me) asking to fix it.
In your case the old solution worked better due to the procedure you used (set Z0 considering the extra shim of the paper). Unless you set your slicer with a first layer width with a high value (e.g. 150%-200%), you should normally need to use babystep to lower the nozzle by the paper thickness.
It's not a problem to revere the logic of the current implementation if bugged, but please try the above suggested procedure first and provide the results. All the leveling menus are currenlty aligned (they use level_z_pos as shim). The only exception could be made for MBL due to the fact that Marlin also allows to set the initial Z for each point and also MBL process requires the use of a paper (while ABL doesn't). For that reason, the Mesh Tuner doesn't use any shim for MBL. The meshes already include it

@sarvenn I will remove the application of the level_z_pos from the MBL procedure.

sarvenn:
Thnx @digant73 for listening and taking action. I appriciate it.
After the fix I will try and let you know about the results.
Will you raise a new PR? How can I follow the news?

digant73:
@sarvenn if you are able to make few changes in the source code and compile the firmware you could simply comment in the file MBL.c the following lines:

line 20:    //probeHeightStart(infoSettings.level_z_pos, false);  // raise nozzle
line 243:  //probeHeightStart(infoSettings.level_z_pos, false);  // raise nozzle

That is enough to avoid to move the nozzle to Z height provided by "level_z_pos". In a next PR a will make it configurable in Configuration.h

sarvenn:
> @sarvenn if you are able to make few changes in the source code and compile the firmware you could simply comment out in the file MBL.c the following lines:
> 
> line 20: //probeHeightStart(infoSettings.level_z_pos, false); // raise nozzle
> line 243: //probeHeightStart(infoSettings.level_z_pos, false); // raise nozzle
> 
> That is enough to avoid to move the nozzle to Z height provided by "level_z_pos". In a next PR I will make it configurable in Configuration.h

Can't you just add an option to make it selectable in config.ini? As you know Marlin has that option already configurable for manual mbl. Or just seperate the corner leveling as a different line item in config.ini. So I can adjust the corner level z height and mbl z height seperately. I suppose that would also solve the issue.

digant73:
I was adding a conf. param on Configuration.h just to disable level_z_pos on MBL process. It's difficult it will be accepted on config.ini. Consider that for the just merged custom LED menu, no parameter was added on config.ini to allow to define a custom RGB LED color

lightmaster:
@digant73 Alright, finally got some time to sit down and recreate a new mesh and see if your instructions in https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1785#issuecomment-824947328 work to fix this issue. Will update after that's all done.

sarvenn:
> @sarvenn if you are able to make few changes in the source code and compile the firmware you could simply comment out in the file MBL.c the following lines:
> 
> line 20: //probeHeightStart(infoSettings.level_z_pos, false); // raise nozzle
> line 243: //probeHeightStart(infoSettings.level_z_pos, false); // raise nozzle
> 
> That is enough to avoid to move the nozzle to Z height provided by "level_z_pos". In a next PR I will make it configurable in Configuration.h

I've tried this. It seems to be working just like in the past. It's ok. Compiling my own fw is not a big deal. It would be better to set this in configuration.h or config.ini

By the way please check the link that I've sent from the bug report topic. There are news from marlin side for mbl z height which might need some changes at tft side. I don't know but you may catch if anything needed or not. @digant73

PR #1796, opened 2021-04-05
 tweak numPadFloat &  restore oldProgress to local variable, save 4byte ram

Msq001:
### Description

*  tweak numPadFloat. 
-- Reset key  add sound
-- Also reset the value when click ‘- ‘ and ’.‘ key first time to avoid too many 'del' key required
* restore `oldProgress `to local variable, save 4byte ram


TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -212,15 +211,12 @@ static inline void reDrawProgress(int icon_pos)
   GUI_DispString(printinfo_points[icon_pos].x + PICON_TITLE_X, printinfo_points[icon_pos].y + PICON_TITLE_Y, (uint8_t *)buf);
   GUI_SetTextMode(GUI_TEXTMODE_NORMAL);
 
-  if (oldProgress != newProgress)  // Update the progress bar only when the progress changes
+  if (prevProgress != newProgress)  // Update the progress bar only when the progress changes
```
kisslorand, line 214: No need for this conditional because `reDrawProgress()` is not called until `updatePrintProgress()` returns `true`. It returns `true` only when the progress advances.

bigtreetech: Got it

PR #1797, opened 2021-04-05
 separate y-axis invert for manual leveling menu

Msq001:
### Requirements

### Description
* Delete useless judgments in `reDrawProgress` for clean
* Separate y-axis invert for manual leveling menu
* clean up `infoPrinting` struct

### DIFF
--- a/Copy to SD Card root directory to update/config.ini
+++ b/Copy to SD Card root directory to update/config.ini
@@ -133,11 +133,16 @@ rotate_ui:0
 #   Options: [show: 1, hide: 0]
 terminal_ack:0
 
-#### Inverted Axes For Move Menu
-# Invert axis button function in Move menu in order it matches the actual axis movement.
-#   Format: [X<option> Y<option> Z<option>]
+#### Inverted Axes For Move Menu And Manual Leveling Menu
+# Invert axis button function in Move menu & Manual Leveling menu in order it matches the actual axis movement.
+#
+# NOTE: The Y-axis of different printer (move hotbed or move nozzle) move in different directions
+#       So Y-axis leveling invert can't follow up invert_axis[Y_AXIS]
+#       We separate a single variable "LY"(Leveling Y-axis) to deal with the y-axis leveling movement direction
+#
+#   Format: [X<option> Y<option> Z<option> LY<option>]
 #   Options: [enable: 1, disable: 0]
-invert_axis:X0 Y0 Z0
+invert_axis:X0 Y0 Z0 LY:0
 
 #### Persistent Temperature Info
 # Show persistent temperature info on all menus.
@@ -211,7 +216,7 @@ marlin_show_title:1
 
 #### Marlin Mode Title
 # Text displayed at the top of the TFT in Marlin Mode.
-#   Value range: [min: 3, max: 75 characthers]
+#   Value range: [min: 3, max: 20 characthers]
 marlin_title:Marlin Mode
 
 #### Marlin Mode Type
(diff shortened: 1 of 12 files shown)


digant73:
@bigtreetech please in setPrintAbort() in Printing.h add the following comment above infoHostprinting = false:

void setPrintAbort(void)
{
  // always reset the flag reporting the host is printing (even if the TFT didn't intercept it yet)
  // due to no further notifications will be sent by the host to notify it is no more printing
  infoHost.printing = false;

  // if printing from onboard SD or remote host
  if (infoPrinting.printing && infoFile.source >= BOARD_SD)

oldman4U:
Hi.

Unfortunately this does not fix the problem.

With Y and LY set to 0 the 5 point levelling is OK, but the Y axis is inverted while Move. So I  set Y1 and LY0, because the is OK. Unfortunately this inverted the axis for both, so the Move was OK, but for the 5 point levelling the y axis was inverted. So this means the Y value not only affects the Move but also the 5 point levelling. Good that we have the ability to change this also (even I believe changing the Y should not change the levelling also), so I defined Y1 (to get the correct Move menu) and LY1 (to invert the wrong invert of LY by defining Y1). Unfortunately this does not change anything, so it seems that LY is ignored as soon an Y is set to 1. So now I have 4 options and not a single one gives me the correct result.

Y0 and LY0: Move wrong, 5 point OK
Y1 and LY0: Move OK, 5 point wrong
Y0 and LY1: Move wrong, 5 point OK
Y1 and LY1: Move OK, 5 point wrong

From here we have 2 options:
1: LY ignores whatever is defined as Y.
2: LY allows to invert what ever the result of Y is, does not matter if it is 1 or 0.

bigtreetech:
> Hi.
> 
> Unfortunately this does not fix the problem.
> 
> With Y and LY set to 0 the 5 point levelling is OK, but the Y axis is inverted while Move. So I set Y1 and LY0, because the is OK. Unfortunately this inverted the axis for both, so the Move was OK, but for the 5 point levelling the y axis was inverted. So this means the Y value not only affects the Move but also the 5 point levelling. Good that we have the ability to change this also (even I believe changing the Y should not change the levelling also), so I defined Y1 (to get the correct Move menu) and LY1 (to invert the wrong invert of LY by defining Y1). Unfortunately this does not change anything, so it seems that LY is ignored as soon an Y is set to 1. So now I have 4 options and not a single one gives me the correct result.
> 
> Y0 and LY0: Move wrong, 5 point OK
> Y1 and LY0: Move OK, 5 point wrong
> Y0 and LY1: Move wrong, 5 point OK
> Y1 and LY1: Move OK, 5 point wrong
> 
> From here we have 2 options:
> 1: LY ignores whatever is defined as Y.
> 2: LY allows to invert what ever the result of Y is, does not matter if it is 1 or 0.

Hello, do you have the newest .bin file compiled by the new source code ? After updating the new .bin, the following settings can make your printer move as expected
`invert_axis:X0 Y1 Z0 LY0`

oldman4U:
I used your branch. Precompiled fw and config.ini with LY in it.

oldman4U:
Ok. Will test again using Master and let you know asap.

bigtreetech:
> Ok. Will test again using Master and let you know asap.

https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/tree/master/Copy%20to%20SD%20Card%20root%20directory%20to%20update
This should work well.

oldman4U:
YES!!!!!

Now it works GREAT. Y affects ONLY Move and LY affects ONLY 5 point levelling.

THANK YOU

PR #1819, opened 2021-04-09
all theme icon filenames compliant to same naming convention

digant73:
**CLEANUP:**
* just a rename compliant to the same naming convention (based on use of "_" as word separator) of all theme icon filenames.
  All the icon filenames should be compliant to the standard naming convention:
  1) underscore character "_" used as word separator
  2) filename in lowercase

  E.g.  "mesh_valid.bmp" instead of "Mesh_Valid.bmp"

  **NOTE:** Removed duplicated "Mesh_Valid.bmp" and "levelcorner.bmp" from THEME_Hybrid Red Menu Material theme. They were also present with the correct naming "mesh_valid.bmp" and "level_corner.bmp".
* cleanup on MKS binary generation. Same naming convention for BTT binaries is also used for all MKS variants. This allows to avoid to overwrite binary files having the same filename (e.g. mkstft28.bin v3.0 with mkstft28.bin v4.0)
* reported Marlin G26_MESH_VALIDATION setting on config.ini to support the new added Mesh Validation menu
* minor code optimization on ZOffset and MeshTuner menu
* Selectmenu and Parametersetting renamed to SelectMenu and ParameterSettings respectively


thisiskeithb:
How are the MKS binaries named correctly after this change? They can’t be updated using the `firmware.bin` file name convention.

digant73:
@thisiskeithb the same naming convention used for BTT fw is used also for MKS. So the MKS_TFT28_V3.0.bin is produced and must be renamed to MKSTFT28.bin. This is the current possible solution with the current folder structure just to avoid fw with same bin filename are overridden. I was thinking to propose in the next future a different strucure for folder "Copy to SD Card root directory to update" providing a first folder for the fw version (e.g. 0.26, 0.27 etc...). In each fw directory, the TFT variant (e.g. MKS_TFT28_V3.0). In each TFT variant directory, we provide the bin file (e.g. mkstft28.bin).
That folders structure will allow a history of the fw version (it could be extended also to build version if they will be used in the future) and also the possibility to have binaries with the same filename (e.g. mkstft32.bin (for v1.3), mkstft32.bin (for v1.4), mkstft28 (for v3.0), mkstft28 (for v4.0) etc...)

PR #1820, opened 2021-04-09
Cleanup Level Corner menu

guruathwal:
- Cleanup Level Corner menu
- misc clean-up

Build Report:

Device | Old RAM Usage | New RAM usage | RAM Reduced | Old Flash Size | New Flash Size | Flash Size Changed
-- | -- | -- | -- | -- | -- | --
TFT35 V1.0 | 17852 | 17852 | 0 | 151652 | 151104 | -548
TFT35 V1.1 | 17852 | 17852 | 0 | 151652 | 151104 | -548
TFT35 V1.2 | 17852 | 17852 | 0 | 151524 | 150976 | -548
TFT35 V2 | 18620 | 18620 | 0 | 161064 | 160516 | -548
TFT35 V3.0 | 19048 | 19048 | 0 | 179440 | 178892 | -548
TFT35 E3 V3.0 | 19072 | 19072 | 0 | 180128 | 179580 | -548
TFT35 B1 V3.0 | 19072 | 19072 | 0 | 180136 | 179588 | -548
TFT43 V3.0 | 19072 | 19072 | 0 | 180114 | 179596 | -518
TFT50 V3.0 | 19072 | 19072 | 0 | 180144 | 179596 | -548
TFT70 V3.0 | 19200 | 19200 | 0 | 181320 | 180772 | -548
TFT28 V1.0 | 17852 | 17852 | 0 | 151668 | 151128 | -540
TFT28 V3.0 | 19072 | 19072 | 0 | 179952 | 179412 | -540
TFT24 V1.1 | 19000 | 19000 | 0 | 175352 | 174804 | -548
MKS TFT32 V1.3 | 18960 | 18960 | 0 | 174892 | 174344 | -548
MKS TFT32 V1.4 | 18960 | 18960 | 0 | 175332 | 174784 | -548
MKS TFT28 V3.0 | 18960 | 18960 | 0 | 174892 | 174344 | -548
MKS TFT28 V4.0 | 18960 | 18960 | 0 | 175332 | 174784 | -548


digant73:
@guruathwal as it was done on Probe Offset and Mesh Editor recently, BL should be disabled otherwise meshes are applied to the target point falsing the results.
Please see at ZOffset.c at lines 117-119 and 299-301 for the code that could be integrated also in LevelCorner.c

guruathwal:
@digant73 I believe the level corner is meant to adjust the bed corners manually by using the z probe to measure distance. With BL enabled the corners will get aligned to the mesh and with BL disabled the corners will get aligned physically which will require re-editing the mesh. Am I correct?

digant73:
@guruathwal I think the idea of the developers was to align the corners physically (or at least this is what I undestood considering the meaning of such kind of feature). But beeing not sure 100% it's better to leave it as it is now.
I also tested the menu (quickly) and the results I got are not so clear to me. I got different results without using the knobs.

Cal1sto:
@guruathwal The process order to prepare the bed for printing is similar than the manual leveling with paper, Leveling -> Z-Offset -> Mesh . I will check your change proposal.

Cal1sto:
@digant73 About your test result, all probe are mostly accurate at +-0.01 for the best one, so values variation has this ground reason, other reason i had is when you have a not stable Y axis this add also some difference, with help of this results i could identify an issue on my Y axis and could solve it, to have only interference from probe accurency.

Cal1sto:
@guruathwal I just checked your change is not working, level corner menu freeze the tft. Will say it is better to leave it as it is now like say @digant73 , you had some good idea but you remove and replace some part without seen the interaction.

Cal1sto (changes requested):
This PR is not working, Level corner menu is freezing the tft

digant73:
@Cal1sto I meant that I was not sure at 100% about the logic of your PR and so I asked for a clarification from your side just to avoid confusion. I also tested the current fw not this PR

guruathwal:
@Cal1sto what @digant73 is proposing was to disable Bed Leveling before entering this menu  and that is was he suggested to leave as it is  because it was not sure what the purpose of the level corner was. 
There are no changes in the functionality in this PR.I removed some unnecessary code and procedures which are not needed. Can you elaborate at what point is it freezes? So it could be fixed.

Cal1sto:
The BL like you must know is from the G28 disabled for taking the right and real Z point. So nothing there to be performed. For the tft soon you open with your change the 'L corner' the screen is blocked and not working anymore.

digant73:
@Cal1sto. I think the freeze will be quickly fixed by guruatwhal. Just a clarification on BL ON or OFF. My Marlin fw is configured with RESTORE_LEVELING_AFTER_G28, so after G28, BL is normally switched to ON (BL state was ON before G28). My doubt is, do we need to always force BL to off before using your feature or it is not required?

digant73:
@guruathwal just a cleanup on config.ini in your previous PR (Terminal).
Please apply the attached files (fixed indentation on last added setting)

[config.ini.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/6295974/config.ini.zip)

Cal1sto:
@digant73 No the G30 used to to read the Z position doesn't trigger le mesh is like the G28 it give back the Z height. So ne need there.

guruathwal:
@Cal1sto Please check if the problem is fixed.
@digant73 This feature should not be placed in ABL menu. It is independent of ABL and is a variant of the already existing Manual leveling menu. The same goes for BLTouch menu.

digant73:
@guruathwal but it requires also a probe (Cap:Z_PROBE), so it belongs to Bed Leveling menu.
The Bed Leveing menu hosts all the basic features common to any BL type (ABL and MBL) (even if no probe is available). e.g. Probe Offset, Z fade, Mesh Edit etc...
The Beb Leveling Layer2 provides all the other features available for the target BL type. So if a probe is supported then BLTouch and Level Corner menus are displayed etc...

guruathwal:
@digant73 `Cap:Z_PROBE` is independent of ABL and can exist even if bed leveling is completely disabled in marlin firmware.

digant73:
@guruathwal but in that case we should add another menu at same level of Manual and Bed Leveling. But I do not see space for that in the main menu

guruathwal:
@digant73  Adding another menu for a same feature will not be good idea. It will be more confusing. An option to choose between Manual leveling and probe assisted leveling (Leveling corner) should be added in config so that only one option is visible.

digant73:
personally I would like to always maintain the manual leveling menu. It is a basic feature for me. The Level Corner is a plus if someone wants to fine tune. Maybe the Tuning menu could be a possible container for that

Cal1sto:
@digant73 I agree with you for the location

Cal1sto:
@guruathwal You made a lot of more memory usage than needed for this little function,  so I don't really understand why you did this, only positive part and good idea was to split of the icon draw and label draw, but the big modification you did in levelcorner and the logic is covered with more data usage than needed, not sure this could be considered as an improvement. I let this merge decision and analysis to repository responsible person or team, please don't ask to test if you didn't do it.

guruathwal:
@Cal1sto There is no extra memory usage involved. The code is changed to reduce flash memory usage and CPU cycles. More than  500 Bytes of fash size is reduced in this PR. Do you think it is wise to let program memory get filled up for the sake of not touching any code once it was merged in the repository? I tested this PR but it is always better if more people could test to catch any bug that one person could not.

Cal1sto:
@guruathwal Remove if(!coordinateIsKnown()) in levelcorner.c this is buggy , when Marlin goes in timeout this function is not accurate and could generate move outside bed limit, I had it in my PR on first comit and i remove it, already in the merged running solution, then not sure your amount of change will keep the value when you go out of this menu and came back. Not yet tested if this change still running like it was.

bigtreetech:
@Cal1sto Hello, thank you very much for your contribution. Generally speaking, it is very meaningful to reduce the memory usage. Because the hardware resource of MCU is limited, adding function will consume the hardware resource. Until one day, the hardware is completely exhausted, and then we have to switch functions through `#define` like Marlin instead of `config.ini` file. Users have to recompile the code to modify the settings. This is not friendly to users.

So I really appreciate your work, I'm sorry Maybe this PR has brought some new problems, which makes you tired. But I hope you can help us test whether this PR works properly, because it is necessary to reduce hardware consumption.
I will merge this PR after your test works.

Is there anything else besides `Remove if(!coordinateIsKnown()) in levelcorner.c` ?

Cal1sto:
@guruathwal PR tested on a MKS TFT28 it look good, please remove in levelcorner.c the if(!coordinateIsKnown())

sarvenn:
@guruathwal , @bigtreetech hi,
In the past this setting only applied to corner leveling. But after this PR this Z value also effects the Z height initial position while doing manual mesh bed leveling. This causes an issue.
#   Format: [level_z_pos: Z<position in mm>]
#   Unit: [distance in mm]
level_z_pos:0.2

New logic overrides my marlin config setting. (configuration.h)
 * The "Manual Probe" provides a means to do "Auto" Bed Leveling without a probe.
 * Use G29 repeatedly, adjusting the Z height at each point with movement commands
 * or (with LCD_BED_LEVELING) the LCD controller.
 */
#define PROBE_MANUALLY
#define MANUAL_PROBE_START_Z 0.2

I prefer to do corner leveling while my nozzle is at Z0. I don't use a paper. I set my nozzle to bed distance at 0 at corner leveling.
Then I do the manual mesh bed leveling. I expect the nozzle to stay at Z0.2 position and adjust my nozzle to slightly touching to the bed at each MBL points.
With this method I almost do not need to adjust a nozzle offset value at all (may be +/-0.01-0.02). I got perfect first layers.
But after this PR, I set this parameter to level_z_pos:0, corner leveing is OK, but at manual MBL, nozzle first stops at Z0.2 for a very small time, then goes to Z0 which causes my nozzle to hit my bed( this is risky, I already damaged my bed due to this yesterday). This is an unwanted behavior.

Can we go back to previous logic please? I may beg for it :) Please exclude mesh leveling Z position from this logic. It can be set in marlin config, or just seperate corner leveling and mesh leveling initial Z positions as different parameters. Please, please, please. 

## Z Position (Manual, Mesh Leveling, Probe/Home Offset and Mesh Tuner)
# For manual leveling and MBL, lower z axis to this absolute position after reaching a leveling point.
# For Probe/Home Offset and ABL in Mesh Tuner, raise z axis by this relative position after reaching a leveling point.
#   Unit: [position in mm]
level_z_pos:0.2

guruathwal:
@sarvenn  This PR does not change anything related to Z Position (Manual Leveling).

sarvenn:
Oh sorry for that. I supposed so. I'm a newbie in Github.
I found that config file has changed at 21fd9c0667b8a824c47beea68afb996c8edefaac.
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/21fd9c0667b8a824c47beea68afb996c8edefaac#diff-195af4469f629def2536697b843ae11c1e30e485a0afa8b1f5eb0fcbcfc8b3f4

But I could not find which PR has changed the logic.
Maybe this one.
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/7b40fec9263200232d7c1363002bc993805384b4#diff-3783eec427840b96dd27a626071fb1cabc30b3b41578b713fe87e5690d23185a

@digant73 hello,
Could you look up the situation above.

And please understand and forgive me if I'm writing this at wrong subject. :/ 
But I need help from developers to get this issue solved.

digant73:
@savern the use of "level_z_pos" was added because asked by other people. it works properly if you didn't also set it on Marlin. I can remove the use of "level_z_pos" but then many other people will complain of that the nozzle is moved to high from the bed (as it was already reported). "level_z_pos" should be used in all the kind of leveling instead of using 0 for that, 0.2 for this etc...

PR #1828, opened 2021-04-11
Add Terminal text color scheme

guruathwal:
- Add color scheme option for the Terminal menu. Any of the 3 color styles (Material Dark, Material Light, High Contrast) can be set in the config file.
- Separate colors for gcode commands and printer/machine response.
- Rename SendGcode.h to Terminal.h


TFT/src/User/Menu/SendGcode.c
```diff
@@ -743,16 +788,21 @@ void menuTerminalWindow(void)
...
 
       terminalDrawPageNumber();
+      pageHeadSrc = getLastSrc(terminalData->ptr[pageBufIndex]);
     }
 
     while (terminalBuf + lastTerminalIndex && (lastTerminalIndex != terminalData->terminalBufTail))
     {
       getCharacterInfo((uint8_t *)(terminalBuf + lastTerminalIndex), &info);
 
+      //read source identifier
```
digant73, line 802: a space after "//" and "if".
Furthermore, the new settings should be provided also in config_rrf.ini.
Maybe SendGcode.c could be now renamed as Terminal.c (matching the menu label and the text repored in config.ini and Configuration.h)

digant73 (changes requested):

TFT/src/User/Menu/SendGcode.c
guruathwal: I agree

digant73:
I would rename also Parametersetting to Parameter**S**etting**s** just to match with the other setting menus

PR #1851, opened 2021-04-18
New custom LED menu + some code optimization

### LINKED ISSUE #1687: [FR] Make the LED/Lights menu in TFT mode more feature complete
Can not set a Custom Light using the TFT Touch Screen mode without resorting to using terminal and `M150` command. If the average user that doesn't know much about gcode commands wants to change the LEDs to something besides Red, Green, Blue, and White, then they have to switch to Marlin mode and can then have full access to Marlin's much more fleshed out Lights menu. Furthermore, I propose expanding the preset colors to also include the other primary and secondary colors used by Marlin (ROY G BIV colors).

Kinda feel that's pretty self explanatory from above, but I'd like the TFT Touch Screen LED menu to be more fleshed out and have the options that Marlin Mode has.

Encourage anyone wanting to play around with the LED colors to use Marlin Mode, or continue to require them to understand GCode and RGB color codes

**Additional context**

digant73:
**NEW FEATURES:**
* added custom LED menu:
 1) it is now possible to set all the parameters for M150 gcode (machine RGB LED)
 2) the set LED color is previewed at runtime. Any change is scheduled to be applied every 1 second at maximun, just to avoid useless traffic (to send M150 gcode) every time a component is changed
 3) RGB component colors and RGB color are also displayed just as color pickup. Obviously RGB color on display is limited to RGB 565 format while machine LED uses the wider range 0-255 for each component
 4) the LED main menu allows to set and save standard/custom color and switch to on/off the saved color
 5) rotary encoder can be used to change the value for the control holding the focus

**IMPROVEMENTS:**
* some code optimization on Terminal, Mesh editor and LED menus. They now reuse some common code
* on terminal menu, the page box is no more clickable now

resolves #1687

IMG_20210418_210420

IMG_20210418_201621

IMG_20210418_201653


TFT/src/User/Menu/LED.c
```diff
@@ -0,0 +1,575 @@
...
+    (*led)[i] = ledValue[i];
+}
+
+void ledSetValue(const LED_VECT * led)
+{
+  for (int i = 0; i < LED_VECT_SIZE; i++)
+    ledValue[i] = (*led)[i];
+}
+
+uint8_t ledGetComponetIndex(uint8_t index)
```
guruathwal, line 176:
```suggestion
uint8_t ledGetComponentIndex(uint8_t index)
```

TFT/src/User/Menu/LED.c
```diff
@@ -0,0 +1,575 @@
...
+uint16_t ledGetRGBColor(LED_VECT * led)
+{
+  return (((*led)[0] & 0x001F) << 11) | (((*led)[1] & 0x003F) << 5) | (((*led)[2] & 0x001F));  // RGB color in RGB 565 16 bit format
+}
+
+uint16_t ledGetComponentRGBColor(uint8_t component, uint8_t index)
+{
+  LED_VECT led = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00};  // component RGB color in RGB 565 16 bit format
+
+  led[index] = component;
```
guruathwal, line 219: is it necessary to add space between just two lines?

guruathwal (changes requested):

guruathwal:
Instead of using the arrow button to change values, it will be much better to use the numpad to change the values.

TFT/src/User/Menu/LED.c
digant73: yes, thanks

TFT/src/User/Menu/LED.c
digant73: usually I always separate variable definition

digant73:
> Instead of using the arrow button to change values, it will be much better to use the Numpad to change the values.
> This PR is using about 1770 bytes of additional flash. Please try to reduce it if possible.

sure, there is also the possibility to use the Numpad (clicking in the value area). I already reduced the code at minimun (sharing also some pieces of code with Terminal and Mesh editor). The remaining logic/checks are needed to preserve consistency of the menu

digant73:
some problems with editing conflicting files. closing the PR and reopen it

PR #1853, opened 2021-04-19
New LED Color menu + some code optimization

### LINKED ISSUE #1687: [FR] Make the LED/Lights menu in TFT mode more feature complete

digant73:
**NEW FEATURES:**
* added custom LED menu:
 1) it is now possible to set all the parameters for M150 gcode (machine RGB LED)
 2) the set LED color is previewed at runtime. Any change is scheduled to be applied every 1 second at maximun, just to avoid useless traffic (to send M150 gcode) every time a component is changed
 3) RGB component colors and RGB color are also displayed just as color pickup. Obviously RGB color on display is limited to RGB 565 format while machine LED uses the wider range 0-255 for each component
 4) the LED main menu allows to set and save standard/custom color and switch to on/off the saved color
 5) rotary encoder can be used to change the value for the control holding the focus

**IMPROVEMENTS:**
* some code optimization on Terminal, Mesh editor and LED menus. They now reuse some common code
* on terminal menu, the page box is no more clickable now

resolves #1687

IMG_20210418_210420

IMG_20210418_201621

IMG_20210418_201653


Guilouz:
Hi, I have a warning with your PR : 

Capture d’écran 2021-04-19 à 15 04 11

Capture d’écran 2021-04-19 à 15 04 24

Guilouz:
And just try your PR, not working for me. I have Neopixels connected to my SKR 1.4 Turbo, at startup I have the color cycle and then the switching on in white (so far everything is fine) but when I go in LED menu, I can't change anything, buttons blue, green, red, white, on and off doesn't work anymore and shutdown my Neopixels. Custom color have no effect no more.

digant73:
The warning is not a problem (the define on LEDColor overrides the one provided in another .h file).
The standard colors (red, green etc...) in the LED Color menu were maintained compatible with the previous ones (so P and I for M150 are set to 0). From the custom menu you have the possibility to set also W, P and I

Guilouz:
Ah we need to set W, P and I in custom menu to use red, green, blue and white button ? Not really intuitive.

digant73:
> Ah we need to set W, P and I in custom menu to use red, green, blue and white button ? Not really intuitive.

currently pressing on RED, BLUE etc... will always set to 0 P and I. It could be easily fixed. Use the custom menu setting the proper value (also I) and it should work

Guilouz:
> > Ah we need to set W, P and I in custom menu to use red, green, blue and white button ? Not really intuitive.
> 
> currently pressing on RED, BLUE etc... will always set to 0 P and I. It could be easily fixed. Use the custom menu setting the proper value (also I) and it should work

Just try, when i set W, P and I my leds light up, but after when I click on red, blue or green button my leds light off instead of changing color and value W, P and I go back to 0.

digant73:
> > > Ah we need to set W, P and I in custom menu to use red, green, blue and white button ? Not really intuitive.
> > 
> > 
> > currently pressing on RED, BLUE etc... will always set to 0 P and I. It could be easily fixed. Use the custom menu setting the proper value (also I) and it should work
> 
> Just try, when i set W, P and I my leds light up, but after when I click on red, blue or green button my leds light off instead of changing color and value P and I go back to 0.

yes that is the bug. an easy fix. but you will need to use the custom menu to set the proper NEO pixel setting. Then P and I will be maintained pressing on RED, BLUE... ON and OFF

Guilouz:
@digant73 Let me know as soon as it's done so I can test.

digant73:
> @digant73 Let me know as soon as it's done so I can test.

fixed

Guilouz:
@digant73 This issue it's now fixed but another issue : P and I components are not saved after restarting the printer, it's really annoying if you have to redefine them every time.

Maybe it will be better if they are set to 255 by default.

Guilouz:
@GregSKR Yes when I (index) is set to 0 only the first led is controlled, if you set 1 the second, etc.. depending of number of LED you have.

Personally, I don't really see the point of this parameter I, I don't change the color of my LEDs one by one but all at the same time.

digant73:
ok so it is enough to use 255 as default value for P and I instead of 0.
The LED color is not saved to flash. You always need to select it after a reboot

Guilouz:
@digant73 Just try your fix and value for P and I are always 0 at startup. Not working.

digant73:
should work now

Guilouz:
@digant73 Just try your last fix, all is working as expected now. Thanks.

GregSKR:
@digant73 Thanks, I confirm it's all fine now for NeoPixels, everything works as expected! 👍

digant73:
good. thanks to both for the testing

PR #1864, opened 2021-04-21
Notifikation M117

### LINKED ISSUE #1833: [Q] What should be shown on the Notification screen?
I use M117 notifications from the Octoprint-DetailedProgress plugin to push progress info to my TFT35 screen. This works fine in recent releases. I was also expecting to find a log of these messages in the 'Notifications' screen after clicking this area on the status screen. However this screen is always empty for me.

Is this a configurable option I can enable somewhere, or a bug or maybe by design?

radek8:
Display of M117 messages on the notification screen.
With the option to turn this view on or off in the CONFIG.INI file
Fix M569 error

resolves #1833 

image


Copy to SD Card root directory to update/config.ini
```diff
@@ -173,6 +173,11 @@ files_list_mode:1
 #     TOAST: A non-blocking Toast notification is displayed for few seconds. No user interaction is needed.
 ack_notification:1
 
+#### Notification M117
+#Display of M117 messages on the notification screen
```
digant73, line 177: leave a space after # and put a dot at the end. So "# Display..... screen."

digant73 (changes requested):

Copy to SD Card root directory to update/config.ini
radek8: Thank you

Tannoo:
A question: If `#define NOTIFICATION_M117 0` is disabled, why is it then commented out by default also?

and will `# define` work just as well as `#define`?

`//# define NOTIFICATION_M117 0  // Default 0`

radek8:
> In `config.c`, don't you want a `break;` for the M117 case?

Fixed
Thank you

PR #1885, opened 2021-04-27
Fix PLR bug

Msq001:
### Requirements

### Description
* fix PLR Busy processing bug
* ~fix PLR active all heater include invaild T~

### DIFF
--- a/TFT/src/User/API/Temperature.c
+++ b/TFT/src/User/API/Temperature.c
@@ -15,17 +15,30 @@ static bool    heat_send_waiting[MAX_HEATER_COUNT];
 uint32_t nextHeatCheckTime = 0;
 #define AUTOREPORT_TIMEOUT (nextHeatCheckTime + 3000)  // update interval + 3 second grace period
 
-static uint8_t fixHeaterIndex(uint8_t index)
+// Verify that the heater index is valid, and fix the index of multiple in and 1 out tool nozzles
+static uint8_t heaterIndexFix(uint8_t index)
 {
-  if (infoSettings.hotend_count == 1)
-    index = (index < MAX_HOTEND_COUNT) ? NOZZLE0 : index;
-  return index;
+  if (index == BED && infoSettings.bed_en)  // Bed
+    return index;
+  if (index == CHAMBER && infoSettings.chamber_en)  // Chamber
+    return index;
+  if (index < infoSettings.hotend_count)  // Vaild tool nozzle
+    return index;
+  if (index < infoSettings.ext_count && infoSettings.hotend_count == 1)  // "multi-extruder" that shares a single nozzle.
+  {
+    return NOZZLE0;
+  }
+
+  return INVALID_HEATER;  // Invalid heater
 }
 
-//Set target temperature
+// Set target temperature
 void heatSetTargetTemp(uint8_t index, int16_t temp)
 {
-  index = fixHeaterIndex(index);
+  index = heaterIndexFix(index);
+  if (index == INVALID_HEATER)
+    return;
+
   heater.T[index].target = NOBEYOND(0, temp, infoSettings.max_temp[index]);
   if (heater.T[index].target + 2 > heater.T[index].current)
   {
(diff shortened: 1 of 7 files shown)


TFT/src/User/API/Temperature.c
```diff
@@ -15,17 +15,25 @@ static bool    heat_send_waiting[MAX_HEATER_COUNT];
 uint32_t nextHeatCheckTime = 0;
 #define AUTOREPORT_TIMEOUT (nextHeatCheckTime + 3000)  // update interval + 3 second grace period
 
-static uint8_t fixHeaterIndex(uint8_t index)
+// Verify the index is invalid
+static bool heaterIndexInvalid(uint8_t index)
 {
-  if (infoSettings.hotend_count == 1)
-    index = (index < MAX_HOTEND_COUNT) ? NOZZLE0 : index;
-  return index;
```
guruathwal, line 22: The reason for this check was to convert the index to `NOZZLE0` when the selected hotend index is greater than `NOZLE0` but less than the maximum possible hotend count. 
This was done to ensure the correct hotend index is selected in case of a mixing hotend which have multiple extruders with a single hotend because all the menu related to the extruder pass the extruder index to the temperature API for queries.
The logic only works when the hotend count is 1 and the selected hotend index is not `0`. It does not affect the heated bed index or the chamber heater index. If the hotend count is greater than 1, then there is no effect.

TFT/src/User/API/Temperature.c
```diff
@@ -68,7 +82,9 @@ void heatSetCurrentTemp(uint8_t index, int16_t temp)
 // Get current temperature
 int16_t heatGetCurrentTemp(uint8_t index)
 {
-  index = fixHeaterIndex(index);
+  if (heaterIndexInvalid(index))
+    return 0;
+
```
guruathwal, line 87: The Extruder menu needs temperature status for a specific extruder instead of hotend. for example, a dual hotend setup will have the same hotend index as the extruders but a mixing hotend with 3 extruders will have the same hotend index (`NOZZLE0`) for all extruders.
With the new logic, the hotend temperature for all extruders except `E0` for mixing hotend setup will always return 0. This will trigger false cold extrusion warnings for all extruders except `E0`.

TFT/src/User/Menu/PowerFailed.c
```diff
@@ -125,21 +125,22 @@ bool powerOffGetData(void)
...
   {
-    if (infoBreakPoint.fan[i] != 0 && fanIsType(i, FAN_TYPE_F))
+    if (infoBreakPoint.fan[i] != 0)
     {
       mustStoreCacheCmd("%s S%d\n", fanCmd[i], infoBreakPoint.fan[i]);
     }
   }
 
   mustStoreCacheCmd("%s\n", tool_change[infoBreakPoint.tool]);
+  for (int8_t i = MAX_HEATER_COUNT - 1; i >= 0; i--)  // infoCacheCmd.queue[0 - 7]
```
guruathwal, line 138:
```suggestion
  for (int8_t i = infoSettings.hotend_count - 1; i >= 0; i--)  // infoCacheCmd.queue[0 - 7]
```

The cause for the Invalid extruder error is that the PLR API is sending temperature command for all hotends (`MAX_HEATER_COUNT`), even if there is only one hotend available. The current logic is sending the commands directly to the cmd cache without checking the actual hotend count reported by the machine. Replacing `MAX_HEATER_COUNT` with `infoSettings.hotend_count` should fix the issue.

guruathwal (changes requested):

TFT/src/User/Menu/PowerFailed.c
```diff
@@ -125,21 +125,22 @@ bool powerOffGetData(void)
...
+  {
+    if (infoBreakPoint.target[i] != 0)
```
guruathwal, line 140:
```suggestion
    if (((i < MAX_HOTEND_COUNT) ? (i < infoSettings.hotend_count) : true) && infoBreakPoint.target[i] != 0)
```

TFT/src/User/API/Temperature.c
```diff
@@ -15,17 +15,30 @@ static bool    heat_send_waiting[MAX_HEATER_COUNT];
...
+  if (index == BED && infoSettings.bed_en)  // Bed
+    return index;
+  if (index == CHAMBER && infoSettings.chamber_en)  // Chamber
+    return index;
+  if (index < infoSettings.hotend_count)  // Vaild tool nozzle
+    return index;
+  if (index < infoSettings.ext_count && infoSettings.hotend_count == 1)  // "multi-extruder" that shares a single nozzle.
+  {
+    return NOZZLE0;
+  }
```
guruathwal, line 30: The logic will never reach to line number 27 (`if (index < infoSettings.ext_count && infoSettings.hotend_count == 1)`) because the index passed will always be less than infoSettings.hotend_count.

guruathwal:
The problem is in these lines (in PowerFailed.c):
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/47f1b7df7dc61f6a300c1bbd66400a4ce9e70203/TFT/src/User/Menu/PowerFailed.c#L59-L62
This is an incorrect implementation. It is trying to get the target temperature for all heaters including hotends that may not exist on the machine(T0 to T5), but it will receive status for `NOZZLE0`.

Below is an alternative to prevent this:
```C
    for (uint8_t i = 0; i < MAX_HEATER_COUNT; i++)
    {
      if (i >= MAX_HOTEND_COUNT || i < infoSettings.hotend_count)
        infoBreakPoint.target[i] = heatGetTargetTemp(i);
    }
```

bigtreetech:
@guruathwal Hello, please check https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1885/commits/e6553116a19bc36794b92a29c6baa7e029af3eb8

guruathwal:
@bigtreetech  yes this is ok.

PR #1894, opened 2021-05-02
Layer height consistent display, MCU load reduction during print, tiny RAM and Flash usage reduction

### LINKED ISSUE #1424: [BUG] Wrong layer on display
It seems, that after some updates (at least as of commit 9148ffe5708ab59e0b5590a4b0e67a44d6bc47d1), the current layer information is wrong and displays random values. E.g. on first layer it displays 0.6mm (instead of 0.2mm).
After pause/resume, it displayed 11.2mm instead of 1.2mm.

kisslorand:
### Requirements

BTT or MKS TFT

### Description

- consistent layer (nozzle height) display during print 
The TFT now shows correctly and in realtime the nozzle height. However, when printing from TFT SD or TFT USB, it shows the nozzle height based on the last sent gCode to Marlin, not the one that Marlin executes.

- code optimization (during print)
The progress percentage and elapsed time display procedure is slightly reworked, it's now a single procedure instead of two. The progress bar is updated only if the progress percentage changed.

- added beep when exiting from InfoScreen

- tiny RAM and Flash reduction

### Benefits

Everyone

### Related Issues

Fixes #1424


guruathwal:
There are too many merge commits. You need to reset your branch to the upstream branch or create a new branch before opening a PR.

kisslorand:
I don't know what I've done but it is now a real mess :(

kisslorand:
Cleaned up the mess I created. Someone must help me out how to get rid of that commit history.

rebel1:
> Cleaned up the mess I created. Someone must help me out how to get rid of that commit history.

The easiest thing is to create a new branch from master and copy your changes over and create a new PR.

kisslorand:
Pretty much that's how I do it. I create a branch from the master and than edit it. 
Before creating a branch I always sync the the master from my repository to the master from this repository.

bigtreetech:
Hello, Almost every gcode command of z-axis will change when gcode file is generated in Cura "vase" mode, In order to avoid printing defects caused by frequent redrawing icon taking up CPU, so limit the frequency of update reDrawLayer is necessary.

But the previous source code had a small bug: If the z-axis is not reDraw because it is limited by time, but the `curLayer` flag will still be modified, so that the z-axis will never be reDraw until the next z-change.
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/c37217f46d33423162584c7455ea5629e0ffc458 There will be no such problem, please review

kisslorand:
> Hello, Almost every gcode command of z-axis will change when gcode file is generated in Cura "vase" mode, In order to avoid printing defects caused by frequent redrawing icon taking up CPU, so limit the frequency of update reDrawLayer is necessary.


Oooops, I forgot about vase mode


> But the previous source code had a small bug: If the z-axis is not reDraw because it is limited by time, but the `curLayer` flag will still be modified, so that the z-axis will never be reDraw until the next z-change.
> [c37217f](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/c37217f46d33423162584c7455ea5629e0ffc458) There will be no such problem, please review

Yup, that's why I removed time dependency for reDraw.

I will review the code.

kisslorand:
@bigtreetech 
I took another approach, layer height is updated only if the delta is above a threshold.
This way there are less `if` conditionals, less calculations, fast reaction to layer change, no congestion on vase mode.
I tested vase mode on an 1 cm diameter cylinder at an insane speed (speed rate was at 500%) and the layer height update period was more than 500 ms.
The coded delta threshold (`LAYER_DELTA)`is 0.1mm.

PR #1888, opened 2021-04-28
added action:resumed and action:resume + code cleanup

digant73:
**IMPROVEMENTS:**
 * added "action:resumed" and "action:resume" on parseAck fixing a missing dialog for the M600 not properly working reported in #1829. Now, resuming a print, the "Resume" button is properly switched to "Pause" button even in case REPRAP_DISCOUNT_FULL_GRAPHIC_SMART_CONTROLLER or ADVANCED_PAUSE_FEATURE is not enabled in Marlin
* Code cleanup. Reviewed .h and .c files just to include only the needed dependency files (in particular avoiding to include general files in the .h files)
* added ENABLE_MBL_START_Z in Configuration.h. It enables/disables "level_z_pos" configurable parameter value as the starting Z height for each point during MBL process.
* minor cleanup

**FIXES:**
fixes #1889
fixes #1895

**PR STATE:** ready for merging

PR #1903, opened 2021-05-06
Fix config update failed after PR #1888, add serial communication debug option

guruathwal:
- Fix config update failing on some devices caused by missing header file inclusion after PR #1888 in some files. 
- Add serial communication debug option to enable real-time monitoring through the spare serial port on some TFT variants.
- Remove some duplicate code.
- Replace u32 with uint32_t.


oldman4U:
How does the serial com debug work?

guruathwal:
you will have to connect a USB to Serial convert on to UART3 port on the TFT if it has one.

TFT/src/User/API/ScreenShot.h
```diff
@@ -6,6 +6,7 @@ extern "C" {
 #endif
 
 #include <stdbool.h>
+#include "variants.h"
```
digant73, line 9: why did you include "variants.h" if there is no dependency here?

TFT/src/User/API/boot.h
```diff
@@ -7,6 +7,8 @@ extern "C" {
 
 #include <stdbool.h>
 #include <stdint.h>
+#include "variants.h"
+#include "config.h"
```
digant73, line 11: is that dependency needed?

TFT/src/User/API/config.h
```diff
@@ -7,6 +7,7 @@ extern "C" {
 
 #include <stdbool.h>
 #include "ff.h"
+#include "includes.h"
```
digant73, line 10: I don't see any dependency on "include.h" in config.h. "include.h" is already present in "config.c"

TFT/src/User/API/flashStore.h
```diff
@@ -7,6 +7,7 @@ extern "C" {
 
 #include <stdbool.h>
 #include <stdint.h>
+#include "variants.h"
```
digant73, line 10: I don't see any dependency on "variants.h" in flashstore.h. If it is needed in flashstore.c, please put it in the .c file

digant73 (changes requested):

oldman4U:
Ok. So not really for the average user like me. 😜

TFT/src/User/API/ScreenShot.h
guruathwal: because I restore these includes in one go. need to check each one.

oldman4U:
Could this be related: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/1904

kisslorand:
With all due respect, what's the purpose of this serial debug option?

guruathwal:
@oldman4U  for TFT70 it is.

@kisslorand  to monitor communication between the printer and TFT.

oldman4U:
Should I install your PR or should I install the current master before to see the problem first?

guruathwal:
to see the problem use the current master.

oldman4U:
I see the problem using the current master. Checking your PR now

oldman4U:
Confirm that it is fixed. Thank you

kisslorand:
> @kisslorand to monitor communication between the printer and TFT.

OK, reformulate, for regular users what's the purpose of the ability to monitoring communication between the printer and TFT?

kisslorand:
Why would a regular user enable it?

guruathwal:
@kisslorand I did not say a regular user should enable it.  Why will it consume program space when it is disabled by default? 
Did you saw the changes in this PR?

image

kisslorand:
Yes, saw it and I do not think any develooer needs this.

guruathwal:
well, I am a developer. It is disabled so it won't hurt any program space. if anyone needs it, then it can be enabled.

oldman4U:
Ordered two of those serial to usb readers.😁

PR #1907, opened 2021-05-08
Load/Unload/Extrude/Retract rework and bugfixes

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This is a rework and bugfix related to "Extrude" and "(Un)Load" menu.

- in these menus a nozzle preheat to minimal extrusion temperature is offered only if the target temperature is lower than that, otherwise just a warning will be displayed that the minimal temperature is not reached (it's the case when the nozzle is preheating above minimal extrusion temperature but it didn't reach yet the minimal/requested temperature)
- after the minimal temperature is reached and the temp falls behind a few degrees (due to PID and temp lag) the procedure will go on (load, unload, etc) but only if the temp is lower by no more than a certain/predefined offset (5 ºC)
- if the target temp is lower but the actual temp is above the minimal extrusion temperature (nozzle cooling) the target temperature will be the actual temp and the procedure will go on (load, unload, etc)
- "Extrusion" and "(Un)Load" menu will restore the E axis coordinate to the value before the menu was called

### Benefits

It improves the handling of temperatures and restores the E axis coordinate to the correct value when using the mentioned menus.

### Related Issues

- if printing from onboard SD and the user paused the print and used "Extrude" menu, the E axis coordinates is restored wrongly upon exiting the menu
- if printing from TFT USB or TFT SD and the user paused the print and used "(Un)Load" menu the E axis coordinate is not restored upon exiting the menu
- if using "(Un)Load" menu the E axis coordinate is not restored at all (loading filament and going from this menu to "Extrude" menu and attempting to load more filament, it actually retracted the filament due to not updating/restoring E axis coordinate)
- even if preheat was done in the mentioned menus, right after the preheat procedure, if the temperature felt back a few degrees (due to PID and temp lag) the Load/Unload/Extrude/Retract command is not executed, instead a warning about minimal extrusion temperature not being reached is shown and the option to preheat is offered again

### DIFF
--- a/TFT/src/User/Menu/common.c
+++ b/TFT/src/User/Menu/common.c
@@ -185,21 +185,41 @@ const void drawStandardValue(const GUI_RECT *rect, VALUE_TYPE valType, const voi
 
 const bool warmupTemperature(uint8_t toolIndex, void (* callback)(void))
 {
-  if (heatGetCurrentTemp(toolIndex) < infoSettings.min_ext_temp)
-  { // low temperature warning
-    char tempMsg[120];
-    LABELCHAR(tempStr, LABEL_EXT_TEMPLOW);
+  #define TEMP_OFFSET 5  // offset temperature to avoid denial of extrusion/retraction due to the nozzle temperature lag
+  if (heatGetTargetTemp(toolIndex) < infoSettings.min_ext_temp)
+  {
+    if (heatGetCurrentTemp(toolIndex) < infoSettings.min_ext_temp - TEMP_OFFSET)
+    {  // low temperature warning and heatup offering
+      char tempMsg[120];
+      LABELCHAR(tempStr, LABEL_EXT_TEMPLOW);
 
-    sprintf(tempMsg, tempStr, infoSettings.min_ext_temp);
-    strcat(tempMsg, "\n");
-    sprintf(tempStr, (char *) textSelect(LABEL_HEAT_HOTEND), infoSettings.min_ext_temp);
-    strcat(tempMsg, tempStr);
-
-    setDialogText(LABEL_WARNING, (uint8_t *) tempMsg, LABEL_CONFIRM, LABEL_CANCEL);
-    showDialog(DIALOG_TYPE_ERROR, callback, NULL, NULL);
-
-    return false;
+      sprintf(tempMsg, tempStr, infoSettings.min_ext_temp);
+      strcat(tempMsg, "\n");
+      sprintf(tempStr, (char *) textSelect(LABEL_HEAT_HOTEND), infoSettings.min_ext_temp);
+      strcat(tempMsg, tempStr);
+      
+      setDialogText(LABEL_WARNING, (uint8_t *) tempMsg, LABEL_CONFIRM, LABEL_CANCEL);
+      showDialog(DIALOG_TYPE_ERROR, callback, NULL, NULL);
+      return false;
+    }
+    else
+    { // contiunue with current temp but no lower than the minimum extruder temperature
+      heatSetTargetTemp(toolIndex, MAX(infoSettings.min_ext_temp, heatGetCurrentTemp(toolIndex)));
+    }
   }
+  else
+    if (heatGetCurrentTemp(toolIndex) < infoSettings.min_ext_temp - TEMP_OFFSET)
+    { // low temperature warning
+      char tempMsg[120];
+      LABELCHAR(tempStr, LABEL_EXT_TEMPLOW);
+
+      sprintf(tempMsg, tempStr, infoSettings.min_ext_temp);
+      strcat(tempMsg, "\n");
+      
+      setDialogText(LABEL_WARNING, (uint8_t *) tempMsg, LABEL_CONFIRM, LABEL_BACKGROUND);
+      showDialog(DIALOG_TYPE_ERROR, NULL, NULL, NULL);
+      return false;
+    }
 
   return true;
 }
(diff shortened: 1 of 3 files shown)


TFT/src/User/Menu/common.c
```diff
@@ -185,21 +185,36 @@ const void drawStandardValue(const GUI_RECT *rect, VALUE_TYPE valType, const voi
...
+    }
+    else
+    {
+      setDialogText(LABEL_WARNING, (uint8_t *) tempMsg, LABEL_CONFIRM, LABEL_BACKGROUND);
+      showDialog(DIALOG_TYPE_ERROR, NULL, NULL, NULL);
+    }
     return false;
   }
+  else
+    // temperature falling down to a target lower than the minimal extrusion temperature
```
digant73, line 213: maybe it's better to use {} for "else" statement

digant73 (changes requested):

TFT/src/User/Menu/common.c
```diff
@@ -185,21 +185,36 @@ const void drawStandardValue(const GUI_RECT *rect, VALUE_TYPE valType, const voi
...
+    if (heatGetTargetTemp(toolIndex) < infoSettings.min_ext_temp)
+      { // contiunue with current temp but no lower than the minimum extruder temperature
```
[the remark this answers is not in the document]
kisslorand: Fixed

TFT/src/User/Menu/common.c
```diff
@@ -185,21 +185,36 @@ const void drawStandardValue(const GUI_RECT *rect, VALUE_TYPE valType, const voi
 
 const bool warmupTemperature(uint8_t toolIndex, void (* callback)(void))
 {
-  if (heatGetCurrentTemp(toolIndex) < infoSettings.min_ext_temp)
-  { // low temperature warning
+  #define TEMP_OFFSET 5  // offset temperature to avoid denial of extrusion/retraction due to the nozzle temperature lag
+  if (heatGetCurrentTemp(toolIndex) < infoSettings.min_ext_temp - TEMP_OFFSET)
+  {// low temperature warning
```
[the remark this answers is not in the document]
kisslorand: Fixed

TFT/src/User/Menu/common.c
```diff
@@ -185,21 +185,38 @@ const void drawStandardValue(const GUI_RECT *rect, VALUE_TYPE valType, const voi
...
+      strcat(tempMsg, tempStr);
+      
+      setDialogText(LABEL_WARNING, (uint8_t *) tempMsg, LABEL_CONFIRM, LABEL_CANCEL);
+      showDialog(DIALOG_TYPE_ERROR, callback, NULL, NULL);
+    }
+    else
+    {
+      setDialogText(LABEL_WARNING, (uint8_t *) tempMsg, LABEL_CONFIRM, LABEL_BACKGROUND);
+      showDialog(DIALOG_TYPE_ERROR, NULL, NULL, NULL);
+    }
```
guruathwal, line 210: Extrusion should not be blocked when the temperature is above minimum extrusion temperatures, it should be blocked only if the target temperature is set through `M109`.

kisslorand: You are right, it should not be blocked. ...and it isnt't blocked, as it should be.

guruathwal: warmupTemperature returns false (Line no: 211) regardless of this. which will always prevent extrusion.

guruathwal: Yes, I missed line no: 197

kisslorand: No problem, thanks for checking.

gordo3di:
@kisslorand 

We've noticed a weird "random" bug where we would preheat the extruder and then Load filament using the Un-Load menu. After the purging was complete if we start a print from SD or USB it will not home and instead just print wherever the extruder was. Restarting the printer would solve the problem.

I can't seem to consistently replicate the issue so sadly I can't offer more info but 3 of us in my shop have had it happen across several TFT35 V3 touch screens when preheating then purging. Not a huge issue I was just wondering if anyone else experienced it. I haven't tried the latest firmware mentioned in this post.

kisslorand:
Do you have G28 in your gCode?
Preheat/extrude/load/unload etc. deals only with the E axis, has nothing to do with X and Y.

gordo3di:
> Do you have G28 in your gCode?
> Preheat/extrude/load/unload etc. deals only with the E axis, has nothing to do with X and Y.

Yes the gcode has G28 and the print is fine normally it only happens with the Un-Load menu. I also have G28 in the "start gcode before print" on the touchscreen firmware. That's why it's so weird since like you said it has nothing to do with X, Y or Z. I'll try this latest PR but I have noticed it from the firmware dated (4/21/21). Again not a big issue I just wasn't sure if others saw it happen.

kisslorand:
Very, very strange..
Do you have set in Marlin not to home if axis are known?

gordo3di:
> Very, very strange..
> Do you have set in Marlin not to home if axis coordinates are known?

Good idea. I know the setting your talking about but I can't seem to find it in marlin. I searched all my Marlin 2.0 files and even older releases. Did that setting get removed or changed?

kisslorand:
My bad. It is not an option in Marlin config, it is an option for G28. It's the "O" parameter that enables home only if coordinates are not known. (ex.: "G28 XY O")
Try, if you can, manually issue M701 in terminal and check if the problem reoccurs. If yes, it is a Marlin bug.

PR #1916, opened 2021-05-13
master->develop

wigw:
Hi, I ask you to reverse the icons. The bottom line is that when you reverse the Z axis in the menu, only the microstep icon is not reversed.


guruathwal:
You need to open an issue instead of a pull request.

Issue #1921, opened 2021-05-15
[FR] M154 - automatic position reporting


radek8:
Permanent problems with incorrect layer height could be solved with the new command 
M154 - automatic position reporting

https://github.com/MarlinFirmware/Marlin/pull/18427

kisslorand:
There are no more problems with incorrect layer height. It's been reworked by PR #1894 (which is already merged).

oldman4U:
Maybe it could help to reduce the traffic.

sarvenn:
> Maybe it could help to reduce the traffic.

https://reprap.org/wiki/G-code#M154:_Auto_Report_Position
Yeah, it says that it reduces serial comm traffic.

oldman4U:
PR1924 my friend

radek8:
Thanks, that's great.

PR #1924, opened 2021-05-17
Add gantry position auto report support

guruathwal:
- Add gantry position auto report support (through M154 if available in Marlin firmware).
- Auto-detect Position auto reporting from M115 response from Marlin firmware.
- Clean up some code.

Resolves: #1921


radek8:
Thank you very much.
Can you also add a comment to the config.ini file?
 
General options:
AUTO_REPORT_POSITION (in Configuration_adv.h)

oldman4U:
I will add it also to the ReadMe once merged.

Copy to SD Card root directory to update/config.ini
```diff
@@ -37,6 +37,7 @@
 #   M115_GEOMETRY_REPORT (in Configuration_adv.h)
 #   M114_DETAIL (in Configuration_adv.h)
 #   REPORT_FAN_CHANGE (in Configuration_adv.h)
+#   AUTO_REPORT_POSITION (in Configuration_adv.h)
```
digant73, line 40: the parameters were listed in the same order they appear in Configuration.h / Configuration_adv.h. So AUTO_REPORT_POSITION should be provided just before M115_GEOMETRY_REPORT

digant73 (changes requested):

PR #1947, opened 2021-05-24
Improve multiline text display to respect words

guruathwal:
- Improve multiline text display to respect words. Long sentences/strings will now be split into multiple lines without breaking words that are difficult to read.


Guilouz:
Why this string #define STRING_STATUS_SCREEN was removed from translation files / Language.inc and not from language_keywords.h ?

This string is not used anymore ?

guruathwal:
yes

Guilouz:
You have also missed to remove #define STRING_STATUS_SCREEN in language_cn.h

kisslorand:
In some cases new line is intended. Please take this into account when removing "\n".

guruathwal:
The display of lines is the same like it would display on a PC.

PR #1952, opened 2021-05-28
"°" not displayed correctly in german language on TFT70

MarkusThur:
### Requirements

UTF-8 Character "°" to be displayed correctly in german language on TFT70, in order to display unit "°C" of temperatures correctly.

### Description

A mistake in the encoding of the previous version of language_de.ini prevents the Character "°" from being fisplayed correctlyl. This regularly happens with "wrong" encoding settings, especially across plattforms.
This change contains correct encoding created on Windows 10 with VIM (Vim Settings :se enc=utf-8 :se fenc=utf-8 :se ff=unix)
 
### Benefits

Character "°" gets recognized and displayed correctly by utf-8 able IDEs / editors and by the TFT70 display.

### Related Issues

This issue may occure also:
- other displays
- other languages

### DIFF
--- a/Copy to SD Card root directory to update/Language Packs/language_de.ini
+++ b/Copy to SD Card root directory to update/Language Packs/language_de.ini
@@ -1,5 +1,5 @@
 #### Language Code:DE
-## Language Version:20210528
+## Language Version:20210522
 ## Maximum byte per keyword is 250 Bytes.
 ## Escape characters are not supported except newline '\n'
 
--- a/TFT/src/User/API/Language/language_de.h
+++ b/TFT/src/User/API/Language/language_de.h
@@ -145,7 +145,7 @@
     #define STRING_TIMEOUT_REACHED        "Zeit überschritten!"
     #define STRING_DISCONNECT_INFO        "Verbindung getrennt!"
     #define STRING_SHUTTING_DOWN          "Fahre herunter..."
-    #define STRING_WAIT_TEMP_SHUT_DOWN    "Warte bis Düsen-\ntemperatur unter\n%d℃ fällt." // Wait for the temperature of hotend to be lower than 50℃
+    #define STRING_WAIT_TEMP_SHUT_DOWN    "Warte bis Düsen-\ntemperatur unter\n%d°C fällt." // Wait for the temperature of hotend to be lower than 50°C
     #define STRING_POWER_FAILED           "Druck fortsetzen?" //Question Sign
     #define STRING_PROCESS_RUNNING        "Prozess läuft bereits!"
     #define STRING_PROCESS_COMPLETED      "Prozess fertiggestellt!"
@@ -188,9 +188,9 @@
     #define STRING_TOUCHMI                "TouchMi"
 
     // Values
-    #define STRING_1_DEGREE               "1℃"
-    #define STRING_5_DEGREE               "5℃"
-    #define STRING_10_DEGREE              "10℃"
+    #define STRING_1_DEGREE               "1°C"
+    #define STRING_5_DEGREE               "5°C"
+    #define STRING_10_DEGREE              "10°C"
 
     #define STRING_001_MM                 "0.01mm"
     #define STRING_01_MM                  "0.1mm"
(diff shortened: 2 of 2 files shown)


rebel1 (approved):

oldman4U:
Herzlich Willkommen!

;-)

rebel1 (changes requested):
You need  to update language_de.h also, because language_de.ini is autogenerated.

MarkusThur:
Danke 

Ich hab nur noch keine AHnung ob man so überhaupt antworten kann

MarkusThur:
> You need to update language_de.h also, because language_de.ini is autogenerated.

As I am new to this project (and didn't do open source coding for 11 years now), can you explain the autogeneration / show me where it happens? Could this cause the encoding problem?

oldman4U:
Man kann. 😁

Unfortunately I have no idea how the auto build works. @guruathwal could you please help with this topic!? Gurmeet knows everything.

guruathwal:
I can't confirm about VIM but all other programs including Atom, VSCode, Notepad, Notepad++ are displaying the Unicode "°C" correctly. Maybe you need to verify encoding settings in your editor.

MarkusThur:
guruathwal, yes those editors display the incorrect coded files correctly, as well as most browsers and github will do. They recognize the different options to encode the special characters in german of "äöüß" and do a autoconversion to display it correctly.

Vim, Vi and the BTT displays do not do that, so the encoding of those characters has to be correct utf-8 encoding, in order to make the display work correctly.
 
One way to check for that problem and solve it is:

open file with VI or Vim, then make sure you give following command:
`:set encode=utf-8`
check for the characters, replace them by the right encoded ones. give following command:
`:set fileencode=utf-8`
save the new file

the result you see here with the commits, the diff displays the line as changed, and marks those characters that seem to be same as changed, cause they are different encodings.
This is because the viewer here also does the "autoconversion" for german for displaying the file, but recognizes the difference.

This is a typical problem with german and embeded systems. And it becomes even worse if you try to programm the same display from the same sources with a windows, a MacOs or a unix-like machine.

If you like / insist I could screenshot the result on the display with the old .ini / .h file and the new one. Or you picj the old revision and the new revision and check yourself :-)

MarkusThur:
@oldman4U @guruathwal 

I found where the autoconversion between .h and .ini. it's in the platform_io.ini where it calls:
`post:buildroot/scripts/auto_gen_language_pack.py`
will check later how that script behave on the conversion, or if it can be modified to output correct .ini even with a incorrect coded headerfile

guruathwal:
@MarkusThur I also checked by uploading the old language_de.ini on the TFT, and all the text is being displayed correctly as expected.

MarkusThur:
Then there may also be a issue in the fonts for the 7" display, as it definitely did not display anything else than the submitted correctly, with the Vx.x.27 set of files for german and the fonts from the unified Theme. That's why I asked also which font editor fits the file format of those fonts in order to check them.

I just wrote it together in order to be able to fix this more reliable:
there are first of all to possibilities to represent 'degree celsius' in unicode.
U+00B0 (&#xB0;) folowed by U+0043 (&#x43;)
or
U+2103 (&#x2103;)

The correct encoding in UTF-8 is:
`0xC2 OxB0 0x43` or `0xE2 0x84 0x83`

and now we run in two issues.. 

First of all, many fonts do not know anything about U+2103 &#x2103; and display a `?` or another wired character instead. Windows Arial font for example displays `℅` (c/o) if you ask the windows character table for U+2103.

some systems may interprete `0xE2 0x84 0x83` as:
&#xE2;&#x84;&#x83; or as &#xE2;&#x8483; or as &#xE284;&#x83;

From what I have seen, the Fonts for the BTT 70 in the Unified theme at revision Vx.x.27 seems to not know that U+2103 character also.

The Representation i submitted and was accepted, cause it works is `0xC2 0xB0 0x43`.

And here we come to the issue it involves:

Windows and other systems encode &#xB0;&#x43; as:
`0x00 0xB0 0x00 0x43` which is the correct encoding in UTF-16 / windows-1252/latin-1,

Text editors doing autoconversion or not aware of UTF-8 encoding even my corrupt a UTF-8 file, if they interprete `0xC2 OxB0 0x43`automatically to UTF-16 / windows-1252/latin-1 encoding.
depending on system, the correct UTF-8 encoding may lead to:
&#xC2B0; &#x43; or &#xC2;&#xB0;&#x43; or something else more wired or do autoconvert to:
'0x00 0xB0 0x00 0x43' which is then misinterpreted by the TFT70 display.

PR #1955, opened 2021-05-30
Temp check overhaul in Extrude, (Un)Load & TuneEsteps menu

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This is an overhaul of the temperature check in the Extrude, (Un)Load & TuneEsteps menus.

### Benefits

- temperature check is done at a shorter interval
- extruder motor is activated only if the desired/target temperature is reached
- if the actual temperature is lower than the desired one due to the nozzle lag, the extruder will be automatically activated once the nozzle temperature is stabilized
- if the nozzle is cooling down but it is still above minimum extrusion temperature, it is possible to activate the extrusion motor, the actual temperature at the moment when the button was pressed will become the target temperature

Here's the logic behind this:
Temp Logic

### Related Issues

-  even if the target temperature was set higher than the minimal extrusion temperature, once the nozzle reached that minimal extrusion temperature, it was possible to activate the extruder motor
- temperature query was done at a slow rate, actual temperature could be lower by a few degrees than the displayed one, hence extruder motor could be activated at a wrong (lower than desired/permitted) nozzle temperature

### Extra work that wasn't the initial intention of this PR

- fixed TFT freezing in some cases in the mentioned menus (big thanks to @digant73  for finding the fix for the bug)
- removed the permission to navigate away from the "(Un)Load" menu whilst filament load or unload is on progress (it could lead to TFT freeze)

### Later edits:

- the same logic for temperature check is applied to the TuneExtruder (eStep calibration) menu also


TFT/src/User/Menu/LoadUnload.h
```diff
@@ -7,6 +7,15 @@ extern "C" {
 
 void menuLoadUnload(void);
 
+typedef enum
```
digant73, line 10: I usually provide declarations in .c  if they are not general declaration used also by other .c

TFT/src/User/Menu/common.c
```diff
@@ -199,55 +199,6 @@ void drawStandardValue(const GUI_RECT *rect, VALUE_TYPE valType, const void *val
   GUI_SetBkColor(origBgColor);
 }
 
-bool warmupTemperature(uint8_t toolIndex, void (* callback)(void))
```
digant73, line 202: I put those functions here (menu directory) because they also use UI (showDialog etc...). In the API directory no reference to GUI should be present

kisslorand: Fixed

TFT/src/User/Menu/LoadUnload.h
kisslorand: Fixed

TFT/src/User/API/Language/language_keywords.h
```diff
@@ -352,6 +352,9 @@ extern "C" {
 #define LANG_KEY_CLICK_FOR_MORE               "label_click_for_more:"
 #define LANG_KEY_EXT_TEMPLOW                  "label_ext_templow:"
 #define LANG_KEY_HEAT_HOTEND                  "label_heat_hotend:"
+#define LANG_KEY_DESIRED_TEMPLOW              "label_desired_templow:"
+#define LANG_KEY_WAIT_HEAT_UP                 "label_wait_heat_up:"
```
guruathwal, line 356: use the already existing similar keywords: LANG_KEY_TUNE_EXT_DESIREDVAL

TFT/src/User/Menu/Extrude.c
```diff
@@ -39,10 +39,6 @@ void menuExtrude(void)
   bool eRelative = false;
   uint32_t feedrate = 0;
 
-  while (infoCmd.count != 0)
```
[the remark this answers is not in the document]
kisslorand: Fixed

TFT/src/User/API/Language/language_keywords.h
kisslorand: Fixed

Issue #1837, opened 2021-04-13
[BUG] TFT35 - Busy processing, please wait.


Magiquall:
Hello,
I have the same problem as another post closed without really solution, with my BTT TFT35, for the moment connected to a MKS SGEN L V2 while waiting for my SKR-2, I have the message "busy processing, please wait "

I reduced the speed from 250,000 to 115,200 it was less frequent, but still present.
I swapped out the COM ports and had no more worries, until I printed. A one hour thirty minute print and the printer is frozen 3 layers from the end.

When it is frozen, it is not easy to put it back into service, you have to turn it off and then wait a little before turning it back on, often twice in a row.

Every time it was frozen there was "Processing busy, please wait" at the top and in red, but nothing at all, the screen gets stuck even after several minutes and as during printing it continues to heat up. have a hole in my impression.

I specify that I put the last update, the Marlin patch and the last firmware available on the github for the screen.

Is it a loss of connection with the motherboard, or the screen which is overloaded?

oldman4U:
Hi.

There has been no solution mentioned in this repository, because literally all users do not have the reported issue, but a few, and we are all using the same firmware. This makes it very unlikely that the firmware of the TFT causes this issue and this repository is about the firmware of the TFT's only. In case there is a way to reproduce the issue and there is a way to fix this in the TFT firmware, I am sure it will be done.

Magiquall:
Ok, I hope this is from the MKS SGEN L V2 board, because I stop the makerbase. I also had a problem with their screen, maybe this is the same source of the problem.
So I will wait for the SKR2, hoping that it fixes the problem and that it does not come from the BTT screen.

Is the memory of the BTT TFT35 display sufficient where you need to be careful?
For example, do not enable all of the features in the config.ini file.

Or are there functions that should not be activated together? Functions which could interfere with the operation.

In any case, thank you for your responsiveness.

And excuse me for my English.

oldman4U:
SKR-2 is for sure a great board, but Marlin support is not finished as we speak, as far as I know.

Resources are very limited on the TFTs but so far you can activate all functions in config.ini. If this makes sense - I do not know;-)

Most of us are not native English - so no worry.

Magiquall:
I got the github firmwaire for the SKR-2, I managed to compile it. Otherwise, I would not have ordered this card.
But as it becomes difficult to find an SKR 1.4 turbo on its own, I opted for the SKR2.

[...]

PR #1956, opened 2021-05-30
avoided possible freezes looping on loopProcess function

digant73:
**BUG FIXES:**
* looping on loopProcess function will create a freeze condition in case a user interaction is requested (e.g. dialog, popup etc... are displayed). The user interaction is not allowed (it's no more possible to click on any button). This PR makes possible the user interaction needed to complete the related task and thus also allowing to create the condition to leave the loop on loopProcess function

**NOTES:**
1) no fix is applied on the loop currently present on Extrude and LoadUnload menus due to the fact that PR #1955 will remove them (they are not needed)
2) it could fix bug #1837

**PR STATE:** ready to merge


TFT/src/User/API/Printing.c
```diff
@@ -388,14 +388,13 @@ void printAbort(void)
         request_M0();  // M524 is not supportet in reprap firmware
       }
 
-      setDialogText(LABEL_SCREEN_INFO, LABEL_BUSY, LABEL_BACKGROUND, LABEL_BACKGROUND);
-      showDialog(DIALOG_TYPE_INFO, NULL, NULL, NULL);
-
-      do
+      if (infoHost.printing)
```
MarkusThur, line 391: 2 Questions about "consistent coding"

1. at least in this file the explicit bool pattern if (boolparam == true) is used. Shouldn't it be same here?
2. at the current state (&hostPrintingConditionCallback == infoHost.printing) should be true always. Wouldn't it be more "future safe" to condtion on if(&hostPrintingConditonCallback == true)?

If I understood the idea of this added if right, it is in order to prevent the "Dialog" calls to happen if loopprocess() will not happen. (compare to old implementations comment on loopProcess in a bottom controled loop instead of a head controled one)
If this old comment is necessary, this implementation will fail on the very day, where &hostPrintingConditionalCallback is false while infoHost.printing is true for whatever reason, as the new implementation of calling loopprocess() is head controled.
At current state this cannot happen, but in future implementations it could happen.

digant73: 1) for boolean types the most used standard style (in the code) is the compact form. Obviuosly the standard form you reported is also correct and it could make sense to uniform it in all the code in the future
2) it's better the curent implementation

that code is correct. It initially check and avoid to display the "busy" message and to iterate on looProcess if the host is not printing anymore. Otherwise it will display the message and wait for the end of the print from the host

MarkusThur: Thank you for your reply.

about topic 2, that the new implementation is better and correct I saw.

The question I have is, at the current state the values of:
`infoHost.printing `
and
`&hostPrintingConditionCallback `
are identical, because
`bool hostPrintingConditionCallback(void)`
just wraps `infoHost.printing`.

so my question / implied suggestion is only a "refactoring", that could prevent issues in future implementations, if the implementation of hostPrintingConditionCallback(void) changes for whatever reason.

I think this line:
`if (&hostPrintingConditionCallback)`
is better readable and represents the intend better.

--- you may stop reading here, if you dont need closer explaination ---

To explain it:
Let's asume in a future implementation there is a "virtual" condition which allows following state:
`(infoHost.printing == true && &hostPrintingConditionCallback == false) == true`
then it would run into the displaying, but not do the till now exspected "one time looping" as the implementation of
`void loopProcessToCondition(bool (*conditionCallback)(void))`
now is a "head controled" instead of "bottom controled" loop.

The question implied is, if it maybe correct that in the case of above mentioned condition (printer is printing, but the callback functions says it's not needed to wait, that the Dialog is set, but thecall to
`loopProcess()`
does still not happen

digant73: although it will  not change in the future, your suggestion is correct so I will change the code

digant73:
closing and reopening the PR due to not working rebase.
PR reopened as #1959

PR #1959, opened 2021-05-31
avoided possible freezes looping on loopProcess function

digant73:
**BUG FIXES:**
* looping on loopProcess function will create a freeze condition in case a user interaction is requested (e.g. dialog, popup etc... are displayed). The user interaction is not allowed (it's no more possible to click on any button). This PR makes possible the user interaction needed to complete the related task and thus also allowing to create the condition to leave the loop on loopProcess function

**NOTES:**
* it could fix bug #1837

**PR STATE:** ready to merge


TFT/src/User/API/ConditionCallbacks.c
```diff
@@ -0,0 +1,24 @@
...
+
+bool noResponseConditionCallback(void)
+{
+  return (!requestCommandInfo.done);
+}
+
+bool hostPrintingConditionCallback(void)
+{
+  return (infoHost.printing);
+}
```
guruathwal, line 24: It does not make sense to place these functions in a new file which compares variable from other files.  It will be much better to place these functions in their respective original sources to avoid scattering of code.

`fullQueueConditionCallback` and `usedQueueConditionCallback` must be in interfaceCmd.c
`noResponseConditionCallback` should be place in gcode.c
`hostPrintingConditionCallback` should be place in Printing.c

They should be renamed to more understandable names:
fullQueueConditionCallback -> isFullCmdQueue
usedQueueConditionCallback -> waitingforEmptyQueue
noResponseConditionCallback -> waitingForResponse
hostPrintingConditionCallback -> isHostPrinting

MarkusThur: That's where my comment on it also pointed to.
The algorithm of that change is right, but the implementation could be in a better maintainable style.

In my way of understanding development, we shall have this change but mark it with a request to redo the coding to something easier to read/understand/maintain.

digant73: 1) Reviewed callback names.
2) A condition callback can be a mix of different conditions related to different exposed APIs, and in general:
 1) The condition callback should rely on the use of already existing functions even from different APIs, so a bind to a specific API could be not even possible. E.g. isNotEmptyCmdQueue() and isHostPrinting() use infoHost.wait that is external to both interfaceCmd and Printing APIs
 2) The condition callback is not meant to provide a new usefull function that could make sense to expose to an API
 3) Defining a condition callback to a specific API is not providing anything new usefull for the API's user. It is not a function meant to be commonly called to perform a task. It is instead a set of condition instructions whose scope and meaning is meant to be limited to the specific function loopProcessToCondition(). For that reason they have been collected in a specific file (also more easy to check and maintain). They are meant to be used by loopProcessToCondition() not to be exposed in an API to remain unused and adding confusion about its purpose.

Anyway, the main reason (with the current available APIs) to force the callbacks in specific files instead of a common file is that the conditions in the callbacks are based on testing attributes instead of using exposed API functions.

PR #1968, opened 2021-06-02
Revert PR #1959 in Extrude and (Un)Load menu

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR reverts the changes made by PR #1959 to the Extruder and (Un)Load menu.
PR #1959 adds unnecessary extra safety measures to these menus that could cause temporary freezing of the TFT.

### DIFF
--- a/TFT/src/User/Menu/Extrude.c
+++ b/TFT/src/User/Menu/Extrude.c
@@ -38,7 +38,10 @@ void menuExtrude(void)
 
   if (eAxisBackup.backedUp == false)
   {
-    loopProcessToCondition(&isNotEmptyCmdQueue);  // wait for the communication to be clean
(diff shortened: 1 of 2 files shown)


digant73:
If you maintain a loop to loopProcess I think it is better to maintain the current logic otherwise if a dialog is requested (e.g. due to an action notification) there will be a freeze as it happened before your merged PR

kisslorand:
I just made a torture test and I could freeze the TFT even using `loopProcessToCondition(&isNotEmptyCmdQueue)`.
I guess another type of fix is needed to avoid TFT freezing.

kisslorand:
Closing it, in tests TFT could be frozen.

PR #1969, opened 2021-06-03
Italian language update

BsCmOD:
Italian language update, I'm sorry if I hadn't correct also the INI file but I've got some problems with VSC


radek8 (commented):
Additional lines have been added, don't you want to translate them as well?

BsCmOD:
> 
> 
> Additional lines have been added, don't you want to translate them as well?

Yes off corse

PR #1972, opened 2021-06-04
read background color from icon for text

Msq001:
### Requirements

### Description
After calling `ICON_PrepareRead`, set the background color of the text to the corresponding icon.

### DIFF
--- a/TFT/src/User/Menu/PrintingMenu.c
+++ b/TFT/src/User/Menu/PrintingMenu.c
@@ -63,7 +63,7 @@ const char *const speedId[2] = {"Speed", "Flow"};
 bool hasFilamentData;
 
 #define TOGGLE_TIME  2000  // 1 seconds is 1000
-#define LAYER_DELTA  0.1   // minimal layer height change to update the layer display (avoid congestion in vase mode)
+#define LAYER_DELTA  0.01   // minimal layer height change to update the layer display (avoid congestion in vase mode)
 
 #define LAYER_TITLE  "Layer"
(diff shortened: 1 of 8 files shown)


TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -63,7 +63,7 @@ const char *const speedId[2] = {"Speed", "Flow"};
 bool hasFilamentData;
 
 #define TOGGLE_TIME  2000  // 1 seconds is 1000
-#define LAYER_DELTA  0.1   // minimal layer height change to update the layer display (avoid congestion in vase mode)
+#define LAYER_DELTA  0.01   // minimal layer height change to update the layer display (avoid congestion in vase mode)
```
kisslorand, line 66: Why so small value (0.01)? Such a small value will defeat its purpose to not overdrive the TFT in vase mode.

bigtreetech: Haha, it's just for testing. Because the whole icon is not redrawn, it seems acceptable to modify the Z layer in real time, but more testing is needed. I will change it back to 0.1 before merging

digant73:
@bigtreetech Hi, in case not already developed, I extended the changes to StatusScreen. It required just few changes also to function showLiveInfo (in menu.c).
Attached the two reviewed files in case you would consider to use them.

[tft.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/6602845/tft.zip)

PR #1977, opened 2021-06-06
Reworked fix to avoid freeze in loops caused by popups

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This is a reworked fix in order to avoid TFT freeze caused by popups during a loop. It is a universal fix, no need for extra functions for extra conditional process loops.

### Benefits

Popups are not triggered during sub-functions of the menus. If there's a popup than it is handled by the menu it occurred in.

### Related Issues

Previous fix could cause unfinished procedures caused by popups during a sub-function of a menu. The TFT wouldn't freeze because of that but results could be unpredictable as menus were restarted without the sub-functions successful finishing. Resets, initializations, final touches could been ignored/skipped.
Also, previous fix wasn't universal, for any new conditional loops more specific functions should be written.

### So, it's ready to be merged?

Not for the moment. It's been impossible to test all possible scenarios. Hopefully other developers would take a look and give their blessing or point out flaws to be fixed.

### Anything else?

Yeah. 

- SplashScreen type popup added.

It displays automatically upon calling it and it is the programmer's responsibility to delete it and redraw the relevant screen when the SplashScreen is not needed anymore.

- Extrude menu counter shows the extruded/loaded amount, not the temporary E coordinate as it was previously

The keypad input selects the desired extrude or unload length. Positive values will extrude and negative values will unload filament.

### DIFF
--- a/TFT/src/User/Menu/FeatureSettings.c
+++ b/TFT/src/User/Menu/FeatureSettings.c
@@ -383,7 +383,7 @@ void menuFeatureSettings(void)
         listViewRefreshItem(index);
       }
 
-    loopProcessWithPopup();
+    loopProcess_MenuLoop();
   }
 
   if (memcmp(&now, &infoSettings, sizeof(SETTINGS)))
--- a/TFT/src/User/Menu/FeatureSettings.c
+++ b/TFT/src/User/Menu/FeatureSettings.c
@@ -383,7 +383,7 @@ void menuFeatureSettings(void)
         listViewRefreshItem(index);
       }
 
-    loopProcess();
+    loopProcessWithPopup();
   }
 
   if (memcmp(&now, &infoSettings, sizeof(SETTINGS)))
--- a/TFT/src/User/API/Gcode/gcode.c
+++ b/TFT/src/User/API/Gcode/gcode.c
@@ -29,7 +29,7 @@ static void resetRequestCommandInfo(
 
   while (infoCmd.count || infoHost.wait)
   {
-    loopProcessNoPopup();  // Wait for the communication to be clean before requestCommand
+    loopProcess_PopupHandle();  // Wait for the communication to be clean before requestCommand
   }
   requestCommandInfo.inWaitResponse = true;
   requestCommandInfo.inResponse = false;
@@ -71,7 +71,7 @@ bool request_M21(void)
   // Wait for response
   while (!requestCommandInfo.done)
   {
-    loopProcessNoPopup();
+    loopProcess_PopupHandle();
   }
 
   clearRequestCommandInfo();
@@ -92,7 +92,7 @@ char *request_M20(void)
   // Wait for response
   while (!requestCommandInfo.done)
   {
-    loopProcessNoPopup();
+    loopProcess_PopupHandle();
   }
 
   //clearRequestCommandInfo();  //shall be call after copying the buffer ...
@@ -121,7 +121,7 @@ char *request_M33(char *filename)
   // Wait for response
   while (!requestCommandInfo.done)
   {
-    loopProcessNoPopup();
+    loopProcess_PopupHandle();
   }
 
   //clearRequestCommandInfo();  //shall be call after copying the buffer ...
@@ -172,7 +172,7 @@ long request_M23_M36(char *filename)
   // Wait for response
   while (!requestCommandInfo.done)
   {
-    loopProcessNoPopup();
+    loopProcess_PopupHandle();
   }
 
   if (requestCommandInfo.inError)
@@ -300,7 +300,7 @@ char *request_M20_macros(char *nextdir)
   // Wait for response
   while (!requestCommandInfo.done)
   {
-    loopProcessNoPopup();
+    loopProcess_PopupHandle();
   }
 
   //clearRequestCommandInfo();  //shall be call after copying the buffer ...
@@ -317,7 +317,7 @@ void request_M98(char *filename)
   // Wait for response
   while (!requestCommandInfo.done)
   {
-    loopProcessNoPopup();
+    loopProcess_PopupHandle();
   }
 
   clearRequestCommandInfo();
--- a/TFT/src/User/API/Gcode/gcode.c
+++ b/TFT/src/User/API/Gcode/gcode.c
@@ -3,11 +3,6 @@
 
 REQUEST_COMMAND_INFO requestCommandInfo = {0};
 
-bool isWaitingResponse(void)
-{
-  return (!requestCommandInfo.done);
-}
-
 static void resetRequestCommandInfo(
   const char *string_start,   // The magic to identify the start
   const char *string_stop,    // The magic to identify the stop
@@ -32,8 +27,10 @@ static void resetRequestCommandInfo(
   if (string_error2)
     requestCommandInfo.error_num = 3;
 
-  loopProcessToCondition(&isNotEmptyCmdQueue);  // wait for the communication to be clean before requestCommand
-
+  while (infoCmd.count || infoHost.wait)
+  {
+    loopProcessNoPopup();  // Wait for the communication to be clean before requestCommand
+  }
   requestCommandInfo.inWaitResponse = true;
   requestCommandInfo.inResponse = false;
   requestCommandInfo.done = false;
@@ -72,7 +69,10 @@ bool request_M21(void)
   mustStoreCmd("M21\n");
 
   // Wait for response
-  loopProcessToCondition(&isWaitingResponse);
+  while (!requestCommandInfo.done)
+  {
+    loopProcessNoPopup();
+  }
 
   clearRequestCommandInfo();
   // Check reponse
@@ -90,7 +90,10 @@ char *request_M20(void)
   mustStoreCmd("M20\n");
 
   // Wait for response
-  loopProcessToCondition(&isWaitingResponse);
+  while (!requestCommandInfo.done)
+  {
+    loopProcessNoPopup();
+  }
 
   //clearRequestCommandInfo();  //shall be call after copying the buffer ...
   return requestCommandInfo.cmd_rev_buf;
@@ -116,7 +119,10 @@ char *request_M33(char *filename)
     mustStoreCmd("M33 %s\n", filename);
 
   // Wait for response
-  loopProcessToCondition(&isWaitingResponse);
+  while (!requestCommandInfo.done)
+  {
+    loopProcessNoPopup();
+  }
 
   //clearRequestCommandInfo();  //shall be call after copying the buffer ...
   return requestCommandInfo.cmd_rev_buf;
@@ -164,7 +170,10 @@ long request_M23_M36(char *filename)
   }
 
   // Wait for response
-  loopProcessToCondition(&isWaitingResponse);
+  while (!requestCommandInfo.done)
+  {
+    loopProcessNoPopup();
+  }
 
   if (requestCommandInfo.inError)
   {
@@ -289,7 +298,10 @@ char *request_M20_macros(char *nextdir)
   mustStoreCmd(command);
 
   // Wait for response
-  loopProcessToCondition(&isWaitingResponse);
+  while (!requestCommandInfo.done)
+  {
+    loopProcessNoPopup();
+  }
 
   //clearRequestCommandInfo();  //shall be call after copying the buffer ...
   return requestCommandInfo.cmd_rev_buf;
@@ -303,7 +315,10 @@ void request_M98(char *filename)
   mustStoreCmd(command);
 
   // Wait for response
-  loopProcessToCondition(&isWaitingResponse);
+  while (!requestCommandInfo.done)
+  {
+    loopProcessNoPopup();
+  }
 
   clearRequestCommandInfo();
 }
--- a/TFT/src/User/API/Printing.c
+++ b/TFT/src/User/API/Printing.c
@@ -24,11 +24,6 @@ static bool updateM27_waiting = false;
 static float last_E_pos;
 bool filamentRunoutAlarm;
 
-bool isHostPrinting(void)
-{
-  return (infoHost.printing);
-}
-
 void setRunoutAlarmTrue(void)
 {
   filamentRunoutAlarm = true;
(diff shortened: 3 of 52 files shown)


digant73:
@kisslorand I think the current logic is correct, more compact and readable.
The logic of your PR is reintroducing a regression (the bugs causing freezes).
With your current PR both loopProcesNoPupup and loopProcessWithPopups will not handle a popup so it will create a freeze if used in a condition loop. E.g. mustStoreCmd will freeze in case a popup is displayed while the current logic will handle the popup and then will force the reload of the menu calling mustStoreCmd

The current loopProcessToCondition is the blocking version of loopProcess. It blocks to condition and/or to popup until they are handled. It also forces the reload of the calling menu (with a dummy menu).
The only improvement I would apply on loopProcessToCondition is simply to:
1) handle also the oneshot unconditional case (so to acts as loopProcess plus handling popups)
2) pass a new boolean argument "reloadMenuOnPopup" to enable/disable menu reload in case popups are handled
3) return a boolean value to inform the caller (e.g. a menu) that loopProcessToCondition handled or not a popup. That information will be used in conjunction with "reloadMenuOnPopup" to let the caller readraw the menu (if reloadMenuOnPopup is false) or to force the menu to exit and reloaded the entire menu (if reloadMenuOnPopup is true)
With that easy enhancement on loopProcessToCondition, you can handle any kind of use case:
1) use standard loopProcess to always force the exit from the menu to handle a popup
2) use loopProcessToCondition to handle the popup and forcing the menu reload
3) use loopProcessToCondition to handle the popup and notify the menu to redraw the menu without forcing the exit from the menu

kisslorand:
@digant73 
Thank you for taking the time but I think you didn't checked it through. Please tell me an example when in my PR the TFT will freeze.

With the method you implemented if in a loop of a sub-function  a popup appears, the sub-function is interrupted, popup handled than main function restarted. When the main function exits it returns to the stopped loop of the sub-function that is not needed anymore since the main function is exited. After the sub-function is finished it returns again to the main function from which it was called from. This can lead to a very unpredictable situation.
I really do not think it is a good idea to call a menu function from a menu function that can also call a menu function. We will have a nest of recursive menu functions which I am not that confident is OK for our tiny processors.

digant73:
@kisslorand I understood your changed but they are introducing the freeze conditions.
Just as examples, all the following code can now cause a freeze:

**Printing.c**

while (infoHost.printing == true)  // wait for the printer to settle down
{
  loopProcessNoPopup();
}


**gcode.c**

while (!requestCommandInfo.done)
{
  loopProcessNoPopup();
}


**interfaceCmd.c**

mustStoreCmd

while (pQueue->count >= CMD_MAX_LIST)
{
  loopProcessNoPopup();
}

what happens if Marlin sends a notification requiring the user to provide an answer to let Malrin complete a procedure and moving the status to ready (instead of remaining on busy)? That use case was the reason of the freeze on LoadUnload menu

Also mustStoreCmd is executed in many menus. That function now freeze

The scope of loopProcessToCondition is to allow a synchronous (blocking) function looping on some conditions and also able to  handle popup which handling is mandatory to avoid the system freezes

kisslorand:
Yup, that's one thing I didn't think of, that Marlin might need an answer to change its state. Back to more work...

I try to come up with an universal solution that doesn't need extra conditional functions for each new conditional loop.

digant73:
yes, and also loopProcessToCondition can be called not directly by a menu but from a not menu function (e.g. mustStoreCmd). In that cases, loopProcessToCondition can simply force the reload of the menu. That condition will be propagated backward to the menu that will force to reload the entire menu

TFT/src/User/Menu/FeatureSettings.c
```diff
@@ -383,7 +383,7 @@ void menuFeatureSettings(void)
         listViewRefreshItem(index);
       }
 
-    loopProcess();
+    loopProcessWithPopup();
```
guruathwal, line 386: It is better to keep the old name `loopProcess()` for when it is expected to run normally.

kisslorand: workin' on it already

kisslorand: ...and done

kisslorand:
@digant73 
I think I got your point covered now with commit  [<f28de1a>](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1977/commits/f28de1ae86d24471044b3d148c8d26de1bf4cd9a)

TFT/src/User/Menu/FeatureSettings.c
guruathwal: I sugested to keep the old name.

guruathwal: the `_MenuLoop()` is not required. it just makes the name longer and confusing.

kisslorand: > It is better to keep the old name `loopProcess()` for when it is expected to run normally.

When is it expected to run normally? Define "normally" please.


> the `_MenuLoop()` is not required. it just makes the name longer and confusing.

`loopProcess_MenuLoop()` is not suitable for any other place than the menu's main while loop (at least for this moment), as its name suggests it.
`loopProcess()` is a generic subroutine whilst `loopProcess_MenuLoop()` is one with a specific usage that uses `loopProcess()` as other function uses that too.
I know you never agreed with me ever, so if you insist, names can be changed to whatever you wish. There are many ways to skin a cat.

digant73: @kisslorand I agree with Gurmeet for different reasons:
1) maintain loopProcess where it is currently used.
  1) In case the menu using loopProcess is stateless, there is no problem on reloading the menu
  2) In case the menu using loopProcess is statefull (e.g. Z Offset, Mesh Edit etc...) there is no problem on reloading the menu
2) your proposed loopProcess_MenuLoop on the menu will not redraw the menu in case a popup is displayed and handled by loopProcess_MenuLoop
3) the use of macro in Extrude and LoadUnload menu is absolutely not need and vanish the goal to handle the popup without the need to reload the Menu

Point 2) and 3) can be easily satisfied by a function with the same logic of loopProcessToContiotion with the exception of loop on the condition:

**bool loopProcessWithPopup(bool reloadMenuOnPopup);**

RETURN:
  * true: if a popup was handled
  * false if no popup was handled

INPUT:
  * reloadMenuOnPopup: taken into account in case a popup was handled:
    * false: to let the caller handle the menu relaod. Usable in case the caller is a menu so it can decide to simply redraw the menu instead of reloading the entire menu
    * true: used in case the caller is not a menu so it is not able to provide any kind of menu relaod (e.g. mustStoreCmd).  The previous menu will be forced to be reloaded

That function can be used on both a menu (e.g. LoadUnload) and a not menu function (such as mustStoreCmd).
If used on menu (e.g. LoadUnload), the caller must check the return value and in case "true" was returned it can simply readraw the GUI. E.g. on LoadUnload menu:

if (loopProcessToCondition(&isNotEmptyCmdQueue, false) == true)  // wait for the communication to be clean
  menuDrawPage(&loadUnloadItems);

if (loopProcessWithPopup(false) == true)
  menuDrawPage(&loadUnloadItems);

If used on a not menu the caller does not need to check the return value. E.g. on mustStoreCmd:

while (infoCmd.count >= CMD_MAX_LIST)
{
  loopProcessWithPopup(true);  // wait for a free slot in the queue in case the queue is currently full
}

kisslorand: Point 2:
You might missed the changes in Main.c. 
`loopProcess_MenuLoop()` triggers the menu exit and the redraw of it in one single step (as opposed to two steps as before).

Point 3:
That macro is there to let the user know the TFT is busy with something. Without it if a popup appears and user selects an action (OK, Cancel) than the popup will hang there until the loop is finished. 
I know it is not needed from a functionality point of view. It's there foe smoother user experience (as so many other things in this FW).

digant73:
@kisslorand point 2) but your loopProcess_MenuLoop will always force to exit from the menu as it is the standard loopProcess. While my prosposed solution : 

if (loopProcessWithPopup(false) == true)
menuDrawPage(&loadUnloadItems);

will avoid to reload the menu. It will simply register the standard menu handler and redraw the menu. With  "false" passed as argument to loopProcessWithPopup, no dummy menu will be added on the menu queue, so no menu reload will be forced.
The registration and redraw of a menu is also already applied executing synchronous GUI functions such as numpad. When Numpad is closed the caller will redraw the menu. So my solution is compliant to an already existing logic

TFT/src/User/Menu/FeatureSettings.c
kisslorand: > The best suggestion I can give is simply to let me provide the ehnamcements on loppProcessToCondition and loopProcessWithPopup, and then you can use them in your new or existing menus according to your needs

The solution in this PR is universal, no need to make new functions if new loops appear with new conditionals. It covers now all situations.

Want a loop without popup handling? You have it, it's the old loopProcess().
Want a loop that handles popups? You have it, it's the loopProcess_PopupHandle().
Want a clean exit from a menu and redraw the menu in one single shot in case there was a popup? You have it, it's the  loopProcess_MenuLoop().
Want a flag that there has been a popup? You have it, loopProcess_PopupHandle() will set it by the value of "lastMenu". If its value is "NULL" there has been a  handled popup.
Want a flag that there has been a popup and a refresh is needed? You have it, loopProcess_MenuLoop() will set the value of  "infoMenu.menu[infoMenu.cur]" to "NULL".

guruathwal: > > It is better to keep the old name `loopProcess()` for when it is expected to run normally.
> 
> When is it expected to run normally? Define "normally" please.

Ok, when its function does not change after merging the changes (which is: it still performs the old routine), then it is better to leave the name as is it. 

> `loopProcess_MenuLoop()` is not suitable for any other place than the menu's main while loop (at least for this moment), as its name suggests it. 
> `loopProcess()` is a generic subroutine whilst `loopProcess_MenuLoop()` is one with a specific usage that uses `loopProcess()` as other function uses that too.

There are already functions For these purposes (`loopBackEnd()` for handling background tasks, `loopFrontEnd()` for handling UI related tasks) but instead of improving on them, more nested functions are being created for the same purpose which will make it hard to track bugs consume more CPU cycles. 

> I know you never agreed with me ever, so if you insist, names can be changed to whatever you wish. There are many ways to skin a cat.

I do not insist. I gave a suggestion. You take it as you like. I assume this is a discussion. It is not a discussion if everyone is having the same opinion. Members provide their views on my PRs also and many of them do not agree with many things. This does not mean I should start complaining about it. Stop acting like a wife. Why do you always become offended as soon as I comment on anything?
why would you skin a cat? I do not know who created this line but that is so cruel!

kisslorand: > Ok, when its function does not change after merging the changes (which is: it still performs the old routine), then it is better to leave the name as is it.

The `loopProcess_MenuLoop()` differs in its function from `loopProcess()`. Their purpose is the same but not their function.


> There are already functions For these purposes (`loopBackEnd()` for handling background tasks, `loopFrontEnd()` for handling UI related tasks) but instead of improving on them, more nested functions are being created for the same purpose which will make it hard to track bugs consume more CPU cycles.

Initially I made it with macros to avoid nesting but I switched to functions because I had a hunch you would disagree (with the macros). I used macro only where I really do not had any other choice.

 
>Stop acting like a wife. Why do you always become offended as soon as I comment on anything?

It's OK, let it out. I guess, after all, it is  a friendly discussion, right?
As far as I remember I took advices from you without "become offended" and most probably will do it  again.

> why would you skin a cat? I do not know who created this line but that is so cruel!

Some say it was Seba Smith in 1854  in the work ’Way down East or, Portraitures of Yankee Life".
You might like the lighter version: "there's more than one way to peel an orange"

kisslorand:
> @kisslorand point 2) but your loopProcess_MenuLoop will always force to exit from the menu as it is the standard loopProcess. 

Not always, only if there was a popup.

While my prosposed solution :
> 
> if (loopProcessWithPopup(false) == true)
> menuDrawPage(&loadUnloadItems);
> 
> will avoid to reload the menu. It will simply register the standard menu handler and redraw the menu. With "false" passed as argument to loopProcessWithPopup, no dummy menu will be added on the menu queue, so no menu reload will be forced.
> The registration and redraw of a menu is also already applied executing synchronous GUI functions such as numpad. When Numpad is closed the caller will redraw the menu. So my solution is compliant to an already existing logic

Great idea !!!
Should it be added to all menus?

digant73:
@kisslorand yes my proposed enhancement could be used for any kind of menu (standard, custom etc...). the only difference will be on the menu handled invoked (in the example it was the standard menu handler).
There is no need to complicate the logic as you did. E.g. no need to change main,c, no need to set infoMenu.menu[infoMenu.cur] = NULL etc... It will also create a loop of dependencies (the main depends of popup, the popu depends on main). The current logic handling menus is correct and easy. If a menu has to be displayed, simply add its handler  on top of the menu list.
I can simply push a PR with the enhancements I reported. You could simply have a look to that and let me know you comments. I think it will take less time.
I agree with Gurmeet to not change the current existing menus (not switching from loopProcess to loopProcess_MenuLoop etc...). As I said, the menus (stateless and statefull) are properly handling the menu relaod. No need to risk to introduce possible regressions in the code.
Also the loopProcessToCondition is already a general function. The condition is passed as an argument. That solution optimized the code and made it more readable.

Both:

loopProcessToCondition(&isNotEmptyCmdQueue, true)

and

while (infoCmd.count != 0)
{
  loopProcessWithPopup(true)
}

are the same, but the first one reduces the code when the same condition is used a lot in the project (as it is)

digant73:
@kisslorand just pushed a PR to show my easy enhancements. Just used the new loopProcessWithPopup function in Extrude and LoadUnload function as a usage example. Consider that this kind of logic is also used by the existing numpad function. When the controll is returned to the caller, the caller needs always to redraw its menu (no need to reload the menu).
If it will satisfy your needs and considered as a better solution I will make my PR available for a merge. Otherwise it will be close.

kisslorand:
@digant73 
Thank you for taking my considerations into account. I took a look and understood the concept. 
I still hold on to my initial opinion, it's not a global solution, instead it must be tailored for each individual  type of conditional loop.

@guruathwal was right, we are doing it the wrong way. Instead of dealing with the core of the problem we're patching the outcome of it using band-aids. 
Finally I was able to come up with a very simple fix that treats the root of the problem. No more nested functions, `loopProcess()` is back as it was, same function in the same places. There's also a global awareness if a popup is present or not on the display so anywhere in the program the right decision can be made (only redraw or reload the entire menu).

kisslorand:
Will make a PR with the new fix.

PR #1987, opened 2021-06-11
Simpler popup handle fix - avoid TFT freeze in loops

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This is a reworked fix in order to avoid TFT freeze caused by popups during a loop. It is a universal fix, no need for extra functions for extra conditional process loops. Popups are handled as soon as they are requested.
This fix treats the root of the problem. ~~Now there's also a global awareness if a popup is present or not on the display so anywhere in the program the right decision can be made (only redraw or reload the entire menu).~~

### Benefits

No more freeze during a keep-alive/background procedures loop.

### Related Issues

Previous fix isn't universal, for any new type of keep-alive/background procedures loop more specific functions has to be written.

### So, it's ready to be merged?

~~Not for the moment. It's been impossible to test all possible scenarios. Hopefully other developers would take a look and give their blessing or point out flaws to be fixed.~~
~~Almost.~~
Yes.

### Anything else?

Yeah. 

- SplashScreen type popup added.

It displays automatically upon calling it and it is the programmer's responsibility to delete it and redraw the relevant screen when the SplashScreen is not needed anymore.

- Extrude menu counter shows the extruded/loaded amount, not the temporary E coordinate as it was previously

The keypad input selects the desired extrude or unload length. Positive values will extrude and negative values will unload filament.


digant73:
@kisslorand honestly I think this solution is not going in the direction to optimize the code and providing a general purpose loopProcess function that can be specialized based on the needs (asynch, synch, with popup support etc...).
My fix was introduced just to fix the bug present in looProcess when nested in a loop. Furthermore the current loopProcessToCondition was done just to optimize the code when the same conditions are repeated in the code. Of course you could use the loopProcessWithPopup I proposed in my sample PR to handle any kind of condition. Also your solution is the opposite of universal (general puropse). It is forcing all the code to handle always the same kind of logic (synchronous) and also to handle a state variable. My and Gurmeet suggestion was simply to maintain the same logic in the existing menus (so async). It also allows to optimize the code. With the exception of the Printing menu (that could affect print quality), the reload of the menu is not a problem for both stateless and statefull menus.

In a C++ project, the general puropse loopProcess could be simply realized by the following signature (using default values for arguments):

**bool loopProcessWithPopup(bool handlePopup = false, bool reloadMenuOnPopup = true);**

With that function you can cover any kind of logic you can need:

loopProcess();    // current loopProcess function (asynch popup handling and asynch menu reloading)

loopProcess(true);    // sync popup handling and asynch menu reloading

loopProcess(true, false);    // sync popup handling and sync menu redraw

The code needs a multipurpose function not a universal (single handling logic) function. My changes were always focused on fixing bugs/or introducing new features with the minor impact on the existing code. For that reason I didn't add loopProcessToMenu in my PR (not yet needed).

kisslorand:
> @kisslorand honestly I think this solution is not going in the direction to optimize the code and providing a general purpose loopProcess function that can be specialized based on the needs (asynch, synch, with popup support etc...).

It's never meant to be that. It just fixes the freeze without the need of individual specific conditional functions for each and every new conditional situation for a loop.  If such a loopProcess is needed in the future, that's a future job. For the moment it works just fine as it is.

> The code needs a multipurpose function not a universal (single handling logic) function. My changes were always focused on fixing bugs/or introducing new features with the minor impact on the existing code. For that reason I didn't add loopProcessToMenu in my PR (not yet needed).

This code needs something different, a different menu structure/logic. There are many many more things that should have been done differently. (ex. DMA cache and its handling) 


So.
"Don't touch the main.c !" - OK, got it
"Don't create more nested functions!" - Aye aye, sir!
"Use menu redraw instead of  menu reload." - Got you
"It's too comlicated." - Made it simple
"Bring back loopProcess() as it was." - Understood, done

>Also your solution ... is forcing all the code to handle always the same kind of logic (synchronous)

So? Is it good? Is it bad? Why? Wasn't it the same since... forever? My solution proposed here is not  a game changer, just an alternative fix of a bug,  a fix that doesn't require individual and specific functions for each present and future conditional  for handling of a 'while' loop of `loopProcess()`.

> ...and also to handle a state variable.  

What's the problem of handling  a state variable? 

> My and Gurmeet suggestion was simply to maintain the same logic in the existing menus (so async). It also allows to optimize the code. With the exception of the Printing menu (that could affect print quality), the reload of the menu is not a problem for both stateless and statefull menus..

So no redraw now? No problem, just need to replace the redraw functions in the "if (popupState == PRESENT)" with `brake` or `return` and we're back to where we were (async), of course, with the exception of  the printing menu.

So, at the end of the day are there any functional problems with this solution proposal?

kisslorand:
Commit [<3094002>](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1987/commits/30940020838c319cbfef0f5e60934e045ee0e103) is the most simplest form of the fix. Now really everything is back just as it was, except the TFT freeze bug.
Developers can carry on the same way as before the first fix for this bug (PR #1956), no need to define any new procedures in case they want the `loopProcess()` inside a conditional loop.

Thank you @digant73 and @guruathwal !

kisslorand:
In case anyone wonders, yes, it is ready for merge.

bigtreetech:
Hello, this PR restores the submission of @digant73 . in the design of  @digant73, the processing of the new menu (such as response pop-up window) is added in while(1), and the menu can be redrawn in the middle of while(1) to avoid blocking (adding nesting method: there is some confusion and resource consumption indeed). After restoration, pop-up window that occur while(1) cannot be redrawn.
Does digant73 agree with your about this PR?

kisslorand:
>... the menu can be redrawn in the middle of while(1) ...

The menu is redrawn only after the while(1) loop is ended.

> Does digant73 agree with your about this PR?

We didn't talk about such aspect, he neither approved or disapproved it that I know of.
What I did in the end is to comply to @guruathwal and @digant73 observations.

digant73:
@bigtreetech, @kisslorand hi, I cannot approve or disapprove other PRs. I also didn't test this PR. I simply suggested (only my opinion) to maintain the current implementation because it seems to me more compact (and less error prone), easy to understand and use. The logic in loopProcessToCondition is similar to a GUI menu, with the loop on a condition instead of UI pointer and the execution of the popup when detected. The polling of other status variables makes the code more confusing and longer. If a blocking loopProcess is needed on some menus, it is more easy to simply call a function that will return a status information (e.g. popup handled or not handled) as with my proposed loopProcessWithPopup function.

kisslorand:
My last commit does exactly the things mentioned above., the execution of the popup when detected and no polling of any status variables. All this without the need of any extra conditional functions.
As said already, developers can carry on the same way as before the first fix for this bug (PR #1956), no need to define any new procedures in case they have to make a keep-alive/background procedures loop.

digant73:
Sorry but I already provided a lot of clarifications.
I'm referring you restored previous code (repeated loop etc...) that were made more compact and easy to handle (if a change is needed, you change the code in one point only. Less risk to forget to apply the change somewhere). We made this also for many other pieces of code (put on common.h) reducing the code and making changes more easy (see Gurmeet's last GUI changes).
The loop conditions are 4 repeated in all the code and there is no forecast for new conditions requiring to provide a loop condition inline. For that reason I didn't provide loopProcessWithPopup.
You are making the polling "if (getMenuType() == MENU_TYPE_DIALOG)" that is not necessary and that could not work (code not robust) if a different kind of menu will be triggered someday. It's not necessary because it would be enough that the loopProcessToCondition returns that information (agnostic for the caller). It could not work because you are limiting the check only to a dialog menu type while the condition should be only on the presence of a new generic menu (any kind of menu) on top of the queue (the dummy menu is not even registering any kind of menu type).
The splashScreen should also be changed. It still continue to call popupDrawPage that will set the menu type, register buttons etc... It should simply invoke GUI_DrawWindow.

kisslorand:
The language used (enounces without explanation and clarification), the avoidance to prove questioned statements, false things presented as facts makes it clear there's not much sense to try to reason, it's not a discussion anymore, it's a show to prove who has it bigger.
I really should not care, it's everywhere the same on this whole internet. I have better things to do, I really do not care anymore if any of my work is used or not in this project. 
You are free to claim more false facts about my work, and compare it to yours praising it to the Moon, I will not even try to defend it. The playground is all yours... I will not interfere with your claims anymore.
Have a beautiful week-end!

digant73:
I'm really sorry for your thoughts more because I think it's not reflecting me at all.
I provided any kind of clarification but you say I didn't.
I hope we both can continue to provide our support sharing our opinions in a more collaborative climate as it was till now.
But you cannot consider people keeping different opinions as personal attacks and start to offend. It happened now with me but it happened also with other users.

Issue #1505, opened 2021-01-11
[FR] Do not display the « on » choice when the smart filament detection is used


GenaUser:
Could it be possible that the “on” button does not appear in the settings menu when the smart filament detector is used in the printer?
When printing, if you want to activate or deactivate the smart sensor, you must select between two "on". This suspends the printing, because there is no end filament sensor if the smart is used.

oldman4U:
Sorry, I do not use such a sensor so maybe I am wrong, but this sounds more like a bug to me!?

guruathwal:
The runout sensor settings in the TFT are for connecting the sensor to the TFT itself. if the sensor is connected to the mainboard then this settings should be set to off.

oldman4U:
I believe there is still not clear what exactly you would like to achieve. Could you please explain a bit more in detail. Maybe some pictures could help

Thank you

GenaUser:
To your TFT display, you can connect a "filament detection module" or a "smart filament detection module" (not the both).
In "Filament sensor" of the TFT screen "Feature" menu, you have 3 possibilities: "OFF" "ON" "SMART".
![Filament sensor](https://user-images.githubusercontent.com/39191971/105525032-c9066000-5ce0-11eb-9979-102e1c690f2d.jpg)
"OFF" deactivates the detection of one or the other
"ON" activates the "Filament detection module" (analyzes whether the signal is activated or not)
"SMART" activates the "Intelligent Filament Detection Module" (analyzes a pulse frequency).
The problem lies in the fact that these three choices are alternated with each click => OFF, ON, SMART, OFF, ON, SMART ...
If, for example, you want to activate "SMART" while printing, you must first go to "ON". This function expects a constant signal (open or closed depending on the settings), whereas if you have a smart sensor, the signal is alternated. This inevitably causes printing to stop due to filament lack.
In short, you cannot activate a "SMART filament sensor" without stopping printing.
For my printer, I removed the "ON" option by changing the code, but automatic selection based on sensor choice would be really appreciated. For example, when "fil_runout_distance:" is enabled in config.ini

killrob:
i have a problem with this sensor, it doesn't work connected to TFT35 or TFT70 or another TFTXX as you like, yesterday my filament runs out and i had no advice from the smart filament sensor, print thrown away. If you have some trick to share.....

GenaUser:
I am using this function with a dual extruder on TFT35 V2 and V3 without problems.
[...]

PR #2002, opened 2021-06-17
Add filament sensor type selection in config

guruathwal:
- Add filament sensor type selection in config to show only relevant options in settings.

Fixes: #1505


guruathwal:
@oldman4U  Can you test this?

oldman4U:
😮 YES I can! 🙏

oldman4U:
Tested with smart filament sensor and it works great. Thank you Gurmeet.

Please merge

Ari-SSO:
Please somebody can explain me why now are 4 options? 

**Before was (If I am correct):**
0=Ignore the filament runout sensor connected to TFT. So let marlin use the filament runout sensor connected to the mainbord (if it is not configured in marlin then the printer has not any filament sensor).
1 and 2 =Use filament runout sensor connected to TFT, if 1 is a Normal sensor, if 2 the is Smart sensor type

**Now:**
0 or 1=Normal filament runout sensor (0=off, 1=on)
2 or 3=Smart filament runout sensor (2=off, 3=on)

But you only have 3 possible state, the state _OFF (option 0 and 2)_ is redundat. 

May be im missing something, but now you have the same functional code than befero but much complex and large, for example:



**Before (One instruction needed: the _if_): **
      
      void loopBackEndFILRunoutDetect(void) {
      if (infoSettings.runout == FILAMENT_RUNOUT_OFF)          // Filament runout turn off
        return;

**Now (thre instructions needed: _and, not, if_):**
    
    void loopBackEndFILRunoutDetect(void) {
      if (!(infoSettings.runout & 1))   // Filament runout turn off
         return;   

I dont see any functional the difference with the previous code, please somebody can help me to see the functional difference if it exists?

Thanks!!!.

guruathwal:
@Ari-SSO Earlier the setting in the TFT was toggled between three values (OFF -> NORMAL -> SMART). This was creating false triggers because to set the option to OFF it had to first set it to the wrong sensor type. 
Now the value in config is set to four different values to be able to identify the sensor type so that only the correct sensor ON values is toggled. Like, if set to 0 or 1 the sensor type will be NORMAL and on TFT it will toggle only between OFF and NORMAL. Similarly, if it is set to 2 or 3 then it will only toggle between OFF and SMART. The 4 options are provided to allow selecting the default state of the feature (Enable/Disabled) on the first boot.

Ari-SSO:
Ah, Ok!. So the problem was the false triggers, then to set if in OFF mode, is necessary to know the correct sensor type.

you help me to understand the problem, so now i understand the solution.

Thanks you!!!

Issue #2000, opened 2021-06-16
[BUG] (Part of the information of status screen stays, when entering U Disk to start a print. Screen capture not working!)


discip:
### Steps to reproduce

1. On main screen enter **`Print`** than **`U Disk`**.
2. Watch the information previously scrolling in the status section partially overlaying the navigation buttons.

**Expected behavior**
Screen gets completely redrawn every time a submenu is entered, except for e.g.: back button.

**Actual behavior**
Unwanted parts of the previous screen stay, if the **`U Disk`** menu is entered.

### Hardware Variant
**MKS_TFT28_V4_0**

### TFT Firmware Version & Main Board Firmware details
both compiled today / Marlin (is the bugfix branch)

### Additional Information
To document the behavior, explained above, I tried to use the screen capture option, by defining it in **`Configuration.h`**.
Unfortunately this did not work!
```
Scrennshot/U:_9.bmp failed!
```
Interestingly a folder named **`ScreenShot`** on the SD Card is created, but without any content.
On the U Disk there is nothing created.

thank you in advance

oldman4U:
I can not confirm the issue you report using BTT TFT35 E3 v3 and Marlin 2.0.9.

Screenshot is for TFT24 V1.1 (ILI9341) and TFT43/TFT50/TFT70 V3.0 (SSD1963) only.

oldman4U:
Before you see the content of the UDisk, you have to select which store you would like to use. Can you see the overlay there also? Has this been working fine with a previous version and if so, which version broke it?

Take a picture with your camera or even better a movie and upload it here.

guruathwal:
Please provide an image to inspect.

oldman4U:
Or a movie. 😁

discip:
Good afternoon,
here are some pictures:

![1](https://user-images.githubusercontent.com/53649486/122416743-45982480-cf89-11eb-909f-3e5966ba046e.jpg)
![2](https://user-images.githubusercontent.com/53649486/122416758-47fa7e80-cf89-11eb-9f3a-63b494252b23.jpg)
![3](https://user-images.githubusercontent.com/53649486/122416772-4a5cd880-cf89-11eb-8322-491f08263186.jpg)

Unfortunately I do not know since when this started.
thanks in advance

kind regards

PR #2003, opened 2021-06-17
Fix text overlay on icons in File menu Icon mode

guruathwal:
- Fix text overlay on icons in file menu with Icon mode. (#2000)

Fixes: #2000


oldman4U:
How is this reproducible? Does it need more than 1 page of gCodes?

PR #2006, opened 2021-06-18
Print time &  progress bar rework

kisslorand:
### Requirements

BTT or MKS TFT.

### Description

This PR brings back the progress bar to its king size on all TFT variants. The progress icon elements choice is configurable and can be altered by pressing the icon during print.
The variants are:
1. upper row -> percentage, lower row -> elapsed time
2. upper row -> percentage, lower row -> remaining time
3. upper row -> elapsed time, lower row - remaining time

For each started print the display shows the variant selected in the config.ini. If no remaining time info is present in the gCode file than Variant 1 is used (as it was before the implementation of elapsed time display).

### Related Issues

Cosmetical ones.
No more different layout on various TFTs., no more empty bottom row in the progress icon if no elapsed time info is present in the gCode file, At the end of the print no more empty bottom row, instead the percentage and the elapsed time is shown.


radek8:
So you were bothered by the percentages :-)

kisslorand:
LoL, no.
I was bothered by the awkwardness of the layout.

radek8:
I liked it with two times. 
100 people, 100 opinions

kisslorand:
Nothing changed in this regard. The default is with elapsed and remaining time displayed.

digant73:
@kisslorand hhmm there are already layout differences across TFT variants (status menu, terminal in TFT70). I don't think you could propose to that users to use a limited layout just because the other TFT variants have a smaller resolution/size. The same approach is being used also for some features (the recent prusa plugin). In the name of identical layout across all TFTs you are sacrifying the better TFT variants providing less info (2 instead of 3) and collecting again some different kind of info (% and time) under the same (wrong) time icon. I see your PR more applicable only to TFT43 and 50 just to allow the display of % in case someone is interested to that. Also there should be no need to make it configurable in config.ini.

oldman4U:
My 2 cents

I installed this PR on a TFT35 and prefer it over all other variants I have used in the past. If needed I can check 43 and 70 also and even the smaller screens but on 35 it is great.

Again my 2 cents from a pure user perspective.

digant73:
Imho filling any pixel in the screen with an icon, a progress bar etc at the cost to need to press here and there on small icons to display different kinds of info under the same group (icon) is not the way for a professional solution. Toggle icons were used in some menus as the only viable solution to provide usefull functions, and I still remember the feedbacks provided by different users. The printing menu is already providing enough info that people using a printer for work is not interested at all. There were also some FRs requesting exactly the opposite (provide a light menu with basic and well visible info). No need to make a Proof Of Concept for a TFT fw when those info are also provided without the need of any toggle icon and organized in correct way

kisslorand:
@digant73 
A few considerations:

1. the progressbar is supposed to be seen from as far as possible
2. as someone said once, the percentage is already there, in the form of the progressbar (you can even have it marked at each 10% for more precise reading)
3.  for me Cura is very very inaccurate, remaining time for me is a useless graphical gimmick, so I do not use it
4. I rather see numerical percentage and elapsed time in that icon and have the king size progress bar, than see only elapsed time and an empty line in that icon and a crippled progressbar
5. the numerical representations on that icon have so small fonts that you have to be within reach to be able to read it, since you're within reach, it is not such a huge effort to click that icon if you wish to see other info that it is displayed there; I think it is a small sacrifice for having the king size progressbar back for all TFTs

You say it is a limited version of the print layout, I say it is an enhanced version.

digant73:
As for many other features, I would simply suggest to make the use of the current and your proposed layout configurable in Configuration.h so anyone can easily choose the preferred solution. That means to avoid to add any configuration param in config.ini. At the end it is also not necessary for your layout. According to the file, your solution automatically switches to a proper view. That will create confusion with the setting in config.ini

kisslorand:
>That will create confusion with the setting in config.ini. 

I think users here are clever enough to understand the explanations I wrote there.

>For me it is absolutely more intuitive and visible the % in a standalone black background close to the progressbar than on a grey background close to other text and icon. Just a personal preference

A bigger progressbar is more visible than a smaller one. Elapsed and remaining time would be also more visible "in a standalone black background" but it seems it's not a problem that they are left in their "grey background close to other text and icon".

digant73:
But during the development of my PR on remaining time I spent more time collecting feedbacks and discussing about the possible layouts than on the target of the PR and no one asked/proposed other viable alernatives configurable.That would also helped me wasting less time. You are now proposing an alternative and I simply suggested to make it configurable in Configuration.h as usual.
For the configuration param in config.ini I simply don't see any need for that and the value could be also inconsistent. If time remaining is not provided then only % and elapsed time can be displayed. Otherwise time elapsed and time remaining can be displayed by default.You then can press on the icon to switch to the third layout % with time remaining. But, even if the config param will be always consistent, does it worth so much to make it configurable in config.ini when you also made a full screen progress bar? We didn't make configurable a lot of more important params imho. 
As I said, the use of a full lenght progress bar, text on top of an icon etc are related to personal preferences and constraints (for me it was grouping the info properly and sacrify redundand info on some TFT variants) and for that reason I asked to contnue to make a new alternative configurable at compile time. Make your layout the default, I don't care at all. The additions I made are not the default choices (not the alternative progress bar, not the terminal etc).
Please do not let pass the message that you are providing a universal solution for something that is a personal preference as it was mine although I was open to proposals.

kisslorand:
It's not a debate anymore, it's getting personal.
[ ...deleted a lot out of respect.. ]

digant73:
And I'm sorry you think that because there is nothing personal from my side.
I didn't agree with your last 2 PRs and I provided many explanations. No one spends so much time if he doesn't care about a person. I simply don't see benefits for me and my TFTs. I'm also suggesting to continue to provide new alternatives configurable in Configuration.h so anyone can use what he prefers.

kisslorand:
Yessssss, we just finished Rubik 360. So much better way to spend your free time...

oldman4U:
I asked it and it seems you know the answer. If so, why not changing it?

kisslorand:
For those who don't want to try a not merged PR, here's a link for a video showing what this PR brings:
https://www.dropbox.com/s/cycks7fb2ji2ibh/20210619_170745.mp4

bigtreetech:
I like this style too.

PR #2018, opened 2021-06-21
Filament runout sensor fix for new Artillery Genius printers

### LINKED ISSUE #2007: [BUG] Filament Sensor Problems
Hello,

I have a firmware for the TFT based on your development:
https://github.com/Dtcreation/Firmware-Molise-TFT

However, I have the problem with my "Artillery Genius" that the LED from the filament sensor is always green whether now filament is inside or not.

Is there a bug here?

Currently, the filament sensor does not work so.

PS. In the firmware of the TFT, the filament sensor is of course activated.

Thanks in advance.

kisslorand:
### Requirements

Artillery Genius - new version with dual led on the filament runout sensor

### Description

This PR fixes the non working filament runout sensors on newer Artillery Genius printers if the sensor is connected to the TFT as it is from factory.

Also, a small, non apparent bug is fixed, the MKS TFT has already pull-up resistors on the PCB connected to PB0 and PB1 pins, they are defined now as floating (no internal pull-up/down) if used as input pins.

### Benefits

Fixes #2007

### Related Issues

On newer Artillery Genius printers the filament runout sensor (with bicolor led) is not working if it is connected to the TFT as it is from factory.

### Notes

Those who compile themself the FW and have this printer, they should use the envelope "MKS_TFT28_NEW_GENIUS" for compiling.


omexlu:
Please really merge, important for the Artillery's new generation :) thanks

kisslorand:
I cannot do that, only @bigtreetech can do the merge.

kisslorand:
Do other Artillery models have the same issue or only Genius?

omexlu:
All with the new batch with bicode led will have this problem, maybe needed two versions for genius/X1.

omexlu:
Because of that I need an option in config may be the best solution so everyone with this issue can active or deactivate that 💪

kisslorand:
Please check  on that Facebook group you mentioned if there are other models affected by this problem.
Creating a selectable option in "config.ini" is out of the question. Once you selected the correct FW for your hardware, it is not going to change.

omexlu:
Many less use this firmware but theyare using it with the new batches have this problem.

omexlu:
All with bicolor have the issue but the most will not notice that 😊

kisslorand:
Sorry to repeat myself, please try to find out from those who complained on that Facebook group you mentioned, if heir printer is Genius or is this bug present on other models too. I have no resources to find which printers are affected by this issue,

omexlu:
The ones that contacted were Genius, but as far as I know the new Sidewinder x1 are also equipped with bicolors led. These are the new versions as also the support of Artillery told me.

omexlu:
Yes merged 💪 so I think this version will be precompiled in the future too?

Many thanks.

Sonicboom247:
just want to comment, i have two X1 and one has a clear runout sensor and the other a black one, both have dual color LED, one green when filament is inserted and red when Not. hope this helps.

PR #2026, opened 2021-06-24
Extended clickable icons in printing menu

### LINKED ISSUE #2015: [BUG] From Print Menu Bed/T0 Heat => LAST Heat settings from status screen
### Description
While Printing, when selecting the Bed Heat or T0 menu button, instead of showing the proper Heat Settings, It shows the LAST  heat settings that was selected from the main menu.

### Steps to reproduce
1. On the status screen, select the Bed Heat menu button
2. Return to the status screen (return)
3. Start a print 
4. Back to the Print status menu, Select T0 Heat menu button

**Expected behavior**
The T0 Heat settings are showing and are being available for modification

**Actual behavior**
The Bed Heat settings are showing and are being available for modification

### Hardware Variant
TFT35-E3-V3
BIGTREE_TFT35_V3.0_E3.27.x.bin
RepRapFirmware for STM32F4 based Boards FIRMWARE_VERSION: 3.3_2 ELECTRONICS: STM32F4 FIRMWARE_DATE: 2021-06-19 17:14:27
### Additional Information
It's as if the method "heatSetCurrentIndex" was not called before going to menuHeat when on the print status menu.
To verify this, after having done the above, once the print is completed, I went back to the status screen, selected the T0 heat menu button... Started a new print, and when on the print status menu, selected the Bed Heat menu button and the T0 Heat settings showed up.

### LINKED ISSUE #2021: Issue with [PR] Clickable upper icons during print #1996
This is a very minor point but it totally confused me initially and is inconsistent with the idle screen operation.

### Description
A nice new feature introduced with PR #1996 is to enable quick access to some of the print settings during an active print. It achieves this by allowing the upper icons to be clicked upon to enable direct access to the controls that adjust these settings - by cutting down on the number of key presses needed to access those controls.

While most of the icons work as intended to access the controls, adjustment of the T0 and Bed temperatures on my system require extra keys presses. This is because when the T0 or Bed icons are clicked on, the state of the T0/Bed toggle icon on the control page is indeterminate - I think it's left in whatever setting was used last. This means that in some instances, an extra key click is required before the temperature can be adjusted, which rather undermines the whole point of the enhancement in the first place.

To illustrate the issue better, here are some screen shots:

image
Screen shot of main screen during active printing

image
Screen after clicking on T0 icon - working as expected

image
Screen shot after returning to main screen after clicking on Back button and pressing Bed icon. Nozzle temperature adjustment still showing.

The Nozzle icon now has to be clicked on to select the Bed icon before the Bed temperature can be adjusted.

Now when I go back to the Main screen and click on the Nozzle icon again, this time, it brings up the Bed temperature adjustment.

Not a big deal but creates a lot of messing around and a strong probability of adjusting the wrong temperature zone - for an enhancement that's meant to cut down on key clicks and make things easier for the user.

For reference, when the printer is idle (Ready), clicking on the T0 icon, always brings up the Nozzle temperature control and clicking on the Bed icon always brings up the Bed temperature control. I feel the same should occur during active printing instead of leaving it apparently random.

### Hardware Variant
BTT-E3-RRF_V1.1
BTT-TFT35-v3.0

RRF 3.3_2
E3.27.x (latest as of today)

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR extends the clickable icons function in the printing menu. Each upper status icon has it's own function, except the Layer icon, which has no function associated yet, albeit it is clickable.

### Benefits

Nozzle status click brings up the Heat menu with the current nozzle preselected, clicking the bed status icon brings up the Heat menu with the bed preselected . Hopefully this clears the confusion created by previous implementation where those two icons had the same functionality, before they just called the heat menu and user had to select in the heat menu the desired element to be heated.
When print is finished, the status icons are not clickable anymore, no click sound when touching them. It should eliminate more possible confusion of users who before could hear the click sound and nothing happened.

### Related Issues

Fixes #2021
Fixes #2015


radek8:
Funguje to skvěle

Wombat37:
It didn't work for me. 

I don't see anything newer than 2 days in your repository so I'm probably not testing the right thing.

radek8:
> It didn't work for me.
> 
> I don't see anything newer than 2 days in your repository so I'm probably not testing the right thing.

Try this

[BIGTREE_TFT35_V3.0_E3.27.x.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/6712321/BIGTREE_TFT35_V3.0_E3.27.x.zip)

Wombat37:
The repository just updated (slow internet today) so I'll try again :)

Wombat37:
OK - testing now. Things looking great so far. Will be about another 15 minutes (want to check on the finish).

Wombat37:
It all looks good to me - both buttons called up the correct zone to adjust which is now really easy with the rotary encoder - just click and twist!

I don't quite understand the comment about the clicks being disabled at the end of the print. Those buttons can still be clicked while the printer is ready to call up temperature adjustment etc. I assume that this is a different part of the software.

Anyway, many thanks for doing this - it is much better!

Shall I close the issue (https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/2021) I raised?

kisslorand:
> I don't quite understand the comment about the clicks being disabled at the end of the print. Those buttons can still be clicked while the printer is ready to call up temperature adjustment etc. I assume that this is a different part of the software.



This is the print finish stage where you cannot click those status icons.

>Shall I close the issue (#2021) I raised?

Since it is not merged yet, the issue #2021 is still present in the current firmware.

Wombat37:
Thanks for the clarification on the Finished screen.

Yes it makes absolute sense to disable the icon clicks in this state (and then reenable them when on the main idle/ready screen).

Wombat37:
Yeah!!!

PR #2029, opened 2021-06-25
config.ini filament sensor description refinement

oldman4U:


bigtreetech:
Hello,

```
end_gcode_enabled:1
start_gcode_enabled:1
```
Is it necessary to keep it in the description? Because these are disabled by default.

oldman4U:
I would keep it but can change it to a kind of description instead of showing it like the function itself. Like a simple:

Enable start and end gcode.

oldman4U:
Hi.

_I just checked my PR and it does not change any setting and adds a text based description already:
# NOTE for users which have a filament sensor connected to the mainboard:
# Enable the start and end gcode below, and add the following commands to the start/end gcode: 
# start_gcode: M75\n
# end_gcode: M77\n_

I believe it would be good to add it like it is.

PR #2034, opened 2021-06-28
Remove unnecessary `LABEL_NOZZLE_STABILIZING`

Msq001:
### Requirements

### Description
Modifying the logic of limiting extrusion is consistent with that of Marlin. The extrusion will be restricted only when the current nozzle temperature is lower than the `min_ext_temp` instead of the target temperature.(The target temperature may be very high, but I only need to exceed the set cold extrusion limit temperature `min_ext_temp` to load/unload filament, reducing unnecessary waiting time)

And in the previous logic, when the temperature of the nozzle is stable, the extruded filament will bring out the heat of the nozzle, and the temperature of the nozzle will drop by about 1 degree, and the screen will frequently prompt unnecessary "waiting for stability"


kisslorand:
Can you extrude ABS at 180 degree?

Msq001:
> Can you extrude ABS at 180 degree? How about nylon?

If we need high temperature filament, we should set the desired cold extrusion temperature in config.ini (`min_ext_temp`)


> As far as I remember the temperature lag is set to 5 degree Celsius, that should cover the nozzle temp drop
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/3bfbd38f58b28c7895de1035b9617a0e8325179b/TFT/src/User/Menu/common.c#L320
According to the logic here, as long as the current temperature is lower than the target temperature (even if it is only 1 degree), the screen will stop extrusion. The following logic is: prompt to wait for LABEL_WAIT_HEAT_UP when the difference is more than 5 degrees, prompt LABEL_NOZZLE_STABILIZING when the difference is less than 5 degrees

kisslorand:
> If we need high temperature filament, we should set the desired cold extrusion temperature in config.ini (`min_ext_temp`)

Many machine can print from PLA to Nylon. Should the user change config.ini based on what material is used in each printing job? Sounds fishy...

> > As far as I remember the temperature lag is set to 5 degree Celsius, that should cover the nozzle temp drop
> 
> https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/3bfbd38f58b28c7895de1035b9617a0e8325179b/TFT/src/User/Menu/common.c#L320
> 
> 
> According to the logic here, as long as the current temperature is lower than the target temperature (even if it is only 1 degree), the screen will stop extrusion. The following logic is: prompt to wait for LABEL_WAIT_HEAT_UP when the difference is more than 5 degrees, prompt LABEL_NOZZLE_STABILIZING when the difference is less than 5 degrees

Yes, right. But..
I remember the discussion where @guruathwal said that for him more than 2 degree is too much of a difference, his extruder would skip.

The logic behind the whole thing was that user will set his extrude temperature and to not allow extrusion below that temperature, even when nozzle dropped just slightly under the desired temp. Now suddenly user would be able to start extrude Nylon at 180 degree or will have to wait to 260 degree to extrude PLA, depending how `min_ext_temp` is set in config.ini. That seems more hassle than wait a few seconds for the nozzle to reach the right temperature for extrusion.

Msq001:
> Yes, right. But..
> I remember the discussion where @guruathwal said that for him more than 2 degree is too much of a difference, his extruder would skip.

guruathwal said that the 2 degree(`TEMPERATURE_RANGE`) is enough, and you said that need 5 degree lag(`NOZZLE_TEMP_LAG`).  It's OK in my experience whether it's 2 or 5 degree. But the logic now is that even 1 degree of lag can lead to limited extrusion, and I will frequently get the prompt  `LABEL_NOZZLE_STABILIZING` in my use.

> Now suddenly user would be able to start extrude Nylon at 180 degree or will have to wait to 260 degree to extrude PLA, depending how `min_ext_temp` is set in config.ini. That seems more hassle than wait a few seconds for the nozzle to reach the right temperature for extrusion.

Use target temperature instead of `min_ext_temp` to limit
Disadvantages: As mentioned above, the logic does not match most of the motherboard software (Marlin, RRF) and requires additional waiting time.
Advantages: Avoid extruding high-temperature filament at low current temperature, resulting in the extruder gnawing filament. But in fact, this limit still depends on the user's own settings. If the target temperature of the user is too low (ABS filament, target temperature setting 180 degree), the gnawing filament still can not be avoided.

Haha, the above comparison of advantages and disadvantages seems like I'm looking for an excuse to convince myself.  

OK, I will restore the limit temperature to the target temperature.  And I will discuss with you again If there are any inappropriate scenes in my actual use.  But the logic of waiting `LABEL_NOZZLE_STABILIZING` must be modified, because it affects the use of extrusion menu.

kisslorand:
I guess there is no right or wrong way because, as you said, in the end it depends on the user anyway.
Just note that in the extruder tuning procedure it is really preferable to extrude at the desired temperature.

My idea was to use the hotend at least in a pseudo-stabile state. If it is not something desirable than yes, all those messages are unnecessary. BTW, you can avoid those messages if you wait for the nozzle to settle down and start extrusion afterwards.

Msq001:
> you can avoid those messages if you wait for the nozzle to settle down and start extrusion afterwards.

` when the temperature of the nozzle is stable, the extruded filament will bring out the heat of the nozzle, and the temperature of the nozzle will drop by about 1 degree, and the screen will frequently prompt unnecessary "waiting for stability"`

As I said above, even if the temperature stabilizes, when I extrude more filament, the filament will take part of the heat and the temperature will decrease by about 1 degree, although it will rise again and stabilize to the target temperature soon. But my extrude operation on the screen will be forced to interrupt and wait for it to stabilize again, which I don't think is necessary.

kisslorand:
Than what if we allow "desired_temp - LAG_TEMP" for extrusion?

Msq001:
That's how it works here https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2034/commits/dceff971fca0af773388cec9baad9c7ce41fffdc

kisslorand:
Yup.
Don't you want to auto-extrude once the temp is in the right value?

TFT/src/User/API/Language/language_cn.h
```diff
@@ -341,10 +341,10 @@
     #define STRING_FILAMENT_COST          "\n已使用耗材成本: %1.2f"
     #define STRING_NO_FILAMENT_STATS      "\n无耗材历史数据"
     #define STRING_CLICK_FOR_MORE         "点击查看详情"
-    #define STRING_EXT_TEMPLOW            "喷头温度低于最小挤出问题 (%d℃)."
+    #define STRING_EXT_TEMPLOW            "喷头温度低于最小挤出温度 (%d℃)."
     #define STRING_HEAT_HOTEND            "加热喷头到%d℃?"
     #define STRING_DESIRED_TEMPLOW        "Hotend temperature is below\ndesired temperature (%d℃)."
-    #define STRING_WAIT_HEAT_UP           "Wait for it to heat up."
+    #define STRING_WAIT_HEAT_UP           "等待加热完成."
     #define STRING_NOZZLE_STABILIZING     "Wait, nozzle stabilizing..."
```
guruathwal, line 348: You can remove the label `STRING_NOZZLE_STABILIZING` from all language files as it is no longer used.

PR #2035, opened 2021-06-28
Fix runout trigger during resume

### LINKED ISSUE #2024: [BUG] Extruder doesn't lower after multiple runout detections
### Description
If the runout sensor triggers but you don't put in filament or it false triggers again immediately, the buffer will keep filling up with the runout raise height and it will not lower back down to the same height. We have ours set to raise 10mm so with each trigger it will add 10mm. Additionally, we babystepped the extruder down (in this case 20mm) to continue the print however when it reached the next layer it tried to lower an additional 20mm into the model causing sad faces all around.

### Steps to reproduce
1. Trigger a runout
2. Don't insert filament
3. Watch extruder raise each time

**Expected behavior**
1. Runout triggers and extruder goes to parked position and raises up.
2. If another runout is detected, the extruder should lower back to the original height before raising again or it should ignore the raising.

### Hardware Variant
SKR v2 rev b
TFT35 V3
Marlin 2.0.8.2
TFT firmware from June 14, 2021

### Additional Information
I'll try to find the code and fix myself and post here if I'm successful unless someone else knows the code better and can do an easy fix. I can't remember seeing this issue posted here yet.

guruathwal:
- Fix runout trigger during resuming which is causing runout to trigger during resuming. (#2024)

Fixes #2024

### DIFF
--- a/TFT/src/User/API/Printing.c
+++ b/TFT/src/User/API/Printing.c
@@ -481,13 +481,12 @@ bool printPause(bool isPause, PAUSE_TYPE pauseType)
   if (infoPrinting.pause == isPause) return false;
 
   loopDetected = true;
-  infoPrinting.pause = isPause;
 
   switch (infoFile.source)
   {
     case BOARD_SD_REMOTE:
     case BOARD_SD:
-      if (infoPrinting.pause)
+      if (isPause)
         request_M25();   // pause
       else
         request_M24(0);  // resume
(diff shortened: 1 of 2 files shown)


gordo3di:
I finally had a chance to test this but I still have the same behavior. Still triggers a runout and adds 10mm (my raise height) each time if i don't insert filament. @oldman4U did you get to try this?

oldman4U:
Have you tried PR or current master? Using PR you can not use pre-compiled firmware.

oldman4U:
I tested and it works.

PR #2047, opened 2021-07-05
Remove duplicate code and misc. clean-up

guruathwal:
- Remove some duplicate code.
- Extruder Calibration/Tuning menu now uses pause/ park location settings instead of hardcoded position.
- Misc code clean-up and reduce some memory usage.

### DIFF
--- a/TFT/src/User/API/HomeOffsetControl.c
+++ b/TFT/src/User/API/HomeOffsetControl.c
@@ -3,7 +3,6 @@
 
 static float z_offset_value = HOME_Z_OFFSET_DEFAULT_VALUE;
 static bool home_offset_enabled = false;
-
 // Enable home offset
 void homeOffsetEnable(bool skipZOffset, float shim)
 {
@@ -38,7 +37,7 @@ bool homeOffsetGetStatus(void)
 // Set Z offset value
 float homeOffsetSetValue(float value)
 {
-  mustStoreCmd("M206 Z%.2f\n", value);
+  sendParameterCmd(P_HOME_OFFSET, AXIS_INDEX_Z, value);
   mustStoreCmd("M206\n");  // needed by homeOffsetResetValue() to retrieve the new value
   z_offset_value = value;
 
@@ -62,7 +61,7 @@ float homeOffsetResetValue(void)
   float unit = z_offset_value - HOME_Z_OFFSET_DEFAULT_VALUE;
 
   z_offset_value = HOME_Z_OFFSET_DEFAULT_VALUE;
-  mustStoreCmd("M206 Z%.2f\n", z_offset_value);  // set Z offset value
+  sendParameterCmd(P_HOME_OFFSET, AXIS_INDEX_Z, z_offset_value);  // set Z offset value
   mustStoreCmd("G1 Z%.2f\n", unit);              // move nozzle
 
   return z_offset_value;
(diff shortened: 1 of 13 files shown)


TFT/src/User/API/HomeOffsetControl.c
```diff
@@ -3,7 +3,6 @@
 
 static float z_offset_value = HOME_Z_OFFSET_DEFAULT_VALUE;
 static bool home_offset_enabled = false;
-
```
digant73, line 6: I suspect you removed the empty line by mistake

TFT/src/User/API/interfaceCmd.c
```diff
@@ -553,7 +553,7 @@ void sendQueueCmd(void)
             setPrintProgressPercentage(cmd_value());
 
           if (cmd_seen('R'))
-            setPrintRemainingTime((int32_t) (cmd_float() * 60));
+            setPrintRemainingTime((cmd_value() * 60));
```
digant73, line 556: the value in the gcode file is a float not a in (or long int)

digant73 (changes requested):

TFT/src/User/API/ProbeOffsetControl.c
```diff
@@ -82,7 +82,8 @@ float probeOffsetResetValue(void)
   float unit = z_offset_value - PROBE_Z_OFFSET_DEFAULT_VALUE;
 
   z_offset_value = PROBE_Z_OFFSET_DEFAULT_VALUE;
-  mustStoreCmd("M851 Z%.2f\n", z_offset_value);  // set Z offset value
+
+  sendParameterCmd(P_PROBE_OFFSET, AXIS_INDEX_Z, z_offset_value);  // set Z offset value
   mustStoreCmd("G1 Z%.2f\n", -unit);             // move nozzle
```
digant73, line 87: align the comment here (usual two empty chars after end of line)

TFT/src/User/API/interfaceCmd.c
guruathwal: I checked the values with Prusa sli3r. The values are reported as minutes (int).
image

TFT/src/User/API/HomeOffsetControl.c
guruathwal: some old habits

TFT/src/User/API/interfaceCmd.c
digant73: this is the gcode file provided to me for testing the M73.
here I see also R1.9 etc...

[Prusa_sample.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/6765447/Prusa_sample.zip)

guruathwal: The code in that file is added manually. The slicer generates the parameters in M73 in whole numbers.
Here is a slicer generated file: https://gist.github.com/guruathwal/9b88045e277c3e00a1a87cddbc8c45e3

digant73: ok, but if I'm not wrong it was reported (maybe by Radek) that it could be set as a fraction of a minute. Maybe someone could also clarify it. just to be sure at 100% it can be only an integer value

guruathwal: It is surely reporting as `int`  in Prusa Slicer: 
 https://github.com/prusa3d/PrusaSlicer/blob/3fa78b52b2958be49c442136d8b4971b0a0d4d5c/src/libslic3r/GCode/GCodeProcessor.cpp#L380-L382

radek8: I didn't talk about the tenth value of switch R ;-)
Test file with decimal value sent by user: belese

PR #2061, opened 2021-07-14
bugfix #2040 + improved #2035 + minor cleanup

### LINKED ISSUE #2040: [BUG] Refresh printing page issue MKSTFT28 - Filament Runout issue - PARTIALLY SOLVED - Use a good USB drive :)
### Description

Menu navigation is fine. When you start a print (USB or SD) and try to click on icons during heating, the screen starts to lag till the temperature reach the target. If you try to stop the print the LCD freeze. While printing everything is ok.

### Steps to reproduce

1. Compiled on the 29th June last commit
2. Start a print from SD or USB
3. Touch the icons on printing menu
4. Generate the Gcode with Prusa slicer and turn on M73

**Expected behavior**
Responsive click

**Actual behavior**
Lag, Refresh rate low, impossible to stop the print

FILAMENT SENSOR NOT WORKING

### Hardware Variant
MKS TFT28 - Artillery (not the last one) 

TFT: Jun 29 commit - MKS_TFT28_V4_0
MKS: Marlin 2.0.7.2 and also 2.0.9.1

Required "LCD_SET_PROGRESS_MANUALLY" in Marlin tried with both on or off, nothing change.

EDIT: using .27 stable release no issue

Sending a gcode WITHOUT M73 on, both LCD touch and Filament sensor are working

digant73:
**BUGFIX:**
* fixed #2040. That bug was introduced by PR #2035. The FIL_IsRunout() function needs always to be periodically performed to maintain the proper runout state while PR #2035 avoided to perform FIL_IsRunout() in case of no printing or paused print. In case of runout, resuming the print forced FIL_IsRunout() to return (as initial state) always "false". That value will not allow to resume the print in case the runout sensor uses a not inverted logic (e.g. MKS TFT). In that case that initial value triggers a runout and the print is always forced to pause with no possibility to resume

**IMPROVEMENTS:**
* improved runout logic. Current runout logic allows to resume the print even in case of a pending runout. In that case the resume will loose/waste (no filament is provided on top of the print) at least one gcode of the print file before the runout is detected and the print is paused. The new logic will simply avoid to resume a print in case of a pending runout. That will avoid any wasted gcode. As already done for pending Marlin dialog (e.g. M600), an error message (Filament Runout!) is displayed on the notification bar trying to resume a print with a pending runout
* simplified loopBackEndFILRunoutDetect() funtion. The runout state is now totally handled on that function (no need to split it across also loopFrontEndFILRunoutDetect() function. Also no need to check the status of a print (not printing, paused etc...). The runout state must always be periodically updated to provide at any time the correct state when requested (e.g. on PrintingMenu to allow or deny the print resume)
* minor cleanup

fixed #2040
resolved #2035

**PR STATE:** ready to merge

### DIFF
--- a/TFT/src/User/API/extend.c
+++ b/TFT/src/User/API/extend.c
@@ -29,13 +29,11 @@ void PS_ON_Off(void)
 // Filament runout detect
 #ifdef FIL_RUNOUT_PIN
 
-static bool update_PosE_waiting = false;
+#define POS_E_UPDATE_TIME   2000
+#define ALARM_REMINDER_TIME 10000
 
-// Set whether we need to query the current position
-void positionSetUpdateWaiting(bool isWaiting)
-{
-  update_PosE_waiting = isWaiting;
-}
+static bool posE_isUpdated = false;
+static bool sfs_isAlive = false;  // Use an encoder disc to toggles the runout. Suitable for BigTreeTech Smart Filament Sensor
 
 void FIL_Runout_Init(void)
 {
@@ -59,6 +57,17 @@ void FIL_Runout_Init(void)
   #ifdef FIL_RUNOUT_PIN_5
     GPIO_InitSet(FIL_RUNOUT_PIN_5, infoSettings.runout_invert ? MGPIO_MODE_IPU : MGPIO_MODE_IPD, 0);
   #endif
+}
+
+// Set whether we need to query the current position
+void FIL_PosE_SetUpdate(bool isUpdated)
+{
+  posE_isUpdated = isUpdated;
+}
+
+void FIL_SFS_SetAlive(bool isAlive)
+{
+  sfs_isAlive = isAlive;
 }
 
 bool FIL_NormalRunoutDetect(void)
@@ -127,16 +136,6 @@ bool FIL_NormalRunoutDetect(void)
   return (runout == infoSettings.runout_invert);
 }
 
-static uint32_t update_PosE_time = 2000;
-// Use an encoder disc to toggles the runout
-// Suitable for BigTreeTech Smart Filament Sensor
-static uint8_t SFS_IsAlive = false;
-
-void FIL_SFS_SetAlive(uint8_t alive)
-{
-  SFS_IsAlive = alive;
-}
-
 bool FIL_SmartRunoutDetect(void)
 {
   static float lastExtrudePosition = 0.0f;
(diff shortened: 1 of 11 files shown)


oldman4U:
Inverted logic is the reason. Do you know how this also was related to the slicer used.

digant73:
@oldman which bug/user feedback are you referring to?

oldman4U:
omexlu's answer 6 days ago in #2040

digant73:
hmm it seems to me omexlu has a different kind of problem. I just asked him to provide the gcode file causing the freeze in his printer

TFT/src/User/API/extend.c
```diff
@@ -202,31 +212,28 @@ bool FIL_IsRunout(void)
...
 }
 
 void loopBackEndFILRunoutDetect(void)
 {
-  if (!(infoSettings.runout & 1))   // Filament runout turn off
-    return;
-  if (!isPrinting() || isPaused())  // Check for No printing or printing paused before reading sensor
-    return;
-  if (!FIL_IsRunout())              // Filament not runout yet, need constant scanning to filter interference
+  if (!(infoSettings.runout & 1))  // Filament runout turn off
```
guruathwal, line 221: Removing the printing/pause check at line 212 will bring back #2024 bug. The sensor state should not be read during pause/resume procedure. This will result in incorrect positions which may cause damage to the nozzle or the print if the nozzle is lowered in to the print.

digant73: Hmm no. After you got the first filament runout message, the print is paused and the nozzle is parked. Then you cannot resume the print until the runout is fixed (e.g. filament is changed). This PR simply avoid you can resume the print if runout is still there (bug 2024 is avoided)

guruathwal: Does the logic work for smart filament sensors also?

digant73: it remained the same as the old logic (previous to your PR). For SFS, the runout value is used just to detect a transition (the value is not even important). I'm going also to mount an SFS.
If you have already it and you could check it it will be appreciated

oldman4U: I will do in a few hours with both kinds of sensors

digant73: I installed the SFS on my printer and made some testing. The logic for SFS works as the previous logic. The runout value is handled in a different way than standard sensor. So it takes some time before triggering a new condition. But that's the current logic provided in FIL_SmartRunoutDetect() function that could maybe improved in a further PR

oldman4U: Tested with both kind of sensors successfully. 

👍🏻

PR #2073, opened 2021-07-25
implemented FR #2032

### LINKED ISSUE #1945: [FR] (Add an option to block the screen on a marlin or graphic mode, remove the switch mode)
add an options for completely disable the switch mode, to stay in the default mode and no longer be able to change mode

### LINKED ISSUE #2032: [FR] (block touch input while in idle mode)
If you want to "wake up" the display it's pretty likely you accidentally click on a button and trigger an action. That's especially the case when you set idle brightness to 0%. 

Add a "block input while in idle" option to the brightness options.

Block input while brigtness is 0%

**Additional context**

### LINKED ISSUE #2077: [BUG] (short description)level corners fails after new update
### Description
I just updated the tft35v3 firmware, after updating, and trying to level the corners in the ABL menu, the BL Touch fails on the 3rd probe. it says probe failed. When the bed id being probed the pin doesnt stow with the new firmware. On the 3rd corner probe, the BL touch stays red, even after the probe touches the bed. 

### Steps to reproduce

1. [First Step]go to abl menu
2. [Second Step] click on L Corner
3. [and so on...] Click start

**Expected behavior**
I expect it to successfully probe 4 corners

**Actual behavior**
when it gets to the right rear corner, the bl touch stays red when the probe touches the bed, and gives an error on screen "probe failed"

### Hardware Variant
tft35 v3
### Additional Information

I went back to the previous firmware for the tft35 v3, and the issue is gone

digant73:
**IMPROVEMENTS:**
* Implemented FR #1945: Now param "default_mode" in config.ini can be configured also as Blocked Marlin and Touch mode
* Implemented FR #2032: For the TFTs supporting brightness dimming (supporting params in config.ini "lcd_brightness", "lcd_idle_brightness" and "lcd_idle_delay").
* New param "block_touch_on_idle" added on config.ini:
  If enabled, when the LCD is on idle (dimmed) then the first touch on the display will
  simply restore the normal LCD brightness. The touch is then skipped, preventing to
  trigger any undesired action due to the dimmed display.

 **NOTE:** The block is always avoided if the LCD brightness is restored by the use of  rotary encoder instead of pressing on the display.
* New param "block_touch_on_idle" configurable on ScreenSettings menu
* Code optimization in MarlinMode and SelectMode
* HAL cleanup: HAL libraries MUST be independent from any printer setting such as infoSettings and from TFT logic such as LCD dimming etc... They must provide only HW related API. The HAL API will be used passing (as parameter) any specific printer param setting. The new HAL "Knon_LED" was created under User/HAL. The new API "LCD_Dimming" was created under User/API etc...
* includes.h cleanup: all the included .h files are organized starting from the basic libs (e.g. HAL) and grouped by project directory hierarchy
* config.ini and Configuration.h alignment: The syntax and format of the parameters are now the same (e.g  the  boolean values "true" and "false" are set to "1" and "0" also in Configuration.h as it happens in config.ini (the boolean value "true" is by definition a value different than "false" (where false is usually considered "0")
* cleanup on .c and .h dependencies. Often the .h contained indirect dependencies needed also in .c. Now direct dependencies in .c and .h are used as much as possible. This make more consistent the code
* Minor cleanup

**BUGFIX / FR:**
resolves #1945
resolves #2032
fixes #2077

**PR STATE:** ready to merge


kisslorand:
So I have to hit the TFT twice after the idle timeout even if my idle brightness is 10% lower than regular brightness?
I don't know about others but it will piss me off for sure.

digant73:
Tomorrow I will add a configurable param in config.ini to block input on idle state

oldman4U:
What is the functionality of a display which is dimmed by 10%?

TFT/src/User/Menu/ScreenSettings.c
```diff
@@ -391,6 +393,12 @@ void menuBrightnessSettings(void)
         listViewRefreshItem(curIndex);
         break;
 
+      case 3:
+        infoSettings.block_touch_on_idle = (infoSettings.block_touch_on_idle + 1) % 2;
```
kisslorand, line 397: infoSettings.block_touch_on_idle = 1 - infoSettings.block_touch_on_idle

digant73: I followed the syntax used all around the project (also already present in ScreenSettings). Although that solution is even more compact

kisslorand: Which one do you think is faster? I have a hunch that the bitwise operation...

digant73: In the project, many times a boolean variable is compared with value 1 (for the true value) so it should be better to avoid the use of XOR or !. The more safe expression is the current one or b = (1 - b)

kisslorand: Just as a fun fact, an if..then is 10 times faster than %,
`i++;`
`if (i==circular_rollover_limit) i=0;`
 is 10 times faster than
`i = (i + 1) % circular_rollover_limit;`

Division always takes a hit on MCU.

radek8:
@digant73 
Can you have the language ini files generated again, after the last change?
Thank you.

digant73:
The languageXX.ini files are automatically generated from languageXX.h files by btt and pushed on github. So no need to apply the changes in the .ini files

Guilouz:
@digant73 In Configuration.h

/**
 * Default Mode
 * Set Marlin/Touch mode as the default mode at startup.
 *
 * NOTE: Mode switching is possible only for Marlin Mode and Touch Mode by a long press of
 *       three seconds on the display or holding down the encorder button for two seconds.
 *
 *   Options: [Marlin Mode: 0, Touch Mode: 1, Blocked Marlin Mode: 2, Blocked Touch Mode: 3]
 */
#define DEFAULT_LCD_MODE MODE_SERIAL_TSC  // Default: MODE_SERIAL_TSC

What is the default mode now ? Options are 0, 1, 2 or 3, what about MODE_SERIAL_TSC ?

digant73:
@Guilouz MODE_SERIAL_TSC is already defined as 1

Guilouz:
@digant73 

For better consistency with the config.ini file, wouldn't it be better to define it as is?

#define DEFAULT_LCD_MODE 1 // Default: 1

PR #2088, opened 2021-08-01
Add more warning for Delta printers

Guilouz:
- Add warning to disconnect probe before use Mesh Editor and ZOffset features.
- Update french translation.


TFT/src/User/Menu/BedLeveling.c
```diff
@@ -82,7 +96,12 @@ void menuBedLeveling(void)
         break;
 
       case KEY_ICON_1:
-        infoMenu.menu[++infoMenu.cur] = menuMeshEditor;
+        #if DELTA_PROBE_TYPE == 2  // if Delta printer with removable probe
```
digant73, line 99: invert the logic. The default matching condition should be for not removable probe. so #if DELTA_PROBE_TYPE != 2 then use old logic #else use the new logic,

TFT/src/User/Menu/BedLeveling.c
```diff
@@ -110,9 +129,14 @@ void menuBedLeveling(void)
       case KEY_ICON_5:
         if (infoMachineSettings.zProbe == ENABLED)
         {
-          storeCmd("M851\n");
-          zOffsetSetMenu(true);  // use Probe Offset menu
-          infoMenu.menu[++infoMenu.cur] = menuZOffset;
+          #if DELTA_PROBE_TYPE == 2  // if Delta printer with removable probe
```
digant73, line 132: same as previous comment

PR #2084, opened 2021-07-31
Clean up and optimize printing screen

guruathwal:
- Clean up and reduce RAM usage in the Printing screen.
- Optimize the Printing screen to load faster.
- Prevent unnecessary status icon redraw in the Printing screen.
- Status icons on the Printing screen are now clickable after the print is finished.
- Clicking on the Layer/Z-layer status icon now opens the Babystep menu.
-  Misc. clean-up and reduce global variables.
-  Add chamber temperature status in status screen and printing screen (#1182). The bed heating icon will toggle between bed and chamber status If chamber heating is available.

Resolves #1182, Resolves #2085

PR #2096, opened 2021-08-03
Layer number display during print

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR adds the option to show the layer number during print . It can be altered between layer height and layer number by pressing the Layer (nozzle) icon/display area. The default layer display type can be set in the `config.ini` file.
The information is read from the gCode. Most slicers put a comment at each layer start, a comment that shows the layer number to be printed. This PR uses that comment.
Unfortunately this info can be read only if printing from the TFT SD Card or TFT USB.

### Benefits

At some point there was a request for this but I cannot find it.

### Related Issues

There are no bugfixes in this PR.

### Notes

This PR opens a door for future features. It introduces the parsing of comments from the gCode file. 
There are many other useful infos in the comments that slicers generate, it is is now easier to interpret and use them if needed.

For the moment this PR is marked as draft as it is in conflict with PR #2084.


radek8:
@kisslorand
And does the layer height icon have to be clicked?
If you give options in Config.ini
- show layer height
- display layer number
- toggle between height and layer number

That would be absolutely great and it will not conflict with the PR2084.

I would like to automatically switch between height and layer number rather than a static display of one or the other information.

oldman4U:
I asked for it and this is great.

Thank you

Guilouz:
Why not display layer height AND layer number at the same time at the same time? Like duration / percentage icon.

kisslorand:
@radek @Guilouz 
Good ideas!

kisslorand:
Option to show both (layer height & number) implemented.

Guilouz:
@kisslorand do you think it's possible if ;LAYER:0 is seen to add +1, to display layer 1 / 1561 on TFT and not 0 / 1561 ? 

Because Cura start the first layer at value 0, like this : 

;LAYER_COUNT:1561
;LAYER:0 (first layer)
...
;LAYER:1560 (last layer)

Otherwise at the last layer we will have 1560 / 1561.

kisslorand:
@Guilouz 
It was in the agenda, it has been implemented.
Now layer number "0" means the printer is not printing an actual layer, instead it is doing some preliminary job before the actual print begins (like nozzle prime, etc).

kisslorand:
I think this PR is ready to go to his master :)

sarvenn:
Does anyone know how can I see layer number using prusaslicer? Verbose mode?

PR #2082, opened 2021-07-30
New HW_Init, ModeSwitching and Touch_Endoder API

digant73:
**IMPROVEMENTS:**
* Created new API HW_Init, ModeSwitching and Touch_Encoder: the hw functions (generic hw and specific for Marlin and Touch mode) were moved on HW_Init API. The SW encoder functions were moved on Touch_Encoder API. The mode switching functions were moved to ModeSwitching API
* extend.c/h renamed to SmartFeatures,c/h: the smart home features was moved from menu.c to SmartFeatures.c
* PowerFailed.c/h menu renamed to PrintRestore.c/h: power failed functions were moved to the new API PowerFailed (this API is used also by other menus and API (e.g. Printing)
* Changed land code for dutch language file: The proper land code is NL instead of DU
* Minor optimizations
* Some cleanup

**PR STATE:** ready to merge

PR #2100, opened 2021-08-04
fixed bug preventing to select and print a file

digant73:
**BUGFIX:**
*Fix for a bug introduced by my last merged PR #2082 preventing to display the dialog box asking to start the print for the selected file.

**BUG SEVERITY:** Major

**PR STATE:** ready to merge

### DIFF
--- a/TFT/src/User/Menu/Popup.c
+++ b/TFT/src/User/Menu/Popup.c
@@ -195,7 +195,7 @@ void _setDialogCancelTextLabel(int16_t index)
 */
 void showDialog(DIALOG_TYPE type, void (*ok_action)(), void (*cancel_action)(), void (*loop_action)())
 {
-  if (!(infoSettings.mode & 1) == MODE_MARLIN)  // if standard/blocked Marlin mode, then exit
+  if ((infoSettings.mode & 1) == MODE_MARLIN)  // if standard/blocked Marlin mode, then exit
     return;
 
   popup_redraw = true;
--- a/TFT/src/User/Menu/Popup.c
+++ b/TFT/src/User/Menu/Popup.c
@@ -195,7 +195,7 @@ void _setDialogCancelTextLabel(int16_t index)
 */
 void showDialog(DIALOG_TYPE type, void (*ok_action)(), void (*cancel_action)(), void (*loop_action)())
 {
-  if ((infoSettings.mode & 1) == MODE_MARLIN)  // if standard/blocked Marlin mode, then exit
+  if ((infoSettings.mode == MODE_MARLIN) || (infoSettings.mode == MODE_BLOCKED_MARLIN))  // if standard/blocked Marlin mode, then exit
     return;
 
   popup_redraw = true;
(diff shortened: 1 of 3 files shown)


TFT/src/User/Menu/Popup.c
```diff
@@ -195,7 +195,7 @@ void _setDialogCancelTextLabel(int16_t index)
 */
 void showDialog(DIALOG_TYPE type, void (*ok_action)(), void (*cancel_action)(), void (*loop_action)())
 {
-  if (!(infoSettings.mode & 1) == MODE_MARLIN)  // if standard/blocked Marlin mode, then exit
+  if ((infoSettings.mode & 1) == MODE_MARLIN)  // if standard/blocked Marlin mode, then exit
```
guruathwal, line 198: The `(infoSettings.mode & 1)` is not needed. The value of **infoSettings.mode** is always assigned absolute values. So performing the bitwise is overkill for this purpose. `(infoSettings.mode == MODE_MARLIN)` will give the same result with 1 less CPU cycle.

digant73: no because default_mode can now assume values from 0-3. 0 and 2 are for Marlin mode while 1 and 3 are for Touch mode. Same logic you used for fil_runout. The bitwise operator is the faster we can use. As mentioned by Kisslorand (on another thread)  a possible optimization could be applied to the logic to alternate the values for boolean value. B = (B + 1) % 2 could be converted to a more optimized B = 1 - B (it's ok for both boolean and integer that can assume values in range 0-1)

guruathwal: > no because default_mode can now assume values from 0-3. 0 and 2 are for Marlin mode while 1 and 3 are for Touch mode.

If this is the case then, IMHO, I think the `LCD_MODE` in Settings.h should be updated to include the 4 values because it is currently providing incomplete values.

Using logical operators like % make the code easier to understand. The C/C++ compiler automatically converts the logical operators like % to bitwise operators, so it does not make sense to change all the existing code for this purpose unless the code is specifically changing each bit individually like the Filament runout and like in parameter settings. 
**(B + 1) % 2** is used to give possibility to change the value range. **B = 1 - B** is unsafe for unsigned integers. Instead of **B = 1 - B** it is better to use **B = !B** to toggle boolean values which is also safe against negative values and faster than **B = 1 - B**.

digant73: for infoSettings.mode I defined MAX_MODE_COUNT 4 (used in config.c). However I could also extend LCD_MODE including the blocked Marlin/Touch modes. But I should also make small changes in the function HW_InitMode (that currently simply needs to know if Marlin or Touch mode has to be configured. blocked Marlin/Touch are equivalent to Marlin/Touch mode).

I know that for boolean type the standard is B = !B but it is equivalent to "if B != 0 ? B = true : B = false)" so it should be less compact and faster. If we guarantee (as it is already now) that initial value for B is 0 or 1 (in config.ini and parsed by config.c), then B = 1 - B is safe and it should be more compact and faster than B = !B. However, it was just an opinion for a  possible optimization. Sure, it's not a must to implement

PR #2105, opened 2021-08-06
added UI Setting menu + code optimization + minor bugfix

### LINKED ISSUE #2104: [BUG] Compilation fails when UNIFORM_LIVE_TEXT_BG_COLOR is defined
Impossible to compile when **#define UNIFORM_LIVE_TEXT_BG_COLOR** is defined in Configuration.h since recent PR.

Capture d’écran 2021-08-06 à 13 22 30

Capture d’écran 2021-08-06 à 13 22 41

@digant73 @guruathwal

digant73:
**IMPROVEMENTS:**
* Added UI Settings menu on ScreenSettings menu: The UI related settings in config.ini are now moved from the FeatureSettings menu to the UI menu
* Added logical port ID on Terminal: For any relayed message (a message received from an supplementary serial port), the ID of the logical port (e.g. 2 for serial port 2) originating the message is provided before ">>" (e.g. "2>>M900"). That will allow to better follow the message flows.
* Optimized code in ScreenSettings menu: some optimization in the code to reduce flash usage
* Improved PrintingMenu menu: UNIFORM_LIVE_TEXT_BG_COLOR (if enabled in Configuration.h) applied also to the new alternating Bed/Chamber icons (flickering text is avoided)
* Improved StatusScreen menu: in case Chamber is not supported/enabled, the Bed icon redrawn is avoided (flickering text is avoided)
* Cleanup on Serial and SerialConnection APIs: The logic to enable/disable all the Multi-Serials is moved from the low level Serial  API to the specific TFT variant's SerialConnection API. That will allow to easily handle (e.g. enable/disable) any Multi-Serials port at run-time
* Improved code robustness on Serial.c and gcode.c: Data allocated at runtime are initialized (0) and a check on the validity of data pointer is made before releasing or allocating the data

**BUGFIX:**
* Renamed wrong icon name "status_chan**d**er.bmp" to "status_chamber.bmp" for TFT24/TFT28 on Round Miracle theme
* Fixed wrong buffer size usage in config.c: It was wrongly used ACK_MAX_SIZE instead of LINE_MAX_CHAR. LINE_MAX_CHAR was also increaded to 200 (instead of 100) just to be sure to parse any setting (e.g. 75 chars of cancel gcode + the setting name)
* Fixed #2104 

fixes #2104

**PR STATE:** ready to merge


Guilouz:
Just try your PR, I have a strange issues : 

- During update screen is inverted then go back to normal on menu

IMG_1607

IMG


- Since new add of status chamber, bed temp, bed temp target and Bed word are flickering when chamber are disabled

Can you take a look ?

digant73:
it should not be related to my PR. However I will check this night. My PR should ask you to provide also config.ini file

Guilouz:
> it should not be related to my PR. However I will check it this night. My PR should ask you to provide also config.ini file

I have done a full update, TFT folder, bin, config.ini and language.ini and what's happened.

For issue with flickering, I think the TFT tries to rotate bed and chamber while this operation should be disabled when chamber is disabled.

digant73:
I don't have chamber in my printer and everything is ok. Even if a enable chamber on config.ini (although no chamber will be obviously displayed due it is not supported by Marlin)

Guilouz:
Perfect working fine now. Thanks.

guruathwal:
"Show ACK in Terminal" is not a UI setting.

TFT/src/User/Menu/StatusScreen.c
```diff
@@ -212,6 +219,7 @@ void drawStatusScreenMsg(void)
 
   IMAGE_ReadDisplay(rect_of_keySS[KEY_INFOBOX].x0, rect_of_keySS[KEY_INFOBOX].y0, INFOBOX_ADDR);
   GUI_SetColor(INFOMSG_BKCOLOR);
+
```
guruathwal, line 222: Please remove the extra line space. I prefer it the way it was.

TFT/src/User/Menu/StatusScreen.c
```diff
@@ -256,8 +264,10 @@ static inline void toggleTool(void)
         currentFan = (currentFan + 1) % MAX_FAN_COUNT;
       } while (!fanIsValid(currentFan));
     }
+
     // switch speed/flow
     currentSpeedID = (currentSpeedID + 1) % 2;
+
```
guruathwal, line 270: Please remove the extra line space. I prefer it the way it was.

TFT/src/User/Menu/StatusScreen.c
```diff
@@ -337,6 +347,7 @@ void menuStatus(void)
     toggleTool();
     loopProcess();
   }
+
```
guruathwal, line 350: Please remove the extra line space. I prefer it the way it was.

TFT/src/User/Menu/StatusScreen.c
```diff
@@ -290,8 +300,8 @@ void menuStatus(void)
       drawStatusScreenMsg();
 
     scrollMsg();
-    key_num = menuKeyGetValue();
```
guruathwal, line 303: Please revert this change. I prefer it the way it was.

guruathwal (changes requested):

TFT/src/User/Menu/StatusScreen.c
digant73: ok, no problem to revert those although they have been made just to be compliant to other files (Printing menu etc...) not reviewed by me

digant73:
> "Show ACK in Terminal" is not a UI setting. The same goes for the file sorting option.

As UI settings I considered any setting affecting how an info is displayed and/or if it is displayed or not.
Show Ack (as persistent temp etc...) enable a info to be displayed.
For the sorting option it seems to me a UI. It determines how the file list must be displayed. It has no impact on printing

Just few words about the General settings. They are settings across all the other categories Printer/Machine, Marlin, TFT or external devices. Emulated M600 (such as the Event LED feature; when it will be available) are features provided/emulated by the TFT. For that reason I didn't move them on Printer/Machine settings

guruathwal:
> As UI settings I considered any setting affecting how an info is displayed and/or if it is displayed or not.
> Show Ack (as persistent temp etc...) enables an info to be displayed.
> For the sorting option it seems to me a UI. It determines how the file list must be displayed. It has no impact on printing

With this all of the settings are UI settings.

digant73:
they are 8 while the printer params are alot more and most of them are not even configurable at runtime. The feature menu was a mix of settings with the UI settings on top so if I would change for example the runout type, cancel gcode you needed to scroll 4-5 pages. Usually the UI are set once and not changed anymore.
Also, now that the Printing menu offers the clickable icons. the More menu could be changed removing the Heat, Fan, Speed etc... that are already available in the Printing menu. Also the Babystep menu could be replaced. Maybe It could be replaced by the Feature menu or Extrude menu. The latter will allow to extrude filament in case of pause or emulated M600. Also the More menu provides access to full menu such as Machine, but at the end only the LED menu can be used during a print. IMHO it should be better to provide in the More menu only the menus usable during a print. As far a I see, from the Machine menu, only LED and Terminal (in case the print is paused) can be used they during a print

guruathwal:
> Also, now that the Printing menu offers the clickable icons. the More menu could be changed removing the Heat, Fan, Speed etc... that are now already available in the Printing menu. 

IMHO those icons should not be removed so early.

> Also the Babystep menu could be replaced. Maybe It could be replaced by the Feature menu allowing to change on the fly printing settings such as Event LED, M600, runout, auto power off, cancel gcode etc... Also the More menu provides access to full menu such as Machine, but at the end only the LED and Terminal (in case the print is paused) can be used during a print. IMHO it should be better to provide in the More menu only the menus usable during a print. The Screen settings could be provided in the More menu (with the exception of Recalibrate, all the features in Screen menu have no negative impact on the print so they can be changed at runtime).
> A possible layout is to replace the Babystep (in Printing menu) with Feature menu, while the More menu could provide Extrude, (Un)Load, Screen, LED, Terminal. Less menus reachable with less touches

The feature settings icon is already in a good place within the more menu. There is no reason to show feature settings on top of the printing screen. The babystep icon should be visible on the printing screen as the preview icon is also displayed on it. 

In my next PR regarding the icon menu layout, the limit of 8 icons will be removed giving the ability to have any number of icons. and giving more control over menu layout. So the issue regarding icon space like this will be eliminated.

digant73:
that will be a good new feature. Do you mean there will be a sort of customizable layout or scrollable windows?

guruathwal:
yes, there will be both.

Hitamku:
@digant73 I tried to compile it but after I tried it hit the screen in the fan section it appeared CtI and CtS. Which part of the code should I change to get rid of it? BIGTREE_TFT35_B1_V3_0 | bugfix-2021-09-16

digant73:
this PR was already merged on BTT main branch. You simply need to download the current BTT fw not the fw from my repo

PR #2111, opened 2021-08-07
Fix typo on status_chamber.bmp

rebel1:
This fixing 2 typing errors in The Round Miracle Theme for the status_chamber.bmp


digant73:
it was already reported as fixed in #2105

rebel1:
Ok,
I did not see this.

PR #2117, opened 2021-08-09
fix filament runout sensor bug

Msq001:
### Requirements

### Description
* Default use `FIL_RUNOUT_PIN` port when multiple extruder and only one filament sensor port. 
* Add the feature of configuring normal open or normal closed of filament sensor port

### DIFF
--- a/TFT/src/User/API/Settings.h
+++ b/TFT/src/User/API/Settings.h
@@ -177,6 +177,7 @@ typedef struct
 
   // Filament Runout Settings (only if connected to TFT controller)
   uint8_t  runout;
+  uint8_t  runout_nc;
   uint8_t  runout_invert;
   uint16_t runout_noise_ms;
   uint8_t  runout_distance;
--- a/TFT/src/User/API/Settings.h
+++ b/TFT/src/User/API/Settings.h
@@ -177,8 +177,8 @@ typedef struct
 
   // Filament Runout Settings (only if connected to TFT controller)
   uint8_t  runout;
-  uint8_t  runout_nc;
   uint8_t  runout_invert;
+  uint8_t  runout_nc;
   uint16_t runout_noise_ms;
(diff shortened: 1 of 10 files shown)


kisslorand:
> * Add the feature of configuring normal open or normal closed of filament sensor port

How is this different from the logic invert?

TFT/src/User/API/Settings.h
```diff
@@ -177,6 +177,7 @@ typedef struct
 
   // Filament Runout Settings (only if connected to TFT controller)
   uint8_t  runout;
+  uint8_t  runout_nc;
```
digant73, line 180: it should be provided after runout_invert

Msq001:
> How is this different from the logic invert?

logic invert is used to set whether high level or low level represents filament runout
runout_nc is used to set the default pull-up or pull-down of the input pin. This is useless for the three pins filament switch sensor, because the three wire always connects the signal line with 3.3V or GND, so it doesn't matter whether it is pull-up or down.
But it is very useful for two pins filament switch sensor because it only controls the on/off with 3.3V or GND.  An additional setting is needed to tell the MCU whether to pull-up or pull-down when the filament switch sensor is cut off and IO is floating.

sarvenn:
How can we decide which setting should we use for runout_nc parameter? My invert logic was set to 1. Should I try two options for runout_nc and see which one works fine or is there a way to find the correct setting without trial?

bigtreetech:
> How can we decide which setting should we use for runout_nc parameter? My invert logic was set to 1. Should I try two options for runout_nc and see which one works fine or is there a way to find the correct setting without trial?

Maybe the name of the configuration needs to be clearer? `invert logic` is changed to the `runout level`, `runout_nc` is changed to the `pull-up/down` setting? Can this reduce confusion?

PR #2122, opened 2021-08-13
RRF: Update fan, extrusion and print speed factors from M408 response

pfn:
### Requirements

No new requirements

### Description

* Update loopFan and loopSpeed to only update a new setting once
 * prevents overriding other control sources
 * replace lastSetSpeed with needSetSpeed boolean
* Don't call M220 M221 on RRF in speed menu
* Handles JSON parsing if `inJson` is true and on RRF
* Skips sending M220 and M221 for speed queries if on RRF
* code error: eAxisBackup not declared extern in coordinates.h, fixed
* code error: missing `_` in `#define` for RRFMacrosParser.hpp
* remove use of `current_key` in json parser, memory bug

### Benefits

* No longer requires `T0` in `config.g`
* Uses a single `M408 S0` query, periodically, to read3 parameters (fan speed, extrusion factor and speed factor)
* Eliminates unnecessary `M220` and `M221` queries when on RRF

### Related Issues

N/A

### Testing

This PR has been tested and validated to work on my MKSTFT28.

### PR Status

Ready to merge, pending any reviews

### Tech Debt

`fanQuery` should be renamed into a general `rrfStatusQuery` function in an appropriate source file.


oldman4U:
This is great. 

Would you be available for more RRF related contributions?

TFT/src/User/API/parseACK.c
```diff
@@ -1,4 +1,5 @@
 #include "parseACK.h"
+#include "parseACKJson.hpp"
```
digant73, line 2: please move this specific include as the last one. Also verify that no tabs and/or trailing spaces are present in all your changes

pfn: I'll make that change soon, probably Monday. There should be no tabs or trailing spaces.

pfn:
> This is great.
> 
> Would you be available for more RRF related contributions?

@oldman4U I primarily work on things that are pain points in my use of the screen. If you have any ideas that align with things that irritate me, I would be happy to investigate if it's somewhat easily doable!

oldman4U:
Hi. 

Currently there is no developer available here using RRF but one enthusiastic user - bttguy. So every help is very welcome.

bttguy:
Nice @pfn! Thank you!

PR #2126, opened 2021-08-16
Add remaining time based on comments from gCode file if generated with Cura or IdeaMaker

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR adds the remaining time info in the printing screen if the gCode is generated with Cura or IdeaMaker. There's no need anymore to use a plugin to insert M73 commands. In case M73 Rxx is present in the gCode, the comments are ignored, M73 has priority.

### Benefits

Everyone using Cura and/or IdeaMaker.

### Note

If there's any other slicer that inserts comments into the gCode file related to total time, elapsed or remaining time please let me know so I can implement that slicer's info too.


kisslorand:
IdeaMaker supprt added

Issue #1856, opened 2021-04-19
PID Autotune not working on E3-TFT35 with E3-RRF in RepRap Mode 


Wombat37:
### Description

I've already posted this in the E3-RRF repository:

https://github.com/bigtreetech/BTT-E3-RRF/issues/9

Someone suggested that it belongs better here.

### Steps to reproduce

Run PID autotune for the hotend or bed from the TFT keypad. Appears to go into autotune mode and then just sits there doing nothing (nothing heats up) until it times out

Works OK when sending M303 g code

**Expected behavior**
Should perform PID autotunes

**Actual behavior**
Does nothing until it times out

### Hardware Variant
TFT35v3.0 and E3-RRFv1.1  
### TFT Firmware Version & Main Board Firmware details
3.0.27 (RRF 3.2.2)

radek8:
The problem you mentioned is that the FW TFT is programmed primarily for Marlin.
RRF support has only recently begun to be implemented.
Marlin and RRF use different parameters and therefore the command does not work.
There are very few programmers who know RRF, commands and their parameters in RRF.
:-(

radek8:
You could try to resolve the issue by editing the Configuration.h file and recompiling.
![image](https://user-images.githubusercontent.com/46979052/115751136-29b53f80-a399-11eb-814e-506feb70b908.png)

oldman4U:
Wombat37

Have you tried what radek8 recommended to do?

Would be nice if you could answer your own ticket.

Wombat37:
My apologies, I know it's important (and courteous) to respond to suggestions like this and I will do so very soon. 

In the meantime I was trying to resolved why my prints were so much worse with the new E3-RRF board. I've just put a temperature probe on the nozzle and it looks like it's 24°C lower than the 220°C setpoint (I've tried two different probes and thermometers) - which may explain things. I'm currently tweaking the M308 L and H settings to try and fix this.

Sorry for the distraction - will try out radek8's suggestions tonight. I'm a little hesitant about recompiling as I've not done it before and I'm worried that I'll break other stuff :)

Wombat37:
Ok, I'm lost. I'm not a programmer although I can do some programming but this system is confusing me lots.

Which Configuration.h file do I need to modify? I know this is used in Marlin but I'm using RepRap firmware and I don't see this file on the BBT site. Do I need to download a version of RRF firmware from somewhere else and edit and compile that?

[...]

Issue #1791, opened 2021-04-03
[FR] RRF improvements


bttguy:
The latest implemented changes made the BTT TFT really useful for RRF users.
Here I request some additional features in a hope that @bigtreetech or other developers would help to improve the useability of the screen.

Original FR: #1400 (now closed, as the core support is already there)
Discussion for RRF: #1763

Requested features:
•	Make the PID tune menu useable - kinda working, just in a wired way, that need some work
•	Macro menu icon view support

Known limitations:
- Thumbnails from onboard SD cant be displayed

Any RRF users is welcomed to suggest FR's here!

Thanks in advance to anyone who helps! :)

PaulG4H:
Macro was working with RRF until this commit I asume:
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/1769

If I compare the macro and rrfmacro files there is a huge difference...

bttguy:
Yeah, sadly something we've wrong lately with the macros, I opened an issue for this bug in #1814

bttguy:
Thanks for #1842 :)

bigtreetech:
Hello. 

`* Menu system for macros + "interactive" macros`
What does "interactive" macro mean? How does it work? Does the current macro menu not meet the requirements?

`* Thumbnail and menu system support for onboard gcodes`
This feature needs to be discussed, because the current thumbnail is to add bitmap data to gcode,  then the TFT parses it. The difficulty lies in
1. The gcode of onboard sd is parsed by the motherboard, so the TFT can't get the data in gcode. Need the motherboard firmware to send the bitmap data to the TFT.
2. TFT communicates with motherboard by UART, the amount of bitmap data is relatively large, and the transmission through UART is very slow. The amount of data will be much smaller if the thumbnail is compressed into JPEG format, But It's very difficult to parse JPEG based on TFT MCU's performance.

`* Make the load/unload menu useable`
`Movement->Extrude` is available after #1909. “(Un)Load” menu is called "M701" gcode to load filament, I'm not sure if RRF support similar gcode. Anyway `Movement->Extrude` is universal and available extrusion menu

`* Make the PID tune menu useable`
The same is already available after #1856

jaysuk:
All gcodes available in RRF are described here
https://duet3d.dozuki.com/Wiki/Gcode#Section_M701_Load_filament
As you can see, M701 is available

bigtreetech:
[...]

PR #2134, opened 2021-08-19
RRF: Support M291 dialogs and interactive macros

pfn:
### Requirements

No new requirements

### Description

Enables support for M291 dialogs when initiated outside of the terminal menu (DWC, macros, printing gcode, etc.). S0, S1, S2 and S3 are supported modes (although for S0, the OK button is always shown). X/Y/Z arguments are not supported.

* Add RRFStatusControl.c, tracks state of RRF and launches M408 requests
* Refactor `fanQuery` into `rrfStatusQuery()`
* Update `request_M98` macro execution to use RRFStatusControl
* Block sending `M27` and `M408` while inside of the terminal menu
* Process `M291` information from `M408` responses
* Add status call `rrfStatusIsMacroBusy`
* Prevent `M27`, `M105` and `M155` from running while macro busy
* Skip waiting for `ok` response for `M98`
* Skip waiting for `ok` response on `M292` and `M408` while macro busy
* Remove prompt to start a macro. Macros will run on click.
  * dangerous macros should have their own M291 prompt
* Updates the running macro display

### Benefits

* M291 dialog support
* Interactive macros

### Testing

This has been tested and verified to work on my MKS TFT28

### PR Status

This is ready to merge pending any reviews.


pfn:
Oh, duh. M98 is a blocking gcode.

digant73:
it's better you rename parseAckJson to RRFParseAckJson as done for all the other RRF specific files.
Also move RRFMacrosParser to folder API instead of Menu. It's not a menu

pfn:
> it's better you rename parseAckJson to RRFParseAckJson as done for all the other RRF specific files.
> Also move RRFMacrosParser to folder API instead of Menu. It's not a menu

done @digant73

TFT/src/User/API/RRFStatusControl.c
```diff
@@ -0,0 +1,71 @@
...
+void rrfStatusSet(char status)
+{
+  rrf_status = status;
+  if (status != 'B')
+  {
+    macro_busy = false;
+  }
+}
+
+void rrfStatusSetBusy(void)
```
digant73, line 24: make all of those functions made by only one line as staic inline. E.g. "static inline void rrFStatusSet etc...) it is more efficient and uses les flash space (in that case)

TFT/src/User/API/parseACK.c
```diff
@@ -1,6 +1,6 @@
 #include "parseACK.h"
 #include "includes.h"
-#include "parseACKJson.hpp"
+#include "RRFParseACK.hpp"
```
digant73, line 3: move this include in inludes.h

digant73 (changes requested):

digant73:
As you already are doing try to put all the logic related to RRF on functions (with rrf prefix). Provide those functions on specific RRF files (with RRF prefix). Always try to not change the existing code to adapt it to RRF but try to engage the rrf functions with all the specif logic. When the functions are called only once or the function are made be just one easy instruction, define them as "static inline" (it will uses less flash). When possible put RRF logic/functions as last choice (not first)

digant73:
In RRFMacros.c I would also move #include "RRFMacrosParser.hpp" to includes.h

bttguy:
This interactive macro support sounds great! Thanks a lot @pfn  ! :)

TFT/src/User/API/parseACK.c
pfn: I'm not a fan of this change, I feel that scope should be minimized, and only parseACK.c should have access to anything in RRFParseACK.

pfn:
> In RRFMacros.c I would also move #include "RRFMacrosParser.hpp" to includes.h

@digant73 similar to what I said for RRFParseACK, I believe the same goes here. Only RRFMacros should have access to RRFMacrosParser.

TFT/src/User/API/parseACK.c
digant73: no problem if the scope will be limited only on parseAck

digant73:
@pfn if possible have a look to #1856. It seams PID is not pefectly working yet

pfn:
> @pfn if possible have a look to #1856. It seams PID is not pefectly working yet

@digant73 I'll give it a look if I have some time

bttguy:
@pfn Is this PR adds support also for "menu/folder stucture" for macros? Right now if I categorize them on DWC, they only show up as a list on the TFT without any folders.

pfn:
> @pfn Is this PR adds support also for "menu/folder stucture" for macros? Right now if I categorize them on DWC, they only show up as a list on the TFT without any folders.

@bttguy no, there is no change in that area. However, folders have been supported for quite some time now. I believe it was added by @Msq001 a while back.

bttguy:
Hmm, I will check it and update the FR-s accordingly. I just checked on the RRF RF I created in #1791 if there is anything I can add/update. :)

PR #2136, opened 2021-08-22
RRF: Remove rrf_macros_on configuration, set automatically

pfn:
### Requirements

No new requirements

### Description

Remove `rrf_macros_on` configuration key from config.ini. Instead, automatically set based on `infoMachineSettings.firmwareType`

### Benefits

* Removes code that is essentially dead. No RRF user will be choosing the `Custom` menu instead of `Macros`
* Reduces the delta in config between standard and RRF configurations.

### Testing

Verified to work on my MKSTFT28

### PR Status

Ready to merge pending reviews.

### Related Issues

n/a


bttguy:
Nice! Make sense, can agree as an active RRF user! Keep it up @pfn :D

PR #2138, opened 2021-08-24
Reduce variables in settings and remove some duplicate code

guruathwal:
- Reduce variables in settings and remove some duplicate code.
  - Combine runout sensor setting variables.
  - Combine sound alert setting variables.
  - Combine send printing gcode setting variables.
- Add bitwise macros.
- Rename **SmartFeatures.h** to **AddonHardware.h**.

### DIFF
--- a/TFT/src/User/API/MachineParameters.c
+++ b/TFT/src/User/API/MachineParameters.c
@@ -173,7 +173,7 @@ uint8_t getEnabledElement(PARAMETER_NAME name, uint8_t index)
 
 static inline void setParameterStatus(PARAMETER_NAME name, bool status)
 {
-  SET_BIT_VALUE(parametersEnabled,name, status);
+  SET_BIT_VALUE(parametersEnabled, name, status);
 }
 
 static inline uint8_t getParameterStatus(PARAMETER_NAME name)
--- a/TFT/src/User/API/MachineParameters.c
+++ b/TFT/src/User/API/MachineParameters.c
@@ -140,18 +140,18 @@ const LABEL autoRetractDisplayID[] = {LABEL_RETRACT_AUTO};
 
 static inline void setElementStatus(PARAMETER_NAME name, uint8_t element, bool status)
 {
-  elementEnabled[name] = (status == 1) ? (1 << element) | elementEnabled[name] : elementEnabled[name] & (~(1 << element));
+  SET_BIT_VALUE(elementEnabled[name], element, status);
 }
 
 static inline uint8_t getElementStatus(PARAMETER_NAME name, uint8_t element)
 {
-  return (elementEnabled[name] >> element) & 1;
+  return GET_BIT(elementEnabled[name], element);
 }
 
 uint8_t getEnabledElementCount(PARAMETER_NAME name)
 {
   uint8_t count = 0;
-  for (uint8_t i = 0; i < parameterElementCount[name]; i++) { count += (elementEnabled[name] >> i) & 1; }
+  for (uint8_t i = 0; i < parameterElementCount[name]; i++) { count += GET_BIT(elementEnabled[name], i); }
   return count;
 }
 
@@ -162,7 +162,7 @@ uint8_t getEnabledElement(PARAMETER_NAME name, uint8_t index)
 
   for (uint8_t i = 0; i < parameterElementCount[name]; i++)
   {
-    state = (elementEnabled[name] >> i) & 1;
+    state = GET_BIT(elementEnabled[name], i);
     count += state;
 
     if (state && count == (index + 1))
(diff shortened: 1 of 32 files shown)


TFT/src/User/includes.h
```diff
@@ -91,7 +91,7 @@
 #include "ScreenShot.h"
 #include "SerialConnection.h"
 #include "Settings.h"
-#include "SmartFeatures.h"
+#include "AddonHardware.h"
```
digant73, line 94: should be better to put it on alfabetic order.
Also isn't it better to call invert_axis as inverted_axis?

TFT/src/User/API/MachineParameters.c
```diff
@@ -173,18 +173,18 @@ uint8_t getEnabledElement(PARAMETER_NAME name, uint8_t index)
 
 static inline void setParameterStatus(PARAMETER_NAME name, bool status)
 {
-  parametersEnabled = (status == 1) ? (1 << name) | parametersEnabled : parametersEnabled & (~(1 << name));
+  SET_BIT_VALUE(parametersEnabled,name, status);
```
digant73, line 176: , name

TFT/src/User/API/Settings.c
```diff
@@ -16,6 +16,8 @@ const uint16_t default_pause_speed[]   = {NOZZLE_PAUSE_XY_FEEDRATE, NOZZLE_PAUSE
 const uint16_t default_preheat_ext[]   = PREHEAT_HOTEND;
 const uint16_t default_preheat_bed[]   = PREHEAT_BED;
 const uint8_t default_custom_enabled[] = CUSTOM_GCODE_ENABLED;
+const uint8_t default_sounds           = 0b00001111;  // all sounds enabled
+const uint8_t default_invertAxis       = 0b00000000;  // all invert axis disabled
```
digant73, line 20: shouldn't be better inverted_axis?

TFT/src/User/includes.h
guruathwal: invert_axis is ok. No reason to increase the name length when it already gives the correct meaning.

digant73: I mean also the meaning of the data. With "invert" do you mean the verb "to invert"? The data structures should not report a verb. The verb is used in function names

TFT/src/User/API/Printing.c
```diff
@@ -356,7 +356,7 @@ void printStart(FIL * file, uint32_t size)
       infoPrinting.file = *file;
       infoPrinting.cur = infoPrinting.file.fptr;
 
-      if (infoSettings.send_start_gcode == 1 && infoPrinting.cur == 0)  // PLR continue printing, CAN NOT use start gcode
+      if (GET_BIT(infoSettings.send_gcodes, SEND_GCODES_END_PRINT))  // PLR continue printing, CAN NOT use start gcode
```
digant73, line 359: maybe SEND_GCODES_START_PRINT?

TFT/src/User/API/Printing.c
```diff
@@ -390,10 +390,8 @@ void printEnd(void)
   setPrintRemainingTime(0);
   preparePrintSummary();  // update print summary. infoPrinting are used
 
-  if (infoSettings.send_end_gcode == 1)
-  {
+  if (GET_BIT(infoSettings.send_gcodes, SEND_GCODES_START_PRINT))
```
digant73, line 393: maybe SEND_GCODES_END_PRINT?

TFT/src/User/includes.h
guruathwal: With invert_axis here means an instruction or order. Here it is still a variable and is used as an array of bits. If I had swapped the words, then it can be like axis_inverted.

TFT/src/User/API/Printing.c
guruathwal: done

TFT/src/User/includes.h
digant73: yes, so my interpretation was correct (it is an action or order). I would try to not use an order for the data naming, so I would use axis_inverted (although I thought that the more appropriate english syntax was inverted_axis). I have already made some cleanup on my next PR. E.g. swapping Emulate M600 to Emulated M600. When possible, all the conf. params should not provide an order/action in the naming (it's not used in italian language but I think also in english language).

TFT/src/User/API/config.c
```diff
@@ -903,15 +903,13 @@ void parseConfigKey(uint16_t index)
     #ifdef FIL_RUNOUT_PIN
       case C_INDEX_RUNOUT:
         if (inLimit(config_int(), 0, 3))
-          infoSettings.runout = config_int();
+          SET_BIT_VALUE(infoSettings.runout, RUNOUT_ENABLED, GET_BIT(config_int(), RUNOUT_ENABLED));
```
digant73, line 906: do you need { } after "if"?

TFT/src/User/API/config.c
```diff
@@ -903,15 +903,13 @@ void parseConfigKey(uint16_t index)
...
+          SET_BIT_VALUE(infoSettings.runout, RUNOUT_SENSOR_TYPE, GET_BIT(config_int(), RUNOUT_SENSOR_TYPE));
         break;
 
       case C_INDEX_RUNOUT_LOGIC:
-        infoSettings.runout_invert = getOnOff();
-        break;
-
       case C_INDEX_RUNOUT_NC:
-        infoSettings.runout_nc = getOnOff();
+          SET_BIT_VALUE(infoSettings.runout, (c_index - C_INDEX_RUNOUT_LOGIC), getOnOff());
```
digant73, line 912: wrong indentation

guruathwal: fixed

TFT/src/User/API/config.c
guruathwal: Yes, I need it. Thanks for providing it. 😉

TFT/src/User/includes.h
guruathwal: invert_axis, inverted_axis, axis_inverted all mean the same thing. I would not bother about renaming the variables as long as the names reflect their purpose. Renaming existing variables also makes it difficult to track changes across a long range of commits.

PR #2140, opened 2021-08-26
RRF: print menu improvements

pfn:
### Requirements

No new requirements

### Description

* Print menu button launches directly into onboard storage (for RRF users)
  * USB and SD card menus are disabled
* Sorts file list alphabetically
* Fixes file list parsing so that file names don't get mangled
* Allows directory navigation
* Refactors RRFMacroParser to RRFM20Parser
  * request_M20_macros refactored to request_M20_rrf
* Removed leftover rrf_macros_enable setting
* Query M552 on start to populate IP address (when booting after board)

### Benefits

Enhances the print menu experience for RRF users

### PR Status

Ready to merge pending reviews

### Testing

Works as intended on my MKSTFT28


pfn:
> Why on earth would you disable the onboard sd and usb printing options?
> 
> ok i can understand the issue with printing directly from usb / sd. but the correct implementation would be
> 
> [specifically for RRF] browse files on tft media select file for printing upload file to main board instruct firmware to print from main board file
> 
> This approach would mimic the workflow using DWC.

Uart is not supported for file uploads. So yes, disabling usb and SD is what is necessary. What is described above is not a viable solution.

pfn:
It's so simple. Thanks. Please do it. Much obliged.

PR #2146, opened 2021-08-30
Implemented #2132 + config.ini cleanup + bugfix/code optimization

### LINKED ISSUE #2132: BLT Z offset wizzard with homing without probe
The TFT BLtouch offset wizard currently only works with the USE_PROBE_FOR_Z_HOMING Marlin configuration.
Please add BLT offset configuration support when homing is with limiters, not probe.

Details BIGTREE TFT35_V3.0_E3.27 BLT Z offset wizzard doesn't work with homing without probe
#2121

digant73:
**IMPROVEMENTS:**
* Implemented #2132: Probe offset now supports also printers configured to home usning min endstop instead of a probe. The previous param "xy_offset_probing" was renamed to "probing_z_offset" to support the new feature
* Optimized LevelCorner menu: Some code reduction and code sharing with LevelingControl API
* Cleanup on config.ini and Configuration.h: Reviewed some param names and comments reflecting better their use in the code. default values in Configuration.h are mapped in 1-to-1 with the related param in config.ini
  Scope of those changes are also to force some people to finally update their config.ini starting from the most updated config.ini provided here instead of simply integrating new params in their old config.ini. This should avoid (I hope) to see so many bug reports referring to missing or old comments present in the config.ini file reporting wrong (old) values range etc...
  Below a list of the renamed params in config.ini:
  emulate_m600 -> emulated_m600
  title_back_color -> title_background_color
  background_color -> menu_background_color
  font_color -> menu_font_color
  reminder_color -> reminder_font_color
  volume_status_color -> status_font_color
  fan_speed_percent -> fan_speed_percentage
  marlin_bg_color -> marlin_background_color
  marlin_fn_color -> marlin_font_color
  onboard_sd_support -> onboard_sd
  long_filename_support -> long_filename
  invert_axis -> inverted_axis
  xy_offset_probing -> probing_z_offset
  z_raise_probing -> probing_z_raise
  preheat_nameX -> preheat_name_X
  ps_on -> ps_auto_shutdown
  ps_on_active_high -> ps_active_high
  auto_shutdown_temp -> ps_auto_shutdown_temp
  fil_runout_inverting -> fil_runout_inverted
  fil_noise_threshold -> fil_runout_noise_threshold
  pl_recovery_en -> pl_recovery
  pl_z_raise -> pl_recovery_z_raise
* Improved and optimized code for parsing queue buffer (in parseAck.c and config.c): A check on the queue length instead of accessing the data is performed. This speedup performance and avoid possible out of buffer access in case the buffer is corrupted (e.g. due to noise)
* Internal cleanup: all the attributes in "infoSettings" follow the namning convention based on lower case and "_" as separator

**BUGFIXES:**
* Fixed wrong usage of auto shutdown temp: default value was used instead of the value configured in "ps_auto_shutdown_temp" in config.ini
* minor bugfixes

resolves #2132


oldman4U:
👍🏻 like always!!!

TFT/src/User/API/LevelingControl.h
```diff
@@ -12,20 +12,31 @@ extern "C" {
...
+
+extern LEVELING_POINT probedPoint;  // last probed point or LEVEL_NO_POINT in case of no new updates
+extern float probedZ;               // last Z offset measured by probe
+
+void levelingGetPointCoords(LEVELING_POINT_COORDS coords);   // get all point coords
+LEVELING_POINT levelingGetPoint(int16_t x, int16_t y);       // get point matching XY coords or LEVEL_CENTER in case of no match
+void levelingMoveToPoint(LEVELING_POINT point);              // move to point
+void levelingProbePoint(LEVELING_POINT point);               // probe point
+void levelingSetProbedPoint(int16_t x, int16_t y, float z);  // set probed point and Z offset for point matching XY coords
+inline LEVELING_POINT levelingGetProbedPoint(void)
```
kisslorand, line 34: "inline" in ".h" files? Why?

digant73: it is simply a way to access a variable meant to be hidden (if it could be declared in .c) through the API function instead of accessing it directly. It was done in the past also in parseAck with isHostDialog. The data needs to be defined as "extern" in .h but the intent is to use the API function to access it instead of accessing it directly.
You could not declare and use externally an inline function if it is defined in .c

kisslorand: > You cannot define and use externally an inline function if it is declared in .c (the scope is inside .c).

I think you can, just try it.

digant73: "extern" is the opposite of "static" and you cannot use a static inline function outside the .c.

kisslorand: My bad, I ignored that they are declared "static".

digant73: no problem

bttguy:
@digant73 Can we include the fix for #1856 to this PR?

digant73:
@bttguy it's better it is tested on RRF. Unfortunately I don't have RRF in any of my printers.
I think it's better to ask @pfn to include it and test it in a further PR

pfn:
> @digant73 Can we include the fix for #1856 to this PR?

@bttguy my general preference is to keep PRs as focused and concise as possible.

bttguy:
@pfn  If you can include it later on, I agree also. I just mentioned it because digant was involved in the conversation regarding PID, and he mentioned that the timeout could be increased.

pfn:
```
Below a list of the renamed params in config.ini:emulate_m600 -> emulated_m600title_back_color -> title_background_colorbackground_color -> menu_background_colorfont_color -> menu_font_colorreminder_color -> reminder_font_colorvolume_status_color -> status_font_colorfan_speed_percent -> fan_speed_percentagemarlin_bg_color -> marlin_background_colormarlin_fn_color -> marlin_font_coloronboard_sd_support -> onboard_sdlong_filename_support -> long_filenameinvert_axis -> inverted_axisxy_offset_probing -> probing_z_offsetz_raise_probing -> probing_z_raisepreheat_nameX -> preheat_name_Xps_on_active_high -> ps_active_highps_on -> ps_auto_shutdownauto_shutdown_temp -> ps_auto_shutdown_tempfil_runout_inverting -> fil_runout_invertedfil_noise_threshold -> fil_runout_noise_thresholdpl_recovery_en -> pl_recoverypl_z_raise -> pl_recovery_z_raise
```

I preferred the old emulate_m600 name, imo it should be a present tense action for key names.

Also, this is pretty cumbersome for users as anyone with a custom configuration will have to go through and make sure everything is correctly modified. This sort of mass update would be nice if there were a clean way to migrate users' config.

digant73:
But it's also a feature like "terminal ACK", "notification M117" etc... which you decide to enable/disable in the Feature/UI menu. Also the verb can create confusion in some cases (e.g. invert axis means more that you have the possibility to change the printer axis direction while it is used by the TFT as a reference to bind the "move" buttons in the right way.
Believe me, it is more cumbersome to see always the same bug reports and reply with always the same answers. In most of the cases obsolete config.ini files are used reporting the old description and value ranges that have been updated almost 10 months ago

TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -236,7 +236,7 @@ static inline void reDrawPrintingValue(uint8_t icon_pos, uint8_t draw_type)
       case ICON_POS_Z:
         if (layerDisplayType == SHOW_LAYER_BOTH)
         {
-          sprintf(tempstr, "%6.2fmm", (infoFile.source >= BOARD_SD) ? coordinateGetAxisActual(Z_AXIS) : coordinateGetAxisTarget(Z_AXIS));
+          sprintf(tempstr, "%3.2fmm", (infoFile.source >= BOARD_SD) ? coordinateGetAxisActual(Z_AXIS) : coordinateGetAxisTarget(Z_AXIS));
```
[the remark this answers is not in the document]
kisslorand: It's me who put %6 there. "%6.2f" means **_at least_** 6 characters before the decimal point. I did it so the displayed text to be centered in the display area, it is not a mistake. 
This is for the lower row of the layer display area. 
For the upper row I put %3 because of the different behaviour of the used method to display text. A left aligned method  is used there. 3 digits before the decimal point is enough, I do not think there's a 3D printer that can print above 999.99mm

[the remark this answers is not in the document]
digant73: Yes I know but the size of tempstr is 10 chars. 6.2mm + '\0'gives 12 :-) so something should be removed (e.g. mm). No problem to revert my change but consider my suggestion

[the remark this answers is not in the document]
kisslorand: For some reason unknown to me it is working just fine the way it is.

PR #2147, opened 2021-08-30
Reduce global variable in Serial connection settings

guruathwal:
- Reduce RAM usage in Serial connection settings.
- Non-editable variables are moved to program memory. 
- Switch Boolean arrays to bit arrays.


TFT/src/User/API/SerialConnection.c
```diff
@@ -5,24 +5,72 @@
...
+  #endif
+  #ifdef SERIAL_PORT_5
+    {SERIAL_PORT_5, SERIAL_PORT_5_QUEUE_SIZE, "5", "5 - UART5"}
+  #endif
+  #ifdef SERIAL_PORT_6
+    {SERIAL_PORT_6, SERIAL_PORT_6_QUEUE_SIZE, "6", "6 - UART6"}
+  #endif
+};
+
+const uint8_t serialPortIndex[_UART_CNT] = {
```
digant73, line 32: it's not correct because it depends on the TFT variants. For TFT28 index 0 could be UART_1 while for TFT35 index 0 could be UART_2 etc... For that reason I needed a not constant array and the value was detected and set in Serial_Init() with the instruction:
serialPortIndex[serialPort[i].port] = i;  // 1 for SERIAL_PORT_2 etc...

For backward compatibility I also maintained the "active" flag that for me it is useless once you have the possibility to disable the unused ports (so the check on active ports is not performed at all)

digant73 (changes requested):

TFT/src/User/API/SerialConnection.h
```diff
@@ -7,8 +7,8 @@ extern "C" {
 
 #include <stdbool.h>
 #include <stdint.h>
-#include "variants.h"  // for SERIAL_PORT_2 etc...
-#include "uart.h"      // for _UART_CNT etc...
+#include "variants.h"  // for SERIAL_PORT_2 etc.
```
digant73, line 10: a very minor comment. In many languages etc is followed by ... (it was not an error to provide ... instead of .)

digant73:
@guruathwal please have a look to my last commit related to a bug in your last PR. The params runoutXX, soundsXX and gcodeXX were not parsed properly

TFT/src/User/API/SerialConnection.h
guruathwal: I am no one to speak about other languages. At least in the English language, as far as I know, adding an ellipsis ('...') after 'etc' is wrong because it is an abbreviation. 'etc' should always end with a dot ('.'). Ellipsis is used when you don't want to complete a sentence or a list. 'etc.' itself is used to end a list.

TFT/src/User/API/SerialConnection.c
guruathwal: solved

guruathwal:
> @guruathwal please have a look to my last commit related to a bug in your last PR. The params runoutXX, soundsXX and gcodeXX were not parsed properly

Yes I saw it.

digant73:
but getSerialID() is more inefficient than before in terms of processing (it always scans the ports array) and also the flash usage (-4 bytes of RAM but more prog FLASH) .

digant73:
I didn't take into account all the 6 possible uart ports handled by the Serial/UART HAL because the maximum number of available ports is 4 (there is no TFT variant using more than 4 ports).
If you extend SerialConnection to support up to 6 ports you should also extend the size of serial_port array in Settings.h/c and Config.c but that means a waste of 2 ports/bytes that currently will be never used by any TFT variant

guruathwal:
> I didn't take into account all the 6 possible uart ports handled by the Serial/UART HAL because the maximum number of available ports is 4 (there is no TFT variant using more than 4 ports).
> If you extend SerialConnection to support up to 6 ports you should also extend the size of serial_port array in Settings.h/c and Config.c but that means a waste of 2 ports/bytes that currently will be never used by any TFT variant

I did not intend to do it either. But in your last PR, you took the max serial ports possible `_UART_CNT` which expanded to 6.
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/e590d96dac9ca368b322d6edffb16403457a7e04/TFT/src/User/API/SerialConnection.c#L22
I just tried to match it. It seems this is not required anymore.

digant73:
@guruathwal ok I understood the misunderstanding. For serialPortIndex I needed to use the UART count (so 6) just because the TFT variants could use a mix of all the six UART ports. So serialPortIndex array needed all 6 UART ports (e.g. in the TFT variant we have  SERIAL_PORT_2  _USART5) to retrieve the index on serialPort array having a max size of 4 ports

Issue #1901, opened 2021-05-05
Re-order file list with newest files at top


Wombat37:
**Is your feature request related to a problem? Please describe.**
My BBT-E3-RRF motherboard SD card must kept in position when using RepRap firmware. 

Files to print are loaded onto this card and, over a period of time, the list of files will occupy multiple pages on the TTF print file selection page. The newest file (i.e. the most likely one to print will be at the bottom of this list and then several pages may have to be scrolled to get to it. See the image below:

![Print_File](https://user-images.githubusercontent.com/42827647/117197819-4482ad80-adb6-11eb-84d7-102abe20861b.png)

 
**Describe the solution you'd like**
I would like to see the order of the listed files to be reversed, so that the newest file would be at the top of the list and be immediately accessible. This is how it worked on the original Ender 3 board.
 
**Describe alternatives you've considered**
One solution would be for me to delete older files and keep the list short - but what's the point of providing a list like this if you can't keep much in it? 

**Additional context**
Not a serious problem but tedious and inconvenient.

kisslorand:
As far as I know the list is given to the TFT by the mainboard. So, it's not a TFT issue.

Wombat37:
Thanks for the clarification.

I seem to be making a habit of posting to the wrong site. If I post in the mainboard site, they tell me I should be posting here.

This seems to be a strange demarcation considering both products are made by the same company.

RotorVulura:
Hi,
If you are using tft sd slot there is not any solution at the moment (as I can see, but I could be wrong).SD write works in blocks. The last file added will goes at the very first free blocks (usually at bottom of the list). But, If you delete some old file (from the top of the list and/or in the middle), new files may be goes to fill free blocks creating a new fantastic "random" list. End of the story.
If you print from mainboard sd slot, you can tell Marlin in configuration.adv.h for sorting files in alphabetical order or the newest at the top of the list.
The best solution, for now, is keeping files subdivided in folders. Also, there is a very nice btt tft plugin for cura slicer that shows a preview of the .stl directly on the screen. Browsing per image instead name is more nice and less boring:)
Cheers

oldman4U:
It is not the hardware but the software which provides this functionality. In both cases it is not BTT. BTT provides "only" the hardware.

kisslorand:
@RotorVulura 
[...]

PR #2150, opened 2021-08-30
RRF: enable A-Z, Z-A, oldest and newest sort in print menu

### LINKED ISSUE #1981: [FR] Files in icon mode in RRF
With the new soon to be merged thumbnail support, I would like to ask some rework/rethinking of the file menu for in RRF.
Currently only list mode is supported, it can't be turned off/switched to icon mode. 
We discussed earlier this, but maybe this new thumbnail PR could help with speed. Even there would be some latency to load thumbnails, I and possibly many others would use thumbnails.

pfn:
### Requirements

No new requirements

### Description

* Adds `stream_handler` to requestCommandInfo to support >5K payloads
* Adjusts scope of parser/handler in RRFParseACK
* Unlocks icon/grid mode print job listing, fixes #1981
* Resets macro menu cur_page to 0 upon entry (fixes state between print
  and macro menus)
* Code in place to support `M20 S3` once the kinks are worked out

### Benefits

Improves RRF support

### Related Issues

Addresses #1981 request for print menu icon mode

### Testing

Verified working on my MKSTFT28

### PR Status

Ready to merge pending reviews.

### Help Wanted

It seems I've made some memory allocation mistakes when parsing the JSON for `M20 S3` -- it would be greatly appreciated if someone could throw some eyeballs at my changes to see where I goofed? I'm thinking I have some problems around allocating/deallocating `file_name` and `display_name` but can't come to terms with what is happening.

### DIFF
--- a/TFT/src/User/API/RRFParseACK.cpp
+++ b/TFT/src/User/API/RRFParseACK.cpp
@@ -39,10 +39,6 @@ static int8_t m291_mode = 0;
 static bool show_m291 = false;
 static uint32_t m291_seq = 0xffffffff;
 static uint32_t expire_time = 0;
-static bool need_parser_reset = false;
-
-static JsonStreamingParser parser;
-static ParseACKJsonParser handler;
 
 static void m291_confirm(void)
 {
@@ -154,11 +150,14 @@ void ParseACKJsonParser::value(const char *value)
 
 void rrfParseACK(const char *data)
 {
+  static JsonStreamingParser parser;
+  static ParseACKJsonParser handler;
+
   parser.setListener(&handler);
-  if (need_parser_reset)
+  if (handler.need_parser_reset)
   {
     parser.reset();
-    need_parser_reset = false;
+    handler.need_parser_reset = false;
   }
   while (*data != 0)
   {
--- a/TFT/src/User/API/RRFParseACK.cpp
+++ b/TFT/src/User/API/RRFParseACK.cpp
@@ -150,17 +150,16 @@ void ParseACKJsonParser::value(const char *value)
 
 void rrfParseACK(const char *data)
 {
-  static JsonStreamingParser parser;
   static ParseACKJsonParser handler;
 
-  parser.setListener(&handler);
-  if (handler.need_parser_reset)
-  {
-    parser.reset();
-    handler.need_parser_reset = false;
-  }
+  jsonStreamingParser.setListener(&handler);
   while (*data != 0)
   {
-    parser.parse(*data++);
+    jsonStreamingParser.parse(*data++);
+  }
+  if (handler.need_parser_reset)
+  {
+    jsonStreamingParser.reset();
+    handler.need_parser_reset = false;
   }
 }
--- a/TFT/src/Libraries/json/JsonStreamingParser.cpp
+++ b/TFT/src/Libraries/json/JsonStreamingParser.cpp
@@ -26,6 +26,8 @@ See more at http://blog.squix.ch and https://github.com/squix78/json-streaming-p
 #include "JsonStreamingParser.hpp"
 #include <string.h>
 
+JsonStreamingParser jsonStreamingParser;
+
 JsonStreamingParser::JsonStreamingParser() {
     reset();
 }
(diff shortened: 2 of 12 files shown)


bttguy:
@pfn Appreciate your contributions! Thank you!

pfn:
Hmm, it looks like it's overflowing user_heap_space on 48K ram devices... I don't know what to do with this...

bttguy:
Which part of/function you added is too large?

pfn:
> Which part of/function you added is too large, and by how much? I mean I see the bytes, but what does it mean? Would it be a sollution to go trough the function of the TFT and disable some of it which is totally or practically useless for RRF users, to enable the RRF related features that would really make a difference for us? What do you think @pfn?

@bttguy I fixed this by removing `M20 S3` support.

bttguy:
> > Which part of/function you added is too large, and by how much? I mean I see the bytes, but what does it mean? Would it be a sollution to go trough the function of the TFT and disable some of it which is totally or practically useless for RRF users, to enable the RRF related features that would really make a difference for us? What do you think @pfn?
> 
> @bttguy I fixed this by removing `M20 S3` support.

I see. And what do you think in general about in the future (with @bigtreetech help), is it possible from your coder's viewpoint to go trough and disable functions so save us RAM, without messing with Marlin users?

pfn:
> I see. So the sorting wont work for now? And what do you think in general about in the future (with @bigtreetech help), is it possible from your coder's viewpoint to go trough and disable functions so save us RAM, without messing with Marlin users?

@bttguy sorting works, but it relies on the implicit oldest-first ordering returned by `M20` -- it's not guaranteed to always be oldest first. Being able to use `M20 S3` means that timestamps can be read and explicitly ordered and there's no reliance on "undocumented behavior"

tl;dr -- it works now, maybe not in the future, depending on how RRF changes.

bttguy:
> > I see. So the sorting wont work for now? And what do you think in general about in the future (with @bigtreetech help), is it possible from your coder's viewpoint to go trough and disable functions so save us RAM, without messing with Marlin users?
> 
> @bttguy sorting works, but it relies on the implicit oldest-first ordering returned by `M20` -- it's not guaranteed to always be oldest first. Being able to use `M20 S3` means that timestamps can be read and explicitly ordered and there's no reliance on "undocumented behavior"
> 
> tl;dr -- it works now, maybe not in the future, depending on how/if RRF changes over time.

Still better than nothing, will do for a while. Thanks for the hard work and bugfix! :) Gonna test soon how reliable it is.

TFT/src/User/API/RRFParseACK.cpp
```diff
@@ -154,11 +150,14 @@ void ParseACKJsonParser::value(const char *value)
 
 void rrfParseACK(const char *data)
 {
+  static JsonStreamingParser parser;
+  static ParseACKJsonParser handler;
```
digant73, line 154: I see you defined multiple instances for parser and handler. Isn't it possible to share them across more files?

pfn: There's really only once instead of each handler in global memory (static) one ParseACKJsonParser and one RRFM20Parser. So there's nothing to share there. At some point, I will probably change all usages of RRM20Parser to be a streaming variant, and then it will be shared instance. (stack vs. heap)

The problem with sharing JsonStreamingParser is that it has to maintain current document state (start/end, etc.). Thinking about it, it doesn't seem like it would be reentrant as there is only 1 active document at any given time, and it's either a something being parsed by parseACK or being passed onto cmd_rev_buf/stream_handler. I'll probably make that change shortly.

pfn:
Looks like the suggestion from @digant73 helped free up a lot of memory to support timestamps, placeholder code for supporting `M20 S3` has been restored.

Wombat37:
I'm really excited about this PR.

Printing from the onboard SD card has been a real pain from the TFT screen because I have to scroll through a 10-page list just to get to the file I've just uploaded!

I requested such a change in Issue #1901 and was told that this was impossible to achieve in the TFT firmware and had to be implemented in the RRF code and I was asked to close the issue!

Thank you so much for finding a possible solution to this - it really is much appreciated!

Subcircuit:
@pfn Hi, are you handling M20 S3?

pfn:
> @pfn Hi, are you handling M20 S3?

@Subcircuit yes, but that code path is currently disabled due to memory issues I can't figure out yet.

PR #2152, opened 2021-09-01
Bugfix - shadow characters in Layer-field

### LINKED ISSUE #2144: Shadow characters in Layer-field
TFT

### Description

When you printed a part with layer hight more than 10mm, and after this you start a new print, then the field where the actual layer ist displayed, still has the last (longer string) in background.

Its no such big bug. See it as feedback to enhance your software :-) 

### Hardware Variant

BIGTREETECH_TFT35_V3.0

TFT35_V3.0.27.x Aug 22 2021
Marlin 2.0.9.1

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR adds more variables to be initialized before the start of a print.

### Related Issues

Fixes #2144
Adds a missing print loop before displaying the layer number.

### Notes

~~This PR is marked as draft because I am on vacation, far away from any 3D printer and have no way to test it. I hope someone will have the good will to test it and report back.~~


oldman4U:
Thank you and enjoy your vacation.

PR #2156, opened 2021-09-03
Optimized parsing in interfaceCmd.c and parseACK.c + configurable M109/M190

### LINKED ISSUE #1974: [Q] Can we now officially remove the M109 / M190 redirections?
I have seen that the M109 / M190 have been redirected to M104 / M140 respectively.

The original issue of not waiting for that process to finish seems to have been corrected.

I, personally, have removed the redirection for now by commenting out these lines:
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/TFT/src/User/API/interfaceCmd.c#L638
and
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/TFT/src/User/API/interfaceCmd.c#L708

Testing it seems to be running just fine and temperature LED transitions work just fine.
Can we revert this redirection? Or is there another fix/angle being worked on still?

### LINKED ISSUE #2151: [FR] Scrolling with the encoder
Navigation in the files or in the list settings require multiple taps with a lot of option/gcode/macros.

Would it be possible to use the encoder to scroll up and down/left or right in grid view/ as an option that can be configured in the UI settings/config.ini(to let people chose to turn on)? This would save a lot of time, and prevent unnecessary taps on the screen.

**Additional context**
This would be useful for all user, regadless the firmware.

### LINKED ISSUE #2153: [Q] Thumbnail Sanity Check compilation problem.
Hi everybody,

i'm currently trying to compile the last commit of the firmware.
i encounter some problem with Sanitycheck compiling directive stops.

When i put 0 in the THUMBNAIL_PARSER in the configuration.hn the compilation is OK.

When i put 2 (Base 64 PNG for PrusaSlicer), the compilation stops and indicate the following error :

```
TFT\src\User/SanityCheck.h:301:6: error: #error "Configuration error: THUMBNAIL_PARSER is not set to a valid value of 0, 1 or 2."
     #error "Configuration error: THUMBNAIL_PARSER is not set to a valid value of 0, 1 or 2."
```

Do you have any advice ?

Thanks a lot !!

(i m French so my english could be approx ^^)

digant73:
**IMPROVEMENTS:**
* Optimized parsing code in interfaceCmd: Reduced the use of array indexing and used length of the key for a fast check on the buffer and matching/not matching conditions
* Optimized parsing code in parseAck: Reduced conditions in the parsing functions to avoid as much as possible continue checks on  the validity of the array elements (e.g. != 0). Used length L2 cache and key for a fast check on matching/not matching conditions. Major benefits are present when searching a key in the middle of L2 cache
* Code reduction: Due to optimizations above
* Configurable feature "emulated M109/M190": New param "emulated_m109_m190" added in config.ini to enable/disable  M109/M190 emulated with M104/M140
* Added rotary encoder support on List menu (list/icon menu): Rotary encoder can be used to navigate (page up/down) on any List menu without the need to use the page up/down buttons.

  **NOTE:** As for all the other menus supporting rotary encoder, the button on rotary encoder MUST/CAN NOT be used to apply an action on a list item (e.g. select the file to print). That's due to the HW shared with mainboard (Rotary encoder is physically connected to the mainboard and any press on rotary button will have effects also on Marlin menu).
To apply an action on the item, you MUST continue to apply a touch on the screen
* Minor cleanup in config.ini: Forgot to apply the following param rename in my prevoius PR:
  rotate_ui -> rotated_ui
  block_touch_on_idle -> lcd_lock_on_idle
* Minor cleanup

**BUGFIXES / FR:**
fixes #2153
resolves #1974
resolves #2151


Copy to SD Card root directory to update/config.ini
```diff
@@ -102,6 +102,16 @@ serial_port:P1:6 P2:0 P3:0 P4:0
 #   Options: [disable: 0, enable: 1]
 emulated_m600:1
 
+#### Emulated M109 And M190
+# The TFT intercepts the blocking M109 and M190 gcodes (set target hotend and bed temperatures)
+# and converts them to the non blocking M104 and M140 gcodes respectively.
+#
+# NOTE: Enable it to allow the TFT can continue to dialog (e.g. using menus sending gcodes to the
```
kisslorand, line 109: "to allow the TFT can continue" - typo?

digant73: hmm no. it should be correct. Do you mean wrong syntax?

kisslorand: Hard to explain, the whole sentence sounds wrong. My English is not that bad but it's hard for me to understand it (ex. "to dialog" is not a verb in English). Only by checking the code I have some idea what it supposed to mean.
IDK, maybe "Enable it so the TFT can still communicate with Marlin firmware even if the target temperature is not reached yet."

kisslorand: Later edit:
Maybe add: "otherwise communication will be frozen until desired/set temps are obtained."

digant73: yes I agree, thanks. Using your comments

bttguy:
Wow @digant73 you already implemented the encoder scroll?

digant73:
rotary encoder is now supported in both List/Icon modes

bttguy:
> rotary encoder is now supported in both List/Icon modes

Awesome! Huge thanks to you!!

Tannoo:
In 'config.c':
```
    case C_INDEX_EMULATED_M600:
    case C_INDEX_EMULATED_M109_M190:
      SET_BIT_VALUE(infoSettings.general_settings, (index - C_INDEX_EMULATED_M600), getOnOff());
      break;
```
Is this intentional or is something missing?

I am trying to implement this now and this doesn't make sense.

digant73:
yes, it is correct

Issue #1979, opened 2021-06-08
[FR] M117 support for RRF


bttguy:
Just checked how's M117 working with RRF, and it needs a little optimization. It would be a nice feature for RRF users too #1791.
Te current behavior after I enabled M117 at config.ini:
M117 sent from TFT terminal -> notification pop up on TFT and on DWC
M117 sent from DWC console -> notification pop up only on DWC
I would like to implement the feature, when M117 sent from DWC it would also pop up on the TFT as a notification. Maybe it could also help in macros where a massege should pop up/user action is required.

jaysuk:
does M118 behave the same way?
https://duet3d.dozuki.com/Wiki/Gcode#Section_M118_Send_Message_to_Specific_Target

bttguy:
@jaysuk Thats what my second thought was, after work I will check it again with M118. What P parameter should I use, 0/2?

jaysuk:
might as well try both and see what happens.
hopefully both will work, if not as long as one of them works we can at
least note that as a current restriction

guruathwal:
M117 is a one way message. It will only be displayed on the LCD or DWC So this will be useless if it did not pass through the TFT. With M118, the message should be relayed to all the connected hosts as an echo message if the message does not contain any of the keywords which are suppressed by the TFT. You should check to which ports/targets the M118 message is relayed without the P parameter.

digant73:
On Marlin, M117 seems forwarded to the TFT as a notification. E.g. the "M117 Time Left xxx" is received as "action:notification Time Left xxx". I Don't know if a specific setting has to be enabled in RRF to provide this kind of feature.

jaysuk:
As far as I understand it, M117 does it automatically if a TFT is
configured.
The next step would be to see what is sent.

bttguy:
> M117 is a one way message. It will only be displayed on the LCD or DWC So this will be useless if it did not pass through the TFT. With M118, the message should be relayed to all the connected hosts as an echo message if the message does not contain any of the keywords which are suppressed by the TFT. You should check to which ports/targets the M118 message is relayed without the P parameter.

It seems like M118 cant be executed without a parameter.
![image](https://user-images.githubusercontent.com/76131447/121747817-4273e300-cb08-11eb-935d-1fb681426527.png)
I tested with all the possible values, but none ofe them was work as I wanted.

Wombat37:
[...]

PR #2171, opened 2021-09-11
RRF:M575 no longer requires RRF compatible marlin

Subcircuit:
### Requirements

Configure SD card config.g file  M575 P1 S1 

### Description

Make the TFT no longer require RRF compatible marlin on the basis of @pfn work. 
Use the M408 S0 to parse most of the information, including temperature, coordinates, and heater status.
synchronize the TFT with the web state.
Separate rrf messages instead of processing them all in parseACK.c,

### Benefits

Reduce unnecessary uart communication,
parseACK.c is not too confusing

### DIFF
--- a/TFT/src/User/API/RRFParseACK.cpp
+++ b/TFT/src/User/API/RRFParseACK.cpp
@@ -233,9 +233,12 @@ void ParseACKJsonParser::value(const char *value)
 
       break;
     case result:
-        strcpy(infoFile.title, value);
-        setPrintHost(true);
-        // setPrintResume(false);
+        if (starting_print)
+        {
+          strcpy(infoFile.title, value);
+          setPrintHost(true);
+          starting_print = false;
+        }
       break;
     case none:
       break;
--- a/TFT/src/User/API/RRFParseACK.cpp
+++ b/TFT/src/User/API/RRFParseACK.cpp
@@ -110,7 +110,7 @@ void ParseACKJsonParser::value(const char *value)
       rrfStatusSet(value[0]);
       break;
     case heaters:
-      if (index == 0 )
+      if (index == 0)
       {
         heatSetCurrentTemp(BED, strtod((char *)value, NULL) + 0.5f);
       }
@@ -120,7 +120,7 @@ void ParseACKJsonParser::value(const char *value)
       }
       break;
     case active:
-      if (index == 0 )
+      if (index == 0)
       {
         heatSyncTargetTemp(BED, strtod((char *)value, NULL) + 0.5f);
       }
@@ -128,6 +128,8 @@ void ParseACKJsonParser::value(const char *value)
       {
         heatSyncTargetTemp(index - 1, strtod((char *)value, NULL) + 0.5f);
       }
+      break;
+    case standby:
       break;
     case hstat:
       if (strtod((char *)value, NULL) == 3)
@@ -159,11 +161,15 @@ void ParseACKJsonParser::value(const char *value)
       break;
     case tool:
       break;
+    case probe:
+      break;
     case fan_percent:
       if (index != 0 && index <= infoSettings.fan_count) // index 0 is an alias for default tool fan
       {
         fanSetPercent(index - 1, strtod((char *)value, NULL) + 0.5f);
       }
+      break;
+    case fanRPM:
       break;
     case mbox_seq:
       seq = strtod((char *)value, NULL);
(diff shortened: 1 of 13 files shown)


TFT/src/User/API/RRFParseACK.cpp
```diff
@@ -42,7 +42,7 @@ static uint32_t expire_time = 0;
 
 static void m291_confirm(void)
 {
-  if (m291_mode > 1) mustStoreCmd("M292 P0\n");
+  if (m291_mode >= 1) mustStoreCmd("M292 P0\n");
```
pfn, line 45: Nice, I remember messing this up but forgetting to fix it

bigtreetech:
@bttguy  Hello, Can you help test this PR? The purpose of this PR is to directly communicate with RRF firmware in JSON format instead of RRF Marlin command mode. This can achieve more functions

bttguy:
@bigtreetech Will test it in the evening! Very welcomed changes.

TFT/src/User/API/parseACK.c
```diff
@@ -662,21 +668,12 @@ void parseACK(void)
...
         char * fileEndString;
-        if (infoMachineSettings.firmwareType != FW_REPRAPFW)
-        {
-          // Marlin
-          fileEndString = " Size:";  // File opened: 1A29A~1.GCO Size: 6974
-        }
-        else
-        {
-          // RRF
-          ack_seen("result\":\"0:/gcodes/");  // {"key":"job.file.fileName","flags": "","result":"0:/gcodes/pig-4H.gcode"}
```
pfn, line 677: It looks like this was removed without a corresponding handler to set the job file name?

Can we create a new parser to handle this message and set the print job name?

TFT/src/User/API/RRFStatusControl.c
```diff
@@ -30,6 +30,8 @@ void rrfStatusSet(char status)
           case 'I':
             // parseACK will take care of going to the print screen
             mustStoreCmd("M409 K\"job.file.fileName\"\n");
+            setPrintHost(true);
```
pfn, line 33: This should probably be deferred until the response of M409 before we start the print status display?

I would think along the lines of a new function in `gcode.c` like `rrf_request_job_name(...)` that would invoke its own parser and launch the job

pfn (changes requested):

bttguy:
@bigtreetech from an user opinion, thighs work like before this PR. With pfn-s coding cleanups is gonna be a good change to use and make use of S1 communication mode! Thanks a lot @Subcircuit

TFT/src/User/API/parseACK.c
Subcircuit: > It looks like this was removed without a corresponding handler to set the job file name?
> 
> Can we create a new parser to handle this message and set the print job name?

My mistake :P  Looks like I'll have to put it in `resq`.

Subcircuit:
@pfn  Thanks for your advice I have improved.

Subcircuit:
> @bigtreetech from an user opinion, things work like before this PR. With pfn-s coding cleanups is gonna be a good change to use and make use of S1 communication mode! Thanks a lot @Subcircuit

Thanks just a small improvement

TFT/src/User/API/RRFParseACK.cpp
```diff
@@ -141,6 +190,53 @@ void ParseACKJsonParser::value(const char *value)
...
+      {
+        pidUpdateStatus(true);
+      }
+      else if (strstr(value, (char *)"Error: M303") || (strstr(value, (char *)"Auto tune of heater") && strstr(value, (char *)"failed")))
+      {
+        pidUpdateStatus(false);
+      }
+
+      break;
+    case result:
```
pfn, line 235: This alone is kind of dangerous, it would be good to add a new variable into RRFStatusControl: `starting_print`, set it to `true` at `mustStoreCmd("M409 K\"job.file.fileName\"\n");` and set it `false` here. This case should only run if `starting_print`.

Create new set/is functions as needed

TFT/src/User/API/RRFParseACK.cpp
```diff
@@ -141,6 +190,53 @@ void ParseACKJsonParser::value(const char *value)
...
+        strcpy(infoFile.title, value);
+        setPrintHost(true);
+        // setPrintResume(false);
```
pfn, line 238: I think the setPrintResume is here only to ensure the pause state is removed if any, doesn't hurt to leave it, I believe.

pfn (changes requested):
Thanks for the other updates @Subcircuit -- just one more minor fix and I think this looks good  :)

bttguy:
@bigtreetech This PR looks great! Merge it, so we can build other inprovements for RRF on top of that more easily :)

pfn (approved):
fair enough, possibly consider renaming it to `rrf_starting_print`

Subcircuit:
> fair enough, possibly consider renaming it to `rrf_starting_print`

 All right

by the way, when I was testing the file list sorting feature(M20 S3), I found that Json would freeze when parsing more files (about 6 files or more) and display "loding...".  Even if I increase the dmaL2Cache space to 4096

pfn:
> > fair enough, possibly consider renaming it to `rrf_starting_print`
> 
> All right
> 
> by the way, when I was testing the file list sorting feature(M20 S3), I found that Json would freeze when parsing more files (about 6 files or more) and display "loding...". Even if I increase the dmaL2Cache space to 4096

there's some serious memory issues I haven't been able to solve yet; happens more often on M20 S3 mode for me (which I've disabled) -- the problem isn't dmaL2Cache, I'm thinking there's some memory leak happening as a result of saving and copying the filename/display_name stuff

TFT/src/User/API/parseACK.c
```diff
@@ -353,11 +353,9 @@ void hostActionCommands(void)
...
       setPrintResume(true);
 
-      if (infoMachineSettings.firmwareType != FW_REPRAPFW)
-      {
-        hostAction.prompt_show = false;
-        Serial_Puts(SERIAL_PORT, "M876 S0\n");  // auto-respond to a prompt request that is not shown on the TFT
-      }
+      hostAction.prompt_show = false;
+      Serial_Puts(SERIAL_PORT, "M876 S0\n");  // auto-respond to a prompt request that is not shown on the TFT
+
```
digant73, line 358: remove this empty line

Subcircuit:
Oh do you have a solution yet?  When I have time try to solve it

pfn:
> Oh do you have a solution yet? When I have time try to solve it

I haven't been able to solve it yet @Subcircuit , I would appreciate the help. I'm not working on solving it any time soon.

Subcircuit:
> > Oh do you have a solution yet? When I have time try to solve it
> 
> I haven't been able to solve it yet @Subcircuit , I would appreciate the help. I'm not working on solving it any time soon.

You're welcome. I'll still find time to study it.

digant73:
@Subcircuit if your PR implements #1979 then mark it as resolved

pfn:
> @Subcircuit if your PR implements #1979 then mark it as resolved

@digant73 it does not yet, easy to implement once this PR is merged, but that's additional work that isn't in this PR

PR #2179, opened 2021-09-16
New Listening Mode + some bug fixes and cleanup

digant73:
**IMPROVEMENTS:**
* Added TFT Listening Mode: The TFT simply listen for ACK messages (e.g. spontaneous) while all the outgoing messages are purged.
This mode is useful on main boards sharing the bus with TFT and USB (e.g. MKS GEN L) allowing Octoprint/Pronterface to properly control the printer (via USB) avoiding command collisions (otherwise present with TFT in normal mode).
The TFT will receive the ACK messages (spontaneous and/or related to gcodes from USB port) displaying on the screen all the status info.
The mode can be activated/deactivated on "Connection" menu. Existing labels (ON/OFF) and icon have been used to avoid to add new ones

**NOTE:** The mode state is stored on FLASH allowing to restore the mode at startup. This allows to power on/off the printer remotely and to control the printer via USB without the need of any touch (from the user) on the display to engage the mode
* Moved some functions from commnet.c/h to Printing.c/h API
* Minor code reduction
* Removed some global variables
* Minor cleanup

**BUG FIXES:**
* Fixed a bug preventing the TFT to reconnect to printer when switching from Marlin mode to Touch mode

Connection_1

Connection_2

Ready_1


Copy to SD Card root directory to update/config.ini
```diff
@@ -824,16 +824,16 @@ end_gcode_enabled:0
 cancel_gcode_enabled:0
 
 #### Start Gcode
-# This gcode will runs before starting a print if `start_gcode_enabled` is enabled.
+# This gcode will runs before starting a print if "start_gcode_enabled" is enabled.
```
kisslorand, line 827: ~runs~ - run

Copy to SD Card root directory to update/config.ini
```diff
@@ -824,16 +824,16 @@ end_gcode_enabled:0
...
 #   Value range: [min: 3, max: 75 characters]
 start_gcode:G28 XY R20\n
 
 #### End Gcode
-# This gcode will runs after a print is completed if `end_gcode_enabled` is enabled.
+# This gcode will runs after a print is completed if "end_gcode_enabled" is enabled.
```
kisslorand, line 832: ~runs~ - run

Copy to SD Card root directory to update/config.ini
```diff
@@ -824,16 +824,16 @@ end_gcode_enabled:0
...
 #   Value range: [min: 3, max: 75 characters]
 end_gcode:M104 S0\nM140 S0\nM107\nM18\n
 
 #### Cancel Gcode
-# This gcode will runs when a print is canceled if `cancel_gcode_enabled` is enabled.
+# This gcode will runs when a print is canceled if "cancel_gcode_enabled" is enabled.
```
kisslorand, line 837: ~runs~ - run

Copy to SD Card root directory to update/config_rrf.ini
```diff
@@ -819,16 +819,16 @@ end_gcode_enabled:0
 cancel_gcode_enabled:0
 
 #### Start Gcode
-# This gcode will runs before starting a print if `start_gcode_enabled` is enabled.
+# This gcode will runs before starting a print if "start_gcode_enabled" is enabled.
```
kisslorand, line 822: ~runs~ - run

Copy to SD Card root directory to update/config_rrf.ini
```diff
@@ -819,16 +819,16 @@ end_gcode_enabled:0
...
 #   Value range: [min: 3, max: 75 characters]
 start_gcode:G28 XY R20\n
 
 #### End Gcode
-# This gcode will runs after a print is completed if `end_gcode_enabled` is enabled.
+# This gcode will runs after a print is completed if "end_gcode_enabled" is enabled.
```
kisslorand, line 827: ~runs~ - run

Copy to SD Card root directory to update/config_rrf.ini
```diff
@@ -819,16 +819,16 @@ end_gcode_enabled:0
...
 #   Value range: [min: 3, max: 75 characters]
 end_gcode:M104 S0\nM140 S0\nM107\nM18\n
 
 #### Cancel Gcode
-# This gcode will runs when a print is canceled if `cancel_gcode_enabled` is enabled.
+# This gcode will runs when a print is canceled if "cancel_gcode_enabled" is enabled.
```
kisslorand, line 832: ~runs~ - run

TFT/src/User/API/Settings.c
```diff
@@ -214,34 +221,29 @@ void setupMachine(void)
...
     infoMachineSettings.longFilename = infoSettings.long_filename;
 
-  mustStoreCmd("M503 S0\n");
+  if (infoMachineSettings.firmwareType == FW_SMOOTHIEWARE)  // Smoothieware does not report detailed M115 capabilities
+  {
+    infoMachineSettings.leveling        = BL_ABL;
+    infoMachineSettings.emergencyParser = ENABLED;
+  }
 
   if (infoMachineSettings.firmwareType == FW_REPRAPFW)
```
kisslorand, line 238: I would use "else if" for redundant "if" statements

TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -187,12 +187,10 @@ void menuBeforePrinting(void)
   // initialize things before the print starts
   progDisplayType = infoSettings.prog_disp_type;
   layerDisplayType = infoSettings.layer_disp_type * 2;
-  setLayerNumber(0);
```
kisslorand, line 190: This was a part of a bugfix. Why do you removed it?

TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -187,12 +187,10 @@ void menuBeforePrinting(void)
...
   coordinateSetAxisActual(Z_AXIS, 0);
   coordinateSetAxisTarget(Z_AXIS, 0);
   setM73_presence(false);
-  setTotalTime(0);
```
kisslorand, line 194: Why did you remove this?

kisslorand:
I do not agree with the things moved out from comment.c/h. In my book it is not an optimization creating new procedures/functions and calling them instead of direct accessing a variable. 
Those variables were used only in conjunction with comment info so I still think they belong to where they were.

Copy to SD Card root directory to update/config_rrf.ini
digant73: yes

TFT/src/User/API/Settings.c
digant73: yes

TFT/src/User/Menu/PrintingMenu.c
digant73: because "layer" and "total time" were moved to "infoPrinting" data struct that is initialized to 0 in printStart() function

TFT/src/User/Menu/PrintingMenu.c
digant73: see comment above

digant73:
@kisslorand but they are printing information used by Printing menu that doesn't need to know/interface with other modules (comment). Layer, time could be set by other modules/gcode while Printing menu simply retrieve and display them

kisslorand:
Now that you removed M73 Rxx check, what happens if it is present?

digant73:
it was always set to "false" in the code unless there is a case where it is set to "true" that I didn't see

TFT/src/User/API/interfaceCmd.c
```diff
@@ -609,12 +618,11 @@ void sendQueueCmd(void)
           if (cmd_seen('R'))
           {
             setPrintRemainingTime((cmd_value() * 60));
-            setM73_presence(0);  // disable parsing remaning time from gCode comments
```
kisslorand, line 612: Oopsie! Here should be `setM73_presence(true)`.
I do not know what happened here because I never use numeric values for boolean, always `true` or `false`.

kisslorand: > it was always set to "false" in the code unless there is a case where it is set to "true" that I didn't see

See my comment in "interface.c" line 612. You just discovered a bug that I cannot explain how it slipped there.

digant73: ok. reverted and fixed

kisslorand: Thank you. Cheers!

digant73:
@guruathwal testing this PR I found a bug on Terminal. It seems that some texts will cause an infinite loop on the "while loop" (at line 787 in Terminal.c). The menuTerminalWindow menu will start an infinite page update (from 1 to 20) with no text displayed on screen.
I found that the following text is one causing the infinite loop:
//action:notification Bed Cooling...

The loop is due to the presence of both "//" and ":". With theese characters, the exit condition on the loop is never met (or probably something inside the loop broke the code (e.g. getCharacterInfo)).

When you have some time, could you please have a look to that issue? it seems quite important.

Another minor bug I saw scrolling up on the window is that the ack messages (that were dispalyed with white color) are then displayed as a request command (orange color).

guruathwal:
@digant73  Thank you for notifying me. I will look into it now.

gordo3di:
I've been having random communication timeouts when using Repetier server connected via USB. Do you think this issue is related?

digant73:
@gordo3di no, I think it's not related to that PR. This PR simply purges (skips) all the gcodes that should be sent to the printer

PR #2194, opened 2021-09-20
Fix - layer number display glitch

### LINKED ISSUE #2184: [Q] Layer Number
Hi. I have an Artillery Genius (TFT28). I'm not able to show the layer number on the print screen, only the height is shown. Do I need to use some specific gcode between layers? PS: I'm using PrusaSlicer. Tks!

kisslorand:
### Requirements

BTT or MKS TFT.

### Description

This little PR fixes a GUI glitch with Layer Number display during print. It occurred when the total layer number was unknown and only the printing layer number  was displayed. For the first 10 layers unwanted or extra characters were also displayed. This could also occur if the printing project contained multiple objects and the printing order was object by object and not layer by layer.

"Config.ini" is updated with info for PrusaSlicer users to have layer number info displayed on the TFT during print.

### Related Issues

Fixes #2184

### DIFF
--- a/Copy to SD Card root directory to update/config.ini
+++ b/Copy to SD Card root directory to update/config.ini
@@ -272,6 +272,9 @@ prog_disp_type:2
 #     also shown, it can be done by introducing a comment at the
 #     beginning of the gCode file in the format "; Layer count: xx".
 #     Separators can be " ", ":", "_" or "=".
+#   - If the total number of layers exceeds 999, this information
+#     will not be displayed because there's no room for both current
+#     and total layer number to be shown.
 #
 #    For PrusaSlicer users, if the layer number/count display
 # feature is wanted, some comment lines must be added in 
--- a/Copy to SD Card root directory to update/config.ini
+++ b/Copy to SD Card root directory to update/config.ini
@@ -273,6 +273,12 @@ prog_disp_type:2
 #     beginning of the gCode file in the format "; Layer count: xx".
 #     Separators can be " ", ":", "_" or "=".
 #
+#    For PrusaSlicer users, if the layer number/count display
+# feature is wanted, some comment lines must be added in 
+# Printer Settings -> Custom G-code section such as:
+#    - ";LAYER:[layer_num]" in "After layer change G-code"
+#    - ";LAYER_COUNT:[total_layer_count]" in "Start G-code"
+#
 #   Options: [Layer height: 0, Layer number: 1, Both - height & number: 2]
 layer_disp_type:0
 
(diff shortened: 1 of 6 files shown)


oldman4U:
👍🏻

kisslorand:
Got some more work to do for object to object printing and prints above 999 layers, hence marked as draft.

kisslorand:
If layer count is above 999, this info is not displayed. I had no other choice, there's no room on the icon for such a long string (ex "2475/3729"). I thought about a toggle between layer number and layer count but I think it would be an unnecessary stress for the MCU during print.

oldman4U:
Current/Remaining?

kisslorand:
Note: Max 8 characters can be displayed (at least that's the limit on the displays I have for testing).

oldman4U:
In case you can change the position of the slash once one value changes the number of digits 8 should be more than sufficient I believe.

kisslorand:
Note2: In case the actual layer number is greater than 999 and the remaining layer is greater than 999 (obviously the total layer number is also greater than 999) than whatever combinations one tries, there are already 8 digits to show, no room for any other character to separate the two values.
It is not that uncommon to have more than 2000 layers, at 0,1mm layer height that's only a 20cm tall print.

oldman4U:
In case the print has 1999 or more layers, only the current layer is shown.

oldman4U:
Or you show a max of 999 for the remaining layers which should give you a max of 9999 current layers.

radek8:
What would it look like if the / character were not used to separate these two values but were distinguished by color?
Would the current value be white and the total number of layers be dark red, for example?

kisslorand:
The idea itself is very interesting but I am afraid it is not possible to do it in a very simple manner and since we are in the middle of a print I would avoid at all cost more stress to the TFT's MCU.

radek8:
I don't want to switch the value, but display these two numbers in a row, each in a different font color.
Would it be understandable when reading the display?

radek8:
It is true that 95% of prints have less than 999 layers.
And if it will be over a thousand, I can do without the value of the total number of layers :-)

kisslorand:
> It is true that 95% of prints have less than 999 layers.
> And if it will be over a thousand, I can do without the value of the total number of layers :-)

That was my thought also! 
Easy on the MCU and probably those more than 999 layer prints are done over night without anyone watching the TFT.

radek8:
> 
> I tested something like "1452/2+K" but it looks really ugly and not intuitive at all

And what about this, it wouldn't look so bad. :-)
1452/>2k

oldman4U:
xK+

PR #2197, opened 2021-09-22
Removed unnecessary functions - size reduction

kisslorand:
### Description

Removed some unnecessary functions resulting in size reduction and  less stress (theoretically) on the MCU during print.

### DIFF
--- a/TFT/src/User/API/AddonHardware.c
+++ b/TFT/src/User/API/AddonHardware.c
@@ -236,14 +236,14 @@ void FIL_BE_CheckRunout(void)
   if (!(GET_BIT(infoSettings.runout, 0)))  // Filament runout turn off
     return;
 
-  setPrintRunout(FIL_IsRunout());  // Need constant scanning to filter interference
+  infoPrinting.runout = FIL_IsRunout();  // Need constant scanning to filter interference
 }
 
 void FIL_FE_CheckRunout(void)
 {
   static uint32_t nextReminderTime = 0;
 
-  if (!getPrintRunout() && !getRunoutAlarm())
+  if (!infoPrinting.runout && !getRunoutAlarm())
     return;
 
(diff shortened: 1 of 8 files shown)


digant73:
@kisslorand we spent a lot of time creating API avoiding direct access to data structures in order to avoid any kind of inconsistencies (bugs). In that way you are rolling back all the work we started more than 1 year ago. It's not a call to a function having significative improvement on performance. The engines are far better/optimized (also in terms of performance) than it was in the past and more easy to debug and fix residual bugs

kisslorand:
Here we go again...
I am sorry that you feel I am doing harm to the code.

digant73:
As clarified by chat, not knowing your history of the project in the last 1,5 years, I briefly (I'm working now) reported the history/explanation for some changes in the code. I know a "raw" code could slightly improve performance and reduce the size but at the same time it was also one of the reason for so much bugs requiring so much effort to identify and fix. The code is public and that's another point to make the code as much as robust and easy to understand and use following some best practice.
There is nothing personal, I simply reported my compromises/vision based on the scope of the project (it's an open source project), maintenance, scalability (distinguish legacy modules by extra modules and making the use of extra modules configurable) and focusing on improvements on critical sections (loop, parsing etc...).

if I see Marlin source code I see a more complex code but that is still better organized. It's also in C++, so adding overhead for performance and size, but probably it is the better solution to manage the sw.

kisslorand:
I removed only the functions that did nothing more than return a variable's value or set one's value. No harm done to the readability of the code, no confusion created. So I guess we're on the same page just seeing things differently. :)
It hurts my eyes to see a function that retrieves a variable's value called in the same module where the variable is defined.

pfn (changes requested):
How about meeting in the middle and marking all the one-liner functions `inline`?

kisslorand:
It's not valid across modules.

PR #2230, opened 2021-10-30
Load/Unload menu - direct nozzle temperature select, Numpad float input overhaul

kisslorand:
### Description

   As I use very often the Load/Unload menu, I found myself many times hitting the temp display zone to set up the nozzle temp but obviously nothing happened... This PR lets you directly select the nozzle desired temperature via numpad right from the menu itself.  Just hit the temperature display zone as you would do in the "Heat" menu.
   Another thing done in this PR is the overhaul of the float number input via numpad. Nothing fancy done, just made some changes in the logic behind the input and avoided empty number display field. Maybe easier for users and lighter on MCU and memory.

### DIFF
--- a/TFT/src/User/API/UI/Numpad.c
+++ b/TFT/src/User/API/UI/Numpad.c
@@ -212,7 +212,7 @@ double numPadFloat(uint8_t * title, double old_val, double reset_val, bool negat
   NUM_KEY_VALUES key_num = NUM_KEY_IDLE;
   touchSound = false;
 
-  uint8_t nowIndex = 0;
+  uint8_t nowIndex;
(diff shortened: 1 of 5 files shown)


GregSKR:
It's useful, thanks.

A suggestion for people using the 'Heat' button, forcing the HOTEND index before showing the menu could be a good idea, so it will not go to the BED temp menu if it was selected before:

**LoadUnload.c** 

```
case KEY_ICON_5:  // heat menu
	heatSetCurrentIndex(currentTool);     <-----
	OPEN_MENU(menuHeat);
```

kisslorand:
Good idea! Thank you!

Issue #2237, opened 2021-11-10
TFT70 icons


mikerocklewitz:
### Description
TFT70 latest firmware from github fails stating out of date icons.  They install fine.

### Steps to reproduce

1. Downloaded the repo from 11/9
2. Flashed tft70 with included themes (I tried unified material and hybrid red - both had same issue)

**Expected behavior**
Icon/theme to work

**Actual behavior**
Display says icons are out of date and need to be updated....even when updated

### Hardware Variant
TFT70

radek8:
Do you have icons on the SD card in the TFT70 folder, BMP subfolder?

mikerocklewitz:
> Do you have icons on the SD card in the TFT70 folder, BMP subfolder?

Yes of course.  I reverted to my download from 3 days ago and it works fine.

Simply reporting an issue with newest repo change.

oldman4U:
@digant73 

Can you see a possible reason why your PR causes the reported problem?

cooldude204:
the tft35 v3.0 also has this issue tried multiple icon packs and seems its trying to find 0.bmp and 2.bmp but can't find them.

digant73:
I also have a TFT35 v3 and no issue at all

cooldude204:
yes I'm using the precompiled version

digant73:
if you are familiar to compile the code, try to compile it. It's not currently possible for me to compile it for you

cooldude204:
I'm familiar with compiling marlin so I should be able to figure it out, I will compile my own .bin and let you know

digant73:
> I'm familiar with compiling marlin so I should be able to figure it out, I will compile my own .bin and let you know

You only need to set "default_envs = BIGTREE_TFT70_V3_0" in platformio.ini and compile

cooldude204:
Compiled the firmware uploaded it with icons on the SD card and its still getting the error icons: update required

mikerocklewitz:
I can try building it myself tomorrow but it looks like someone else found the issue as well.

digant73:
@cooldude204 

try if this works

[BIGTREE_TFT35_V3.0.27.x.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/7516232/BIGTREE_TFT35_V3.0.27.x.zip)

cooldude204:
@digant73 unfortunately it didn't work. still says Icons: update required

[...]

Issue #2231, opened 2021-10-31
Printing freezes randomly when using the TFT50 SD card


magnificu:
I have the same issue with TFT50_v3.0, SKR2 and TMC5160_v1.2.
When I start the print form the TFT SD Card, randomly the print freeze (sometimes after one or two hours of printing), but the hot end and the bed keeps the temperature. The only way to recover is to press the restart button.
I use the latest TFT firmware version BIGTREETECH-TouchScreenFirmware-master from 30.10.2021.

In the BIGTREETECH-TouchScreenFirmware official release-xx.27 this issue is not present.

_Originally posted by @magnificu in https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/1650#issuecomment-955647289_

oldman4U:
Okay. And you can tell us how this is related to the TFT firmware!?

magnificu:
With the old version of the firmware I don't have this issue or when I print from the SKR2 SD card.
This happens only with the latest firmware version and when I print from the TFT SD card.
Printing from octoprint is fine as well.
I think it is something on how the serial communication is handle. I tried 250000 and 115200 bound rates, but I had no luck.

oldman4U:
So please check the version which broke it by installing every version between the one which worked and the current one and let us know.

magnificu:
The one that is working is the release-xx.27 released on 10.04.2021. Since then I think there are hundreds of commits and I don't know how to test them. Is there any way to simulate the prints and to save all the serial communication on a log file?

oldman4U:
You download the version and test it. Maybe every tenth at the beginning until it breaks. I do not believe that there is a shortcut.

Keep in mind that you are currently the only one with this problem so besides the firmware version there is likely something else which causes the issue.

magnificu:
Is there any recommended setting for #define SDCARD_CONNECTION in Marlin 2.0.9.2 ?

Configuration_adv.h

  /**
   * Set this option to one of the following (or the board's defaults apply):
   *
   *           LCD - Use the SD drive in the external LCD controller.
   *       ONBOARD - Use the SD drive on the control board.
   *  CUSTOM_CABLE - Use a custom cable to access the SD (as defined in a pins file).
   *
   * :[ 'LCD', 'ONBOARD', 'CUSTOM_CABLE' ]
   */
  #define SDCARD_CONNECTION LCD

radek8:
[...]

PR #2233, opened 2021-11-06
Some improvements + bug fixes

digant73:
**IMPROVEMENTS:**
*Implemented #2232
*Added "Live Text Background Color Rendering Technique" compile option also in Status Screen menu: New compile option "LIVE_TEXT_BG_COLOR_STATUS" was added in Configuration.h
 allowing to set (according to the used icons theme) the rendering of the live text and eliminating/reducing the flickering on alternating icons
*Reviewed "Live Text Background Color Rendering Technique" compile option also in Printing menu: Previous "UNIFORN_LIVE_TEXT_BG_COLOR" compile option was also renamed to "LIVE_TEXT_BG_COLOR_PRINTING" in Configuration.h
*Updated clock for MKS TFT: MKS TFT (based on STM32F10X_CL) supports 72MHz (as all the others TFT based on STM32F10X). Previous clock for MKS TFT was 48MHz. All credits for finding and fixing the bug go to @Kisslorand
*Improved performance on GUI_DispOne function on GUI.c: With a cost of a small increase of Flash size (due to duplication of code), function GUI_DispOne now avoids to make some checks in the two main loops
 (for width and height) for each char to be displayed

**SOME SUGGESTIONS:**
*For all icon themes with the exception of "The Round Miracle" (all TFT sizes) and "Unified" (only TFT70 size) suggested values are:
LIVE_TEXT_BG_COLOR_PRINTING 2
LIVE_TEXT_BG_COLOR_STATUS 6
*For theme "The Round Miracle":
LIVE_TEXT_BG_COLOR_PRINTING 2
LIVE_TEXT_BG_COLOR_STATUS 2
*For theme "Unified" on TFT70 (providing a color gradient on Status Screen menu):
LIVE_TEXT_BG_COLOR_PRINTING 2
LIVE_TEXT_BG_COLOR_STATUS 4

**BUG FIXES:**
*Fixed bugs on navigation menu in Print menu: ListMode was affected by a twice redrawn scrolling up/down due to wrong keys matching (KEY_DECREASE/KEY_INCREASE). Proper keys are KEY_PAGEUP/KEY_PAGEDOWN.
 In addition, title bar was also wrongly updated causing an avoidable flickering and lost of Temperature info (if enabled) on both icon/list mode. Now title bar is properly updated with no flickering
 and no loss of Temperature info (if enabled)
*Fixed type for param "pl_recovery_z_raise" in config.ini. It was wrongly defined as integer type while it was then used as float type

resolves #2232

[...]

PR #2241, opened 2021-11-11
Octoprint support + configurable gcode file comment parsing + bugfixes + cleanup

### LINKED ISSUE #438: [FR] OctoPrint and TFT comunication.
I use my TFT along with OctoPrint, but the TFT and OctoPrint don't communicate which makes things a bit fragmented.

I connect my RaspberryPi to my TFT E3 via the Rx & Tx pins so could there be a new Gcode command that the RaspberryPi and the TFT use to tell each other what they are doing? For example, if I started a print on the TFT Octoprint would be able to show a print is running and how much time was left, or vice versa. I am guessing a plugin would be needed for Octoprint for this to work.

### LINKED ISSUE #2180: [FR] Way for the TFT to detect prints started by octoprint
I suppose my problem would be that If I start a print job from OctoPrint then the BTT TFT doesn't get the "hint" and switch to the printing screen.

I poked around the source came up with a super complicated solution that's way over my head...I put in the alternatives I considered question below. My actual solution was to just add some gcode to octoprint's "gcode scripts" settings and a small patch to the tft's source code.

P2 is just the serial port that the tft is attached to. `P2` isn't required and I could let it go out all serial ports but I didn't want to risk octoprint getting confused.

Octoprint start `M118 A1 P2 action:printstart`  
Octoprint complete `M118 A1 P2 action:printcomplete`  
Octoprint canel `M118 A1 P2 action:cancel`  
Octoprint pause `M118 A1 P2 action:paused`  
Octoprint pause `M118 A1 P2 action:resumed`  

Then in the BTT TFT source code I added this small patch:

```diff
diff --git a/TFT/src/User/API/parseACK.c b/TFT/src/User/API/parseACK.c
index 850dadfd..dfd639ad 100644
--- a/TFT/src/User/API/parseACK.c
+++ b/TFT/src/User/API/parseACK.c
@@ -304,6 +304,14 @@ void hostActionCommands(void)
       }
     }
   }
+  else if (ack_seen(":printstart"))
+  {
+    setPrintHost(true);//Think this is the right function based on code comments. Triggers the printing screen like I want.
+  }
+  else if (ack_seen(":printcomplete"))
+  {
+    printComplete();
+  }
   else if (ack_seen(":paused") || ack_seen(":pause"))
   {
     if (infoMachineSettings.firmwareType == FW_MARLIN)
```

Writing a LCD driver in marlin to handle serial comms between marlin and btt tft. That would require a significant rewrite of the current serial code in this repo. The plus side would be "Marlin Mode" would, in theory, work over serial and support touch (faked encoder messages). Left and top of screen would be counter clockwise, right and bottom clockwise, and center is a click. Complicated right? That's where my head went first. I even have a couple pages for the basics of the serial protocol. 

**Additional context**
So far this seems to be working with minimal issues. The one issue being that after a print completes the top left of the main screen displays `:?` where normally is says `Ready`. The print summary popup has the same `:?` in it for the "title"  
I'm sure it is something I'm not handling correctly. Input on that would be great.

If desired I could eventually turn this into a PR once all the bugs are worked out.

### LINKED ISSUE #2216: [FR] RRF M150 support
Seems like the firmware as is, can't handle RBG LED settings on RRF.

Please modify the current code to be compatible with RRF.

**Additional context**
None of the button's works for now (r, g, b, w, on, off or the colour mixer)
https://marlinfw.org/docs/gcode/M150.html
https://duet3d.dozuki.com/Wiki/M150
If you have a solution, I have configured NeoPixels to test with on RRF 3.3

### LINKED ISSUE #2246: I gave distance values ​​for each corner in L corner manual leveling, but it is not normal
The product is using the latest firmware on TFT35.

Currently using BL Touch.
With L corner, it moves relative to the BL touch probe. I wonder if this is correct. I want to manually set the L corner with the nozzle and A4 paper, but it doesn't seem to work.

And if the edge spacing is given by distance, the distance is correct for the edge close to the x/y limit s/w, but the distance does not match for the edge close to the x/y max value. I wonder if there is a way to set all of these to the same distance.

The printer is an anet A8 modified product and the board uses skr 1.4.

It was translated by Google, so I hope you can understand even if the sentence is strange.

### LINKED ISSUE #2255: [BUG] TFT no more able to connect to Marlin
Just to avoid panic, last merged PR #2250 made the TFT not able to connect to mainboard at startup for any fw different than RepRap.
The issue is already fixed by PR #2241

digant73:
**IMPROVEMENTS:**
*Configurable Gcode file comment parsing feature: This previously hidden feature can now be enabled/disabled in config.ini (with param name "file_comment_parsing") and also in Feature menu. This gives visibility/control of the feature so it can be used according to the used slicer (e.g. Cura) and/or user preference.
If enabled, the current implementation parses and processes print time and print layer information from the G-code file (nothing else).
If disabled, the "layer_disp_type" setting provided in "UI Settings" section in config.ini becomes redundant.
*Optimized loopPrintFromTFT function in Printing.c API: The parsing of each line in the gcode file is now split in order to speed up both the gcode parsing and the comment (if any) parsing
*Added error handling and error notification in loopPrintFromTFT function in Printing.c API: The error handling function handleDiskError currently implemented forces a device re-initialization for a maximum number of retry attempts. If maximum number of retry is reached a print abort is forced. Previously, in case of error the file position (infoPrinting.cur) was never updated causing an infinite print loop (the print could never reach the end due to a permanent error reading from file)
*Unified name for Controller Fan Active: Id for Active PWD in config.ini (CtL) was not matching the Id displayed in the menus (CtS) and its meaning (Active). The name was unified to CtA on both config.ini and menus
*Implemented Octoprint support: It resolves #2180 and #438
*Some cleanup

**Octoprint requested commands/protocol:**
The following gcodes must be sent by Octoprint for triggering actions to the TFT:
**start:**
  M118 A1 action:print_start
**end:**
  M118 A1 action:print_end
**cancel:**
  M118 A1 action:cancel
**pause:**
  M118 A1 action:pause
**resume:**
  M118 A1 action:resume
**time remaining progress:**
  M118 A1 action:notification Time Left \<XX\>h\<YY\>m\<ZZ\>s (e.g. 02h04m06s)
  M117 Time Left \<XX\>h\<YY\>m\<ZZ\>s
**file data progress:**
  M118 A1 action:notification Data Left \<XXXX\>/\<YYYY\> (e.g. 123/12345)
  M117 Data Left \<XXXX\>/\<YYYY\>

When the trigger "print_start" is received, the TFT switches to Print menu.
Once on Print menu, the "pause", "resume" and "stop" buttons will be disabled.
That means only Octoprint will control the print.
Only on print end or cancel the TFT Print menu is finalized (statistics available etc...) and unlocked (it can be closed).

**NOTE:** A new plugin on Octoprint implementing the above protocol should be the preferable way (available to everyone)

**BUG FIXES:**
*Fixed wrong stringify expansion for icon names: In boot.c a double macro expansion was used causing a macro value expansion (in case of an existing macro definition) instead of macro name. The fix will avoid further issues such as happened with #2237. Previous definitions (LANGUAGE and ONBOARD_SD) have been restored
*Fixed TFT freeze due to permanent read/write error printing from SD slot: functions disk_read / disk_write in diskio.c performed a loop invoking SD_Init function. The loop was never terminated in case of a persistent error. That bug could be the main reason for the freeze reported in #2231
*Fixed MinX/MinY Leveling Distance: The origin of the bed is never less than 0. A negative minX/minY indicates the distance of the bed from endstops and must not be considered to determine the four leveling edges in Manual Leveling menu and Leveling Corner menu. It fixes #2246

resolves #2180
resolves #438
fixes #2246
could fix (#2231) (to be confirmed)
fixes #2255
fixes #2216

**PR STATE:** Ready to merge. It adds 560 bytes to program size


TFT/src/User/Configuration.h
```diff
@@ -484,25 +486,74 @@
...
+ *   Value range: [min: 10, max: 12000]
  */
-#define NOZZLE_PAUSE_RETRACT_LENGTH               15.0f  // (mm) (Default: 15.0f)
-#define NOZZLE_RESUME_PURGE_LENGTH                16.0f  // (mm) (Default: 16.0f)
-#define NOZZLE_PAUSE_X_POSITION     (X_MIN_POS + 10.0f)  // (mm) Must be an integer (Default: 10.0f)
-#define NOZZLE_PAUSE_Y_POSITION     (Y_MIN_POS + 10.0f)  // (mm) Must be an integer (Default: 10.0f)
-#define NOZZLE_PAUSE_Z_RAISE                      10.0f  // (mm) (Default: 10.0f)
-#define NOZZLE_PAUSE_XY_FEEDRATE                   6000  // (mm/min) X and Y axes feedrate (Default: 6000)
-#define NOZZLE_PAUSE_Z_FEEDRATE                    6000  // (mm/min) Z axis feedrate (Default: 6000)
-#define NOZZLE_PAUSE_E_FEEDRATE                     600  // (mm/min) retract & purge feedrate (Default: 600)
```
guruathwal, line 495: you should not remove the unit names.

digant73: they are reported in the comment above (as in config.ini)

guruathwal: The description above is wrong. 
you are explaining how to set the values in the **configuration.h** file, but the details are tailored for values **config.ini** file.
Like: `Format: [pause_feedrate: XY<feedrate> Z<feedrate> E<feedrate>]` does not apply to **configuration.h**

digant73: yes I know that. There are other params like that but it seems quite clear how they are related each other and with config.ini. Just to not waste too much time on that

guruathwal: The existing descriptions were similar to that in marlin firmware.
image

digant73: ah ok. it is not a problem to restore the previous description for that block of param (pause and level)

digant73: restored the old inline description

kisslorand:
While you're at it please set`setM73_presence` back to `setM73R_presence`. I don't know who changed it and why.
Thx.

digant73:
@kisslorand done

Copy to SD Card root directory to update/config.ini
```diff
@@ -113,6 +113,14 @@ emulated_m600:1
...
 emulated_m109_m190:1
 
+#### G-code File Comment Parsing
+# The TFT parses and processes extra information provided by the slicer as comments in the G-code file.
+#
+# NOTE: Enable it in case the slicer (e.g. Cura) supports extra information.
+#
+#   Options: [disable: 0, enable: 1]
+file_comment_parsing:1
+
```
kisslorand, line 123: You could mention that for the moment print time and print layer information are parsed and processed from the G-code file (and nothing else). 
Also it is not obvious for a regular user that if file comment parsing is not enabled than "layer display type" option becomes redundant.

TFT/src/User/API/icon_list.inc
```diff
@@ -149,4 +149,3 @@ X_ICON (MESH_EDITOR)
 X_ICON (BRIGHTNESS)
 X_ICON (MESH_VALID)
 X_ICON (CALIBRATION)
-//after changing/adding/removing an icon, change the ICON_FLASH_SIGN in Settings.h
```
guruathwal, line 152: Pleas restore this line.

TFT/src/User/API/boot.h
```diff
@@ -58,9 +58,9 @@ extern "C" {
 // This List is Auto-Generated. Please add new icons in icon_list.inc only
 enum
 {
-  #define X_ICON(NAME) ICON_##NAME ,
+  #define  X_ICON(NAME) ICON_##NAME ,
   #include "icon_list.inc"
-  #undef  X_ICON
```
guruathwal, line 63: Please remove the double spaces after #define

TFT/src/User/API/icon_list.inc
digant73: it is on top of the file as in all the other .inc files

TFT/src/User/API/boot.h
digant73: I aligned both #define, #include and #undef

Copy to SD Card root directory to update/config.ini
digant73: Just to avoid to provide wrong information/explanation, could you please provide those info or add them in a further PR?

TFT/src/User/API/boot.h
guruathwal: it does not make any sense to align them. #include is here is a sub-statement here.

digant73: ok. I will revert them

TFT/src/User/API/icon_list.inc
guruathwal: I deliberately added the line at the end to let anyone editing the files know it. Most of the time the description at the top is not read by many people and this cause problem for others. These types of comments should be left at the places they are relevant to.

TFT/src/User/API/boot.h
digant73: original #undef  X_ICON has double spaces. Shall I provide 1 spaces or what? It seems to me more coherent to align all or nothing

guruathwal: There should be 1 space. It can be like this:
```C
#define X_ICON(NAME) ICON_##NAME ,
#include "icon_list.inc"
#undef X_ICON
```
or like this to show that icon_list.inc only relevant for this portion only
```C
#define X_ICON(NAME) ICON_##NAME ,
  #include "icon_list.inc"
#undef X_ICON
```

digant73: ok for the second option (same convention used in all files)

TFT/src/User/API/Printing.c
```diff
@@ -375,67 +418,58 @@ void printStart(FIL * file, uint32_t size)
...
   {
-    case BOARD_SD_REMOTE:  // present just to make the code robust. It should never been executed
+    case REMOTE_HOST:      // present just to make the code robust. It should never be executed
+    case BOARD_SD_REMOTE:
+      return;
+
     case BOARD_SD:
-      //infoHost.printing = true;  // Not so fast! Let Marlin tell that he's done!
-      request_M27(infoSettings.m27_refresh_time);  // check if there is a print running from onboard SD or remote host (USB)
+      infoHost.printing = true;
```
kisslorand, line 435: I wouldn't re-enable this. I left it there commented with a purpose. Maybe I commented it wrong, it should have been "Not so fast! Let Marlin tell that he started printing!"

TFT/src/User/API/Printing.c
```diff
@@ -452,11 +486,15 @@ void printAbort(void)
...
   switch (infoFile.source)
   {
+    case REMOTE_HOST:  // nothing to do
+      loopDetected = false;
+      return;
+
     case BOARD_SD_REMOTE:
     case BOARD_SD:
-      //infoHost.printing = false;  // Not so fast! Let Marlin tell that he's done!
+      //infoHost.printing = false;  // not so fast! Let Marlin tell that he's done!
```
kisslorand, line 495: Sentences in English always start with capital letter.

TFT/src/User/API/Printing.c
digant73: In case of "remote onboard SD" or "remote host" (new octoprint support) printing it is now executed (on parseAck) the function printRemoteStart (which also set infoHost.printing).
Function printStart (the one you are referring to) is instead executed printing from TFT or onboard SD (so TFT SD, TFT USB and onboard SD requested by TFT). Previously there was also a bug with "onboard SD" printing. It was (correctly) executed printStart and then printRemoteStart (wrongly) (that also changed infoFile.source to ONBOARD_SD_REMOTE)

TFT/src/User/API/Printing.c
digant73: I simply use all the same comment style (lower case) in the files

TFT/src/User/API/Printing.c
kisslorand: As far as I know BOARD_SD is when Marlin prints from the SD card of the motherboard. Let Marlin tell to the TFT that he started the print. In case Marlin dosn't start the print the TFT will wrongly assume that the print has been started.

digant73: but the Print menu was already displayed by the caller of printStart and then again by printRemoteStart that also changed the source to ONBOARD_SD_REMOTE

TFT/src/User/API/Printing.c
kisslorand: It's basic grammar, sentences always have to start with capital letter. It's really not debatable.

TFT/src/User/API/Printing.c
kisslorand: So? That doesn't mean that it is correct only to ASSUME that Marlin started the print. Marlin has started the print only when it says so.
Let's look at it the other way, what happens if you let Marlin to tell the TFT that is started the print and don't just simply assume it did?  Will it do more harm than just blindly assuming Marlin has started the print?

digant73: if for any reason  (e.g. SD detached from onboard slot) the print cannot be started after M24 was sent, Marlin will provide an error message and, in both scenarios (old and new), you need to abort the print to close the Printing menu.

kisslorand: My question remains. What is wrong leaving Marlin to tell the TFT that the print has been started?
If `infoHost.printing` is false that means that Marlin hasn't start the print . Telling the TFT that Marlin has started the print, even if you know that it hasn't, it's just wrong. You can explain whatever you want, it is wrong.

kisslorand:
@digant73 
You might want to take a look in "print.c -> menuPrintFromSource()" where `bool` type `update` variable is used. 
In menuPrintFromSource() the update gets other values than `true` or `false` (0 and 2). This was introduced in your PR #2233.

digant73:
@kisslorand added the comment on variable definition. I don't see any reference to update = true/false (it is already 0/1).
I replaced "if (update)" with "if (update != 0)"

kisslorand:
Sorry, I was checking the current master. Alles ist in Ordnung!

sarvenn:
Hello @digant73 ,
Could you please explain this feature for a basic user? I could not get what is it for and how should I used or adjust it.
For example I'm using PrusaSlicer currently. What should be done in my slicer SW and config.ini files?

*Configurable Gcode file comment parsing feature: This previously hidden feature can now be enabled/disabled in config.ini (with param name "file_comment_parsing") and also in Feature menu. This gives visibility/control of the feature so it can be used according to the used slicer (e.g. Cura) and/or user preference.

digant73:
@savenn see the reviewed explanation

Hitamku:
@digant73 
This popped up after I updated the latest file. Line Number is not Last Line Number+1, Last Line: 0
What error is that?

digant73:
that was a bug caused by #2250 and fixed by this PR (as reported in #2255)

CRCinAU:
> that was a bug caused by #2250 and fixed by this PR (as reported in #2255)

I came hunting for this after wondering WTF had happened to my TFT35 ;)

For the moment, I just rolled back a build and its all good for now.... Thanks for fixing these...

ETE-Design:
@digant73   If the OctoPrint part a default setting in OctoPrint, or do you have to add the functions somehow?

CRCinAU:
You can likely add them in Octoprint Settings -> GCode Scripts...

mhegering:
Hi,
i am a little confused how to get this whole thing to work.
What connection and settings do i have to do when i want the printing screen in TFT mode when printing with Octoprint?
I tried different things but nothing seems to work as it should.
It would be very nice if someone could tell me what exactly is to do to get the connection and everything setup.

digant73:
@mhegering you simply need to provide the gcode commands provided at the beginning of the thread **Octoprint requested commands/protocol:**.
So for example when you start a print, send the gcode "M118 A1 action:print_start" and you will see the Printing menu will be displayed. When the print is finished, send the gcode "M118 A1 action:print_stop". Etc...

mhegering:
@digant73 Thanks for your answer.
I tried it already to put the Gcodes into Octoprints GCode Scripts and in the start and end GCode of my slicer. Nothing happens..
Anything else i have to do?

Guilouz:
@digant73 Where need to put these in Octoprint ? When print started ? 

time remaining progress:
M118 A1 action:notification Time Left hms (e.g. 02h04m06s)
M117 Time Left hms
file data progress:
M118 A1 action:notification Data Left / (e.g. 123/12345)
M117 Data Left /

Capture d’écran 2021-11-26 à 15 22 48

mhegering:
> @digant73 Where need to put these in Octoprint ? When print started ?
> 
> time remaining progress: M118 A1 action:notification Time Left hms (e.g. 02h04m06s) M117 Time Left hms file data progress: M118 A1 action:notification Data Left / (e.g. 123/12345) M117 Data Left /
> 
> Capture d’écran 2021-11-26 à 15 22 48

@Guilouz If I understood right, you just have to put it there how you did already. Doesn't work for me?
Does it work with your screen?

Guilouz:
@mhegering Not tried yet, I'm trying to know where to put time remaining progress and file data progress commands

digant73:
For time remaining progress you can ask your slicer (e.g. Cura) to provide that. "M117 Time Left hms" is also the gcode used by Cura, so you simply need to enable time remaining feature in Cura.

Guilouz:
@digant73 Just need to put M117 ? It's not clear, what about these commands ? 

M118 A1 action:notification Time Left hms
M118 A1 action:notification Data Left

mhegering:
> @mhegering Not tried yet, I'm trying to know where to put time remaining progress and file data progress commands

You also could use the Octoprint Plugin "Detailed Progress".

Can you please tell me when you get this working @Guilouz ? I'm still struggling.

digant73:
If you use Cura, Cura will provide periodically provide the gcode "M117 Time Left hms" in the file, so you don't need to do anything manually.
"M117 Data Left ..." was added by me (similar to M117 Time Left supported by Cura). I will say to skip it if Octoprint doesn't provide any trigger for the current file progress

Guilouz:
@mhegering Yes I know I already use DisplayLayerProgress plugin to show layers and time left in info case.

But I still did not understand the usefulness of the command M118 A1 action:notification

I thought that this allowed to display the layers and the time remaining in their appropriate boxes like when printing from the SD or the USB but not in fact they are just notifications

digant73:
M118 is an alternative to M117. So, if M117 is supported (e.g. by Cura or some Octoprint plugins) you don't need M118 at all

Guilouz:
@digant73 Just try, status change with Octoprint doesn't work, I have added M118 A1 action:print_start in startup of Octoprint and screen doesn't switch

digant73:
@Guilouz Strange you needed to provide P0. P0 should be the default value in case Pn is omitted.
Hmm no, I think it isn't worth it to provide a translation for "Remote printing...". There are also many other hard coded strings not translated.

inigoml:
> M118 is an alternative to M117. So, if M117 is supported (e.g. by Cura or some Octoprint plugins) you don't need M118 at all

Hi guys. Fantastic upgrade. I've also gotten almost the same functionality from Octoprint that locally from SD.
Regarding this issue, I was using my customized Cura post processing script to insert an M118 A1 P2 in order to notifiy progress to my printer via "action:notification". However, now, since I send command action:print_start, there is no notification available. Messages arrives to TFT (it sounds) but there is no message screen.

digant73:
@inigoml you should see the Printing menu is open

inigoml:
Hi @digant73, what do you want to say with "printing menu is open"? 
By the way, I've modified Cura postscript in order to accommodate print progress % for printing menu. Now it`s updated properly with M118 command.
Only Time Left is not visible.

inigoml:
[DisplayProgressOnLCD.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/7627792/DisplayProgressOnLCD.zip)

inigoml:
PXL_20211130_181225850

inigoml:
PR for cura plugin already created. Will be in next version or not! :)

digant73:
@inigoml I meant that sending "M118 A1 P0 action:print_start" should trigger the TFT to open the Printing menu. But as I see in your picture, the Printing menu is now properly open.
The Time Left should be already supported in Cura (both M118 and M117).

inigoml:
Hi @digant73. The problem is that after print has started, no ETA is displayed on screen because message box for notifications is not available.
That is, ETA messages are received (Time Left XXhXXmXXs) but no message box is available in order to be displayed. 

M118 A1 P0 action:notification Time Left 1h2m3s has no way to be displayed. ;)

Guilouz:
Just try, I have elapsed time, time remaining, percentage, layer height in their respective case but not number of layers (just show ---)

digant73:
@inigoml the Time Left notification is displayed directly in the Printing menu (you will see the time elapsed and below the time remaining). The notification is not saved in the Notification box

inigoml:
> @inigoml You use 'M118 A1 P0 action:notification Data Left' in your script as percentage right ? It's not supposed to be used for number of layers ?

It depends. You can modify behavior in config.ini. 
#   Options: [Percentage & Elapsed time: 0, Percentage & Remaining time: 1, Elapsed time & Remaining time: 2]

Basically, I'm emulating M73 command via M118.

> Another little issue, percentage are not the same on octoprint and on the screen (example : 35% on screen, 37.6% on octoprint)

Octoprint has it's own calculation algorithm while Cura has another. In this case, since the script is for Cura, it's normal to have minor discrepancies. If an Octoprint plugin manages this instead of Cura, both could be equal. But no plugin is yet available for Octoprint (if I find myself strong, I could take a look about how to create an Octoprint plugin).

inigoml:
> @inigoml the Time Left notification is displayed directly in the Printing menu (you will see the time elapsed and below the 
 
Thank you @digant73. I will check config again. I'm almost sure I have prog_disp_type:2 set, but behavior is like if I had prog_disp_type:0. Perhaps it's not being properly parsed due to some error at script level.

inigoml:
At Gcode level I think it's everthing OK...

```
M118 A1 P0 action:notification Data Left 0/100
M118 A1 P0 action:notification Time Left 2h29m56s
```

Guilouz:
@inigoml yes you can switch time info (percentage, time remaining, elapsed time) by clicking on the time box, no problem for that I have all info.

What I mean is that normally M118 A1 P0 action:notification Data Left must be used to display number of layers (example : layer 127/528) and not and not a percentage out of 100, no ?

inigoml:
OMG.... You are right! :-) Everything working properly. 

Since I've reused same calculation at plugin level that was used for M73 command, I applied the same percentage. Just that.
If it was not accurate for M73, it's not either for M118.

Guilouz:
To summarize, we have all the info except the number of layers (displayed as '---', and all without notifications, all info are in boxes.

inigoml:
For layer, we need a parser at TFT level in order to process specific message via M118. @digant73 could add it and we can include it at Cura script level or eventually at script level.
So, we cannot add layer number until digant prepares the parser for it. ;) Let fill a FR!

digant73:
the layer number is already supported by the TFT. You must enable "File comment parsing" in Feature menu

inigoml:
@digant73, would this file comment parsing work also from Octoprint?
Since in this case GCODE does not arrive at TFT, we need to notify TFT controller some way layer number, isn't it?
;LAYER comment at Gcode never arrives at TFT and so must be notified some way (M118 A1 P0 action:notification "message")
Would  `M118 A1 P0 action:notification ";LAYER:12" ` work?

Guilouz:
@digant73 yes working when printing from TFT ports but not in remote printing.

inigoml:
Understood.

digant73:
ok, so it is not supported also by Octoprint. Not a big problem for me.

inigoml:
Adapting octoprint_detailedprogress plugin for TFT BTREETECH should be easy.... Will take a look.

Eddiiie:
I added these lines to Octoprint GCODES and it works great.
I can't find where to put the time remaining and filedata GCODEs; I use Prusa Slicer... ?   @inigoml  can your Python script be run from slicer post processor?

inigoml:
Python script is only for Cura. I was testing with Prusa Slicer but
eventually you would need a new script. Although Slicer can execute python
code, I didn't find proper programing variable suitable in order to do the
same from Prusa Slicer.

Eddiiie:
Thank you.

PR #2249, opened 2021-11-15
Update and replace some icons & ASCII font for TFT24/28

guruathwal:
- Replace the unlicensed/copyrighted delta calibration icon.
- Rename **'calibration.bmp'** to **'delta_calibrate.bmp'**
- Update corner/manual levelling icons.
- Remove duplicate marlin mode icon and add touch mode icon.
- update ASCII font for TFT24/28 to be slightly taller and easy to read.


radek8:
You can look at the icons marlin.bmp a marlin_mode.bmp?
These two icons could be merged into one. It is an unnecessary waste of space on two icons with the same meaning.

guruathwal:
Yes that makes sense. I will change it.

radek8:
Round icons (touch_mode and marlin_mode) updated in PR # 2251

Was there a license problem on the Delta icon? Will it be necessary to fix the round delta icon?

guruathwal:
Yes, it should be fixed as the icon is not licenced to be used here.

digant73:
good you updated fonts for TFT24/28. Many people (one of that it's me) had problems reading in particular some letters/numbers (e.g. 0 very similar to 6 and 8)

kisslorand:
The update of the ASCII fonts for 2.4" and 2.8" TFT is highly welcomed. In some places the text position should be lowered a bit (1-2) pixels to be vertically centered (first place would be the status screen).

PR #2250, opened 2021-11-16
RRF: add support for sending gcode checksums

pfn:
### Requirements

No new requirements.

Testing help on Marlin and other firmware beside RRF would be appreciated.

### Description

When the firmware type is not yet detected, or RRF, send gcode commands with checksums.

Background:

RepRapFirmware 3.4beta4 changed the behavior of checksum handling and now works best when all host gcode is sent with checksums.

### Benefits

Allows using checksum-mode on RRF in 3.4beta4 and later.

### PR Status

Ready to merge pending code review and testing on other firmware.

### Testing

Verified to work on my MKS TFT28 with RRF 3.3 (I have not yet updated my printers to 3.4 betas)

### DIFF
--- a/TFT/src/User/API/RRFSendCmd.h
+++ b/TFT/src/User/API/RRFSendCmd.h
@@ -1,0 +1,5 @@
+#ifndef _RRF_SEND_CMD_H_
+#define _RRF_SEND_CMD_H_
+
+void rrfSendCmd(const char* cmd_ptr);
+#endif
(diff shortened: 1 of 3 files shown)


TFT/src/User/API/RRFSendCmd.h
```diff
@@ -0,0 +1,5 @@
+#ifndef _RRF_SEND_CMD_H_
+#define _RRF_SEND_CMD_H_
+
+void rrfSendCmd(const char* cmd_ptr);
+#endif
```
digant73, line 5: it seems you should provide a blank line after \#endif

PR #2251, opened 2021-11-16
update round icons for PR 2249

radek8:
Update round icons for PR #2249


digant73:
as far as I see, Gurmeet removed only the icon "marlin.bmp". You removed also "marlin_mode.bmp".
Gurmeet's PR is also renaming "calibration.bmp" to "delta_calibrate.bmp"

PR #2257, opened 2021-11-18
No purge on resume if manual extrusion was done after a pause during print

kisslorand:
### Requirements

BTT or MKS TFT

### Description

~- Text inside icons vertically centered in StatusScreen and Preheat menu.~
- No more purge on resume if manual extrusion or filament unload/load was done after a pause during print (after pause and filament change, the resume made an unnecessary purge leaving filament debris on the print) 

### Note

~Text vertical alignment was done only for TFTs with 320x240 and 480x320 resolution as I only have these for testing. For the remaining resolutions testing and suggestions from others are welcomed.~
The no-purge applies only if printing from TFT's SD or USB.
This PR adds 4 bytes to RAM and 88 bytes to program memory.


TFT/src/User/API/Printing.c
```diff
@@ -584,10 +592,14 @@ bool printPause(bool isPause, PAUSE_TYPE pauseType)
...
-          {
+          if (extrusionDuringPause == true)  // check if extrusion done during Print -> Pause
+          { // no purge
+            extrusionDuringPause = false;
+          }
+          else if (heatGetCurrentTemp(heatGetCurrentHotend()) > infoSettings.min_ext_temp)
+          { // purge
             mustStoreCmd("G1 E%.5f F%d\n", tmp.axis[E_AXIS] - infoSettings.pause_retract_len + infoSettings.resume_purge_len,
-                         infoSettings.pause_feedrate[FEEDRATE_E]);
+                        infoSettings.pause_feedrate[FEEDRATE_E]);
```
digant73, line 602: as far as I see in all the code, the code is aligned after "(", so the previous code was correct

kisslorand: I didn't touch it, must be my VSCode acting on its own... 
It's fixed now..

digant73:
It seems that (maybe with the last font pack update provided by Gurmeet) that the text is already centered for both 320 and 640 resolution.
Below some screenshots (from TFT28 and TFT35 V3):

IMG_20211121_115731

IMG_20211121_115749

Ready_3

Preheat_2

In 320 res, on both Status and Preheat menu the text has the same number of pixels above and below (white + blue)
In 640 res, on both Status and Preheat menu the text has 1-2 pixels above more than below

So, lowering the text a bit more will (in particular for 640 res) not improve alignment

kisslorand:
@digant73 
You did the same mistake that
 I did, didn't check with other themes. I wrongly assumed the icon makers did their homework and made the icon packs compatible among them.


PreHeat
StatusScreen

This is TFT28 (320x240) with Round Miracle icon pack.

On my TFT35 v3 the status screen text is vertically centered and in the PreHeat menu it's 1 pixel up.

So... we cannot save both the cabbage and the goat .

digant73:
I had the feeling you were using another icon pack than the standard one. But at the end it seems to me more a change that should be done on the icon packs (maybe also other icon packs are affected) otherwise the problem is moved to the standard icon pack (that at the end is the most maintained)

kisslorand:
OK. I will revert the changes and use it only on my personal builds.

radek8:
I fix round icons. Which displays will be involved?
Only 24 "(28") and 35 ", or more?

kisslorand:
I could only test resolutions 320x240 and 480x320. These resolutions are for 2.4", 2.8" and 3.5" TFTs.
4.3" and 5.0" TFTs use 480x272 and 7.0" TFTs use 800x480.

Issue #1151, opened 2020-10-03
[BUG] leveling not on right point


Tillema:
### Description

To start, imagine a printer with endstops that are not on coordinates (0, 0), but at e.g. (-5, -18) (in Marlin, this is configured with X_MIN_POS, Y_MIN_POS, X_MAX_POS, and Y_MAX_POS).
I've set the leveling to 20 mm from edge. After flashing, the leveling points are not on (20, 20), but on (15, 2).
I fixed it with setting the leveling positions in Leveling.c, but there are a lot of users who don't know this shit.
Could you please update the positions, that is moves to the spot you told to go to?
Currently, on the X-axis, the first leveling point is set as `infoSettings.machine_size_min[X_AXIS] + infoSettings.level_edge`. For the min point it is quite easy to fix: just use `infoSettings.level_edge`. Don't know how to do this on the max position, because that could be outside the bed as well.
### Hardware Variant
Used the latest firmware with the TFT35 E3 V3.0

If you need more information, or testing, I would like to help you.

clon1958:
Hello
I also have the same problem machine zero does not match table zero
Show me where to revise the code
Instead of this:
{infoSettings.machine_size_min [X_AXIS] + infoSettings.level_edge, infoSettings.machine_size_min [Y_AXIS] + infoSettings.level_edge}
Write like this:
{infoSettings.level_edge, infoSettings.level_edge}

zonalimitatore:
I've no understand.... Where? in marlin mode or touch mode? Ubl,bilinear or corner?

Tillema:
C'mon dudes. Just ping the search bar to find the phrase. It's in TFT/src/User/Menu/Leveling.c (for completeness: lines 6 - 11). Pretty clear it regards manual leveling points in Marlin mode.

zonalimitatore:
Ok,so it don't keep the marlin's offsets?

Tillema:
It's very clear it uses the absolute minimum, which could be negative as well. This is maybe because some printbeds has their origin in the center (mainly delta printers). Somewhere a function to add real bed size and home offsets (from origin) need to be specified in order to get the right spots on where to level.

clon1958:
And the solution to the problem will be

oldman4U:
Sounds to me like you would know the solution to the problem you reported also....

So what's next?

Tillema:
> Sounds to me like you would know the solution to the problem you reported also....

[...]

PR #2264, opened 2021-11-24

NOVAXIM:
### Description
Currently, during manual bed leveling the edge bed coordinates are counted from the minimum value (Home position). Unfortunately, when we have coordinates other than 0 at the Home position (e.g. size_min:X-10 Y-5), the edge positioning is faulty. With level edge distance value at 20, the lower left corner will be at  position X10 Y15, not X20 Y20.

### Benefits
This solution calculates the correct bed edge coordinates for manual leveling.
The `level_edge_distance` will be counted from true 0 position. For example, if you have `size_min: X-5 Y18` and `level_edge_distance: 30`, the lower left leveling point would be at X30 Y30. This applies to all corners and center point too (if you have the correct `size_max`).

#### Note:
When you have enabled `M115_GEOMETRY_REPORT` option in Marlin (`Configuration_adv.h`), then don't need to set `size_min` and `size_max` in file(s): `config.ini` and/or `Configuration.h`.

### Related Issues
#1151

### DIFF
--- a/TFT/src/User/API/LevelingControl.c
+++ b/TFT/src/User/API/LevelingControl.c
@@ -6,10 +6,10 @@ float probedZ = 0.0f;                         // last Z offset measured by probe
 
 void levelingGetPointCoords(LEVELING_POINT_COORDS coords)
 {
-  int16_t x_left = infoSettings.machine_size_min[X_AXIS] + infoSettings.level_edge;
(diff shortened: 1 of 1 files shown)


radek8:
PR #2241 solves this problem.

If you enable M115_GEOMETRY_REPORT and use the current Marlin, you will have no problem.
You won't even need a fix

PR #2266, opened 2021-11-25

NOVAXIM:
### Description
Currently, during manual bed leveling the edge bed coordinates are counted from the minimum value if >=0 (Home position). Unfortunately, when we have coordinates other than 0 at the Home position (e.g. size_min:X-10 Y-5), the physical edge positioning will is faulty. With level edge distance value at 20, the lower left corner will be at **real** position X10 Y15, not X20 Y20 (liked in Marlin). The last fix did not completely delete the problem.
**Positioning with manual leveling is an internal affair of TFT Firmware and should not be fixed otherwise as here.**

### Benefits
This solution calculates the correct bed edge coordinates for manual leveling **regardless of**:
- Marlin version,
- `M115_GEOMETRY_REPORT` Marlin option,
- end stoppers position.

The level_edge_distance will be counted from true 0 position (liked in Marlin). For example, if you have `size_min: X-5 Y-18` and `level_edge_distance: 30`, the lower left leveling point would be at **real** X30 Y30 (liked in Marlin). This applies to all corners and center point too (if you have the correct `size_max`).

**Note:** When you have enabled `M115_GEOMETRY_REPORT` option in Marlin, then don't need to set `size_min` and `size_max`. However, if you don't have `M115_GEOMETRY_REPORT` option in Marlin, then you then you have to enter values for `size_min` and `size_max` in file(s): `config.ini` and/or `Configuration.h`.

### Related Issues
#1151


digant73:
The current solution forces "at least" 0 for minX and minY (with or without M115_GEOMETRY_REPORT enabled) but allows minX, minY > 0 just to apply a sort of inset (reducing also maxX and maxY). E.g. real build volume is 300x300 but you prefer to force an inset of 10mm so you set mixX=10, minY=10, maxX=290, maxY=290 in config.ini. That scenario will be possible from config.ini in case M115_GEOMETRY_REPORT is disabled (no override from Marlin)

NOVAXIM:
Sorry but I couldn't correct the file, so I close this PR.

Issue #2128, opened 2021-08-17
How to make corner Bed leveling use the Correct Dimensions of my Printer?


XoBoR:
Greetings,

I'm trying to make corner bed leveling work, however it does not use the correct bed dimensions, instead it uses the default values of 235x235x250.

The documentation states: 

> The TFT will auto-detect the machine size (min and max) in Marlin firmware (requires enabling `M115_GEOMETRY_REPORT` in Configuration_adv.h in Marlin firmware)

I have enabled `M115_GEOMETRY_REPORT` in my Marlin firmware, and going through the Terminal of my TFT70 I can see that it is reporting the correct dimensions of my Printer.

I have tried to set the correct Dimensions manually in Configuration.h, but the values I've put in are being ignored.

I have tried comment out Dimensions in config.ini, no change.

Putting the correct values in config.ini does work, but still, why isn't the TFT using the dimensions reported by the board?

Is this a Bug or am I missing something?

radek8:
I can confirm the reported problem.
FW TFT will use the values size_min: and size_max:
in the config.ini file,
but does not use values obtained from the printer via the M115 command.

radek8:
The problem could be that Marlin reports parameters as lowercase xyz, but TFT expects uppercase XYZ.

![image](https://user-images.githubusercontent.com/46979052/129811722-e62f2116-ca0f-42d5-8cca-2aaa9c1b5c33.png)

radek8:
It won't be a problem to fix it.
However, another problem occurs if you have the origin of the x and y axes defined as a negative value. Just like I have on my printer.
In the config.ini file I had a defined desktop as min 0, max 230, and everything worked fine.
if my printer detects the starting point as a negative number, I have a problem with the corner alignment feature.

radek8:
Unfortunately, if I rewrite large XYZ tags to small xyz, the bed size will not be detected correctly.

      // parse and store build volume size
      else if (ack_seen("work:"))
      {
        if (ack_seen("min:"))
        {
          if (ack_seen("x:")) infoSettings.machine_size_min[X_AXIS] = ack_value();
          if (ack_seen("y:")) infoSettings.machine_size_min[Y_AXIS] = ack_value();
          if (ack_seen("z:")) infoSettings.machine_size_min[Z_AXIS] = ack_value();
        }
        if (ack_seen("max:"))
        {
[...]

PR #2267, opened 2021-11-25

NOVAXIM:
### Description
Currently, during manual bed leveling the edge bed coordinates are counted from the minimum value if >=0 (Home position). Unfortunately, when we have coordinates other than 0 at the Home position (e.g. size_min:X-10 Y-5), the physical edge positioning will is faulty. With level edge distance value at 20, the lower left corner will be at **real** position X10 Y15, not X20 Y20 (liked in Marlin). The last fix did not completely delete the problem.
**Positioning with manual leveling is an internal affair of TFT Firmware and should not be fixed otherwise as here.**

### Benefits
This solution calculates the correct bed edge coordinates for manual leveling **regardless of**:

- Marlin version,
- M115_GEOMETRY_REPORT Marlin option,
- end stoppers position.

The level_edge_distance will be counted from true 0 position (liked in Marlin). For example, if you have size_min: X-5 Y-18 and level_edge_distance: 30, the lower left leveling point would be at **real** X30 Y30 (liked in Marlin). This applies to all corners and center point too (if you have the correct size_max).

**Note:** When you have enabled `M115_GEOMETRY_REPORT` option in Marlin, then don't need to set `size_min` and `size_max`. However, if you don't have `M115_GEOMETRY_REPORT` option in Marlin, then you then you have to enter values for `size_min` and `size_max` in file(s): `config.ini` and/or `Configuration.h`.

### Related Issues
#1151

### DIFF
--- a/TFT/src/User/API/LevelingControl.c
+++ b/TFT/src/User/API/LevelingControl.c
@@ -6,9 +6,9 @@ float probedZ = 0.0f;                         // last Z offset measured by probe
 
 void levelingGetPointCoords(LEVELING_POINT_COORDS coords)
 {
-  int16_t x_left = ((infoSettings.machine_size_min[X_AXIS] < 0) ? 0 : infoSettings.machine_size_min[X_AXIS]) + infoSettings.level_edge;
+  int16_t x_left = infoSettings.machine_size_min[X_AXIS] - infoSettings.machine_size_min[X_AXIS] + infoSettings.level_edge;
   int16_t x_right = infoSettings.machine_size_max[X_AXIS] - infoSettings.level_edge;
-  int16_t y_bottom = ((infoSettings.machine_size_min[Y_AXIS] < 0) ? 0 : infoSettings.machine_size_min[Y_AXIS]) + infoSettings.level_edge;
+  int16_t y_bottom = infoSettings.machine_size_min[Y_AXIS] - infoSettings.machine_size_min[Y_AXIS] + infoSettings.level_edge;
   int16_t y_top = infoSettings.machine_size_max[Y_AXIS] - infoSettings.level_edge;
 
   if (GET_BIT(infoSettings.inverted_axis, X_AXIS))


NOVAXIM:
I just finished checking on my printer and this solution works like this:
<a href='https://postimages.org/' target='_blank'>Leveling-from-real-0</a>

digant73:
In case of max endstops, maxX and maxY cannot be satisfied (even by the current solution). Also infoSettings.machine_size_min[X_AXIS] - infoSettings.machine_size_min[X_AXIS] and infoSettings.machine_size_min[Y_AXIS] - infoSettings.machine_size_min[Y_AXIS] is always equivalent to 0

radek8:
Isn't it better to upgrade to Marlin which doesn't contain an error?

NOVAXIM:
> In case of max endstops, maxX and maxY cannot be satisfied (even by the current solution).

I don't understand well. Can you write what you mean?

In my situation I have:
`infoSettings.machine_size_min`[X or Y] liked end stop position (at me is X-10 and Y-5),
`infoSettings.machine_size_max`[X or Y] liked bed size.

My revision counts `level_edge_distance` from **real** zero position. What is wrong?
I have this sam place in Marlin mode for all corner leveling points and center leveling point.

NOVAXIM:
> Isn't it better to upgrade to Marlin which doesn't contain an error?

What's the bug in Marlin?!
Which version does not contain an error?
Why do you negate the solution that correctly counting point `level_edge_distance`?

Does my solution work badly with the latest version of Marllin?
Please report me here when it is.

**Note:**
My printer show X-10 Y-5 after start (flashing before homing) and all is right. My printer go to home before position at the leveling point. Everything works as it should. I am satisfied.

NOVAXIM:
Marlin has nothing to do with this, because the formula for the manual leveling position **is based on internal variables.** Is it so hard to understand?!

radek8:
> Marlin has nothing to do with this, because the formula for the manual leveling position **is based on internal variables.** Is it so hard to understand?!

Otevři terminál pošli na příkaz příkaz M115 a odpověď pošli.
Ukaž ti kde byl problém v Marlin.

digant73:
@NOVAXIM please see my answer on your previous PR. The current solution is already fixing the case were min must be "at least" 0. That approach should also be applied to max (in case there are max endstops) but that is impossible because we have no info about the real bed size. For that reason a gave the possibility to set min even greater than 0 (because you can also set max lower than the real bed size)

NOVAXIM:
I'm sorry you waited so long. This is the printer's response to the M115: 
area:{full:{min:{x:-10.0000,y:-5.0000,z:0.0000},max:{x:220.0000,y:220.0000,z:250.0000}},work:{min:{x:-10.0000,y:-5.0000,z:0.0000},max:{x:220.0000,y:220.0000,z:250.0000}}}

radek8:
Here is a bug in the old version of Marlin
The workspace starts at zero, not a negative value.
Update Marlin and it'll be fine

image

radek8:
see
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/2128
https://github.com/MarlinFirmware/Marlin/issues/22598

PR #2271, opened 2021-11-27
Controller fan

radek8:
If I enable the controller fan, the speed is displayed on the status screen along with the speed of the cooling fan.
This is fine, I can see the speed of all the fans and I can edit them.

But if they start printing, I'm only interested in the speed of the cooling fan and I don't need to see information about the speed of the controller fan on the print screen. This information is confusing.
This PR eliminates cycling between the speed of the chiller and the controller fan on the print screen.

### DIFF
--- a/TFT/src/User/Menu/PrintingMenu.c
+++ b/TFT/src/User/Menu/PrintingMenu.c
@@ -381,7 +381,7 @@ static inline void toggleInfo(void)
     {
       do
       {
-        currentFan = (currentFan + 1) % MAX_FAN_COUNT;
+        currentFan = (currentFan + 1) % MAX_COOLING_FAN_COUNT;
       } while (!fanIsValid(currentFan));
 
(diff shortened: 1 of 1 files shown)


sarvenn:
@radek8 hi,
controller fan is the board cooling fan, right?

radek8:
Yes

radek8:
M710

NOVAXIM:
Now, if you turn off the mainboard / steppers fan, then you won't notice any risk of burning the electronics. This is a stupid fix.

radek8:
> Now, if you turn off the mainboard / steppers fan, then you won't notice any risk of burning the electronics. This is a stupid fix.

Why would I turn off the fan when printing? Turns on and off automatically depending on whether steppers are active. I only set the speed in the menu.
If I set the speed to zero, it is the same error as guiding the nozzle to bed without the previous home position.
A user error is always possible, but cycling 3 values on the fan icon when printing will not help.
Usually the fan is also audible, so there is also audio feedback.
It suits me and I assumed that others would appreciate it as well.

Issue #1830, opened 2021-04-12
TO DO List


bigtreetech:
### Bugfix
 - [x] Klipper 12864 mode #290 
**progress:** The API Application of the upper layer is basically implemented in #705 . In my test, the problem occurs in the SPI slave part of the bottom layer. The SPI data received from klipper will shift by one bit. 
 - [x] RGB Printing heating color not change #1786
**progress:** In fact, it is not a really bug, It's a known limitation. Marlin will activate the LED change only after receiving the "M190/M109" gcode, but this gcode will cause the communication interruption between the TFT and Marlin (because Marlin will not respond to other gcode before reaching the target temperature), so TFT will convert this gcode to "M140/M104" to avoid communication interruption, but Marlin will not be able to activate the LED change normally.

### Improve
 - [x] Restructure `platformio.ini` File, using the relying on expansion to avoid redundancy
 - [ ] Restructure `icon/list`/etc... menu API (and related touch API) provides a more general API interface
 - [x] Neopixel header LED number #1764
 - [x] Machine RGB setting custom color value #1687

### Development direction
 - [x] Vertical Screen #524
 - [ ] CNC Menu #1750

oldman4U:
I would add "Developers wanted" to the headline.

digant73:
I can implement the machine RGB settings custom color.
My idea is to to provide in the current RGB menu:
1) a standard color select button. A list menu reporting standard colors will be used to select the color
2) a custom color select button. A new menu to select the RGB components will be available to set color. The menu will be usable as general RGB color pickup. A proper callback to handle the selected color will be registered. E.g. a callback to set the printer LED color etc...
3) a button to switch between standard/custom selected color
4) a ON/OFF button to switch on/of the printer LED

digant73:
could it OK as menu to select a custom LED RGB color?
I will add also the W, intensity and index for NEO Pixel

![IMG_20210415_231203](https://user-images.githubusercontent.com/64427768/115012745-7b992980-9eb0-11eb-80a1-614d9682d69e.jpg)

oldman4U:
Can you click on the value to get a keypad?
Can you see the resulting colour somewhere (approximately;-))?

digant73:
yes for both questions

sinole1987:
can you add mid print nozzle freezing to the to do list. Its a very big issue and never got fixed. last night I had another freez. It mostly happen on long prints.

[...]

Issue #524, opened 2020-04-03
Vertical Screen GUI


ChengQian1007:
It will be super, if a vertical GUI is supported.

drphil3d:
Please Please Please Please

bigtreetech:
How's this?
![image](https://user-images.githubusercontent.com/38851044/90588618-a3939880-e20e-11ea-8cb4-05b75876926d.png)

![image](https://user-images.githubusercontent.com/38851044/90588547-7515bd80-e20e-11ea-836d-453c515a2ee0.png)

OVDEN13:
how to get it?

ChengQian1007:
Cool! That is what I want!

ChengQian1007:
Thanks a lot!

hazeyjay77:
At last!!! I need i need i need please give me lol :)

9zdot:
Nice, that's exactly what I was looking for. 
When this feature will be available?
Thanks

9zdot:
Hey @bigtreetech do you have plans to release it? I'll be glad if you can give us more information what was changed so me and others can compile ourselves to use in our personal projects.

drphil3d:
@bigtreetech I love you man / woman

oldman4U:
Could you please be so kind and close this ticket. Your feature request is covered in the pinned Feature Request ticket #1177 

Thank you

SlackHacky:
Any update on release of portrait mode

oldman4U:
Hi. As far as I know there is currently no development on this. Are you a developer which could help to develop this?

radry:
Why is this issue closed? It hasn't been implemented yet.
@bigtreetech please release vertical GUI

oldman4U:
Hi radry.

Many abandoned tickets made this system unusable in the past. Therefore a bot has been installed which closes tickets which have not been used for more than 2 months. Unfortunately because of this, also important tickets like this one have been closed, so there were two new tickets introduced, where verified bugs and feature requests are collected. You can find the two tickets at the top of the front page of this repository.

Unfortunately I can not tell you when a vertical interface will be released, but probably the lack of open source developers who are willing to spend their spare time developing it will be a part of the reason. Are you a developer who is willing to help or is there anything else you believe you could help us with?

Hope this answers your question, please let me know

Thank you

radry:
@oldman4U 
[...]

PR #2273, opened 2021-11-28
Implement Vertical GUI (Portrait Mode)

radry:
### Requirements

**Currently vertical GUI is only implemented for TFT24.** 
Adding all other screens is easy since the base code is all the same but requires an actual screen to test it (which I don't have).
Only screen and controller specific files have to be changed/added.
How to do this is described below.

**Compile:**
To use TFT24 in portrait mode you need to compile for "BIGTREE_TFT24_V1.1_P" (Default enviroment is set to this in this commit).

**Config.ini:**
set "rotated_ui" to 2 or 3

**Theme/Icons:**
logo.bmp must have a portrait resolution (for TFT24 that's 240x320 instead of 320x240). Do not simply rotate it, the logo must still stay the same orientation.

### Description

This is a first implementation of a vertical GUI (Portrait Mode).

**How is it implemented:** 
My way of approaching this is as follows:

- Treat the vertical orientation as if it's its own screen. That means to use a screen in vertical orientation, it needs to be compiled for portrait mode. It can then only be used in portrait mode. To use landscape mode you just use the regular .bin and reset the settings.

- Most code that handles vertical GUI is filtered by `#ifdef PORTRAIT` and thus only compiled when target is a Portrait mode screen (PORTRAIT is defined in Vartiants/pins_TFTxxxx which has its own version for vertical orientation). 
This keeps the resulting .bin small. 

### How it looks
**UPDATE 06.12.2021**

**Status Screen**
Status_new
**Heat Menu**
heat_new
**Main Menu**
main_new
**Move**
move_new
**Level**
level_new
**Printing**
printing_new_1
**List Layout**
ListMenu

### How to add portrait mode to other screens

_Coming Soon_

### Benefits

Adds vertical GUI

### Related Issues

#524 
#1830 
#1170


digant73:
as a personal suggestion, I would bind KEY_INFOBOX area as you made in Status Screen menu (so centered at the bottom). That means the temp info in the Heat menu will be centered at the bottom of the screen.
In the Printing menu I would swap the "stop" button with "more" or "babystep". Also I would move the progress bar below the "speed" icon (so just above the "pause/resume", babystep (or "more") and "stop")

radry:
@digant73 Thank you for your suggestion.  

Placing the Temperature in the Heat menu (and similars) on the bottom is possible and I agree it would look cleaner but it will put the "back" button in a different location than the main menu (and most other menus). The benefit of the current implementation is, that the back button is always in the same position, regardless of the menu. 
For example you can hit "back" in the heat menu and afterwards hit "back" in the main menu in the same location. Personally I prefer the possibility to hit "back" quickly in succession without having to switch finger position every time.
I think this will lower the risk of accidently pressing a wrong button. 

This is surely a point worth discussing.

TFT/src/User/API/menu.c
```diff
@@ -184,21 +370,34 @@ const GUI_RECT rect_of_keyPS_end[] = {
...
   // infobox
-  {1*ICON_WIDTH+1*SPACE_X+START_X,  1*ICON_HEIGHT+1*SPACE_Y+ICON_START_Y,  3*ICON_WIDTH+2*SPACE_X+START_X,  2*ICON_HEIGHT+1*SPACE_Y+ICON_START_Y},
+  {1*ICON_WIDTH+1*SPACE_X+START_X,  1*ICON_HEIGHT+1*SPACE_Y+ICON_START_Y,  3*ICON_WIDTH+2*SPACE_X+START_X,  2*ICON_HEIGHT+1*SPACE_Y+ICON_START_Y}
+#endif
 };
 
 const GUI_RECT rect_of_keyListView[ITEM_PER_PAGE + 1] = {
-  // 8 icons area
+  #ifdef PORTRAIT
+  // list area VERTICAL GUI
```
digant73, line 379: four spaces, not two

TFT/src/User/API/menu.c
```diff
@@ -184,21 +370,34 @@ const GUI_RECT rect_of_keyPS_end[] = {
...
   {START_X,  2*LISTITEM_HEIGHT+ICON_START_Y,  LISTITEM_WIDTH+START_X,  3*LISTITEM_HEIGHT+ICON_START_Y},
   {START_X,  3*LISTITEM_HEIGHT+ICON_START_Y,  LISTITEM_WIDTH+START_X,  4*LISTITEM_HEIGHT+ICON_START_Y},
   {START_X,  4*LISTITEM_HEIGHT+ICON_START_Y,  LISTITEM_WIDTH+START_X,  5*LISTITEM_HEIGHT+ICON_START_Y},
-
+  // navigation button area
+  {START_X,  5*LISTITEM_HEIGHT+ICON_START_Y,  START_X+1*LIST_ICON_WIDTH,  5*LISTITEM_HEIGHT+ICON_START_Y+1*LIST_ICON_HEIGHT},
+  {START_X+1.5*LIST_ICON_WIDTH,  5*LISTITEM_HEIGHT+ICON_START_Y,  START_X+2.5*LIST_ICON_WIDTH,  5*LISTITEM_HEIGHT+ICON_START_Y+1*LIST_ICON_HEIGHT},
+  {LCD_WIDTH-START_X-1*LIST_ICON_WIDTH,  5*LISTITEM_HEIGHT+ICON_START_Y,  LCD_WIDTH-START_X,  5*LISTITEM_HEIGHT+ICON_START_Y+1*LIST_ICON_HEIGHT},
+  #else 
+  // list area LANDSCAPE GUI
```
digant73, line 390: four spaces, not two

TFT/src/User/API/menu.c
```diff
@@ -265,8 +464,24 @@ void menuClearGaps(void)
 
 #else
 
+//clear gaps all other screents except TFT70
 void menuClearGaps(void)
 {
+#ifdef PORTRAIT
```
digant73, line 470: you should indent this #if, #else block

digant73 (changes requested):

digant73:
@radry just a brief review.
Also I would consider to move the info area always on top (in all menus). The back button will be moved at the bottom (last row, third column).
That means for example, the Heat menu could be:
row 1: Temp info (infobox)
row 2: -, unit, +
row 3: nozzle, stop, back

radry:
I reworked the UI and also implemented @digant73 suggestion:

- The heat menu (and similar) has the temperature on the top.
- Thus I also changed the Status Menu to use such a layout.
- The main menu (and sub menus) have the bottom row icons now aligned left/right. This is done to have the back button always on the bottom right (same as heat menu).

HOWEVER:
Some sub menus like "move" and "level" should be reworked because the order of the icons is still not tidy. They work fine in the current state but are not pretty.
This must be done in each sub menus .c file by adjusting the order (of drawing and the key query) when compiled for portrait.
I don't know yet how many menus are affected. It will take some time to test them all.

How is this? If you like it I will push it to the PR after cleaning up the code as mentioned by @digant73
Please excuse the blurry pics.
**Status Screen**
Status_new
**Heat Menu**
heat_new
**Main Menu**
main_new

**Current state of "Move"**
move_new
**Current state of "Level"**
level_new

labotecno:
nice ! i like it

digant73:
@radry The menus with the infobox area (e.g. Heat, Status Screen etc...) are now OK.
For the menus with no infobox (e.g. Main, Move, Leveling etc...) it should be enough to simply remap icon 4 on the standard layout (non portrait mode) to icon 7 on portrait mode. That means you will have:

Main menu:
row 1: Heat/Fan, Movement, Un(Load)
row 2: Terminal, Custom, Settings
row 3: Em. Stop, , Back

Move menu:
row 1: Z-, Y+, Z+
row 2: X-, Y-, X+
row 3: 0.1mm, , Back

Level menu:
row 1: Point 4, Point 3, Distance
row 2: Point 1, Point 2, Point 5
row 3: Disarm XY, , Back

radry:
@digant73 
> Main menu:
row 1: Heat/Fan, Movement, Un(Load)
row 2: Terminal, Custom, Settings
row 3: Em. Stop, , Back

> Move menu:
row 1: Z-, Y+, Z+
row 2: X-, Y-, X+
row 3: 0.1mm, , Back

> Level menu:
row 1: Point 4, Point 3, Distance
row 2: Point 1, Point 2, Point 5
row 3: Disarm XY, , Back

That's already how it is, look at the screenshots.  

> Fort the Printing menu, please provide a screenshot (and possibly consider to apply the changes in my first post).

working on it

>Also, the portrait mode is an alternative to the standard mode, so I would convert all the "#ifdef PORTRAIT...#else...#endif" to "#ifndef PORTRAIT...#else...#endif"

The code for "TFT70_V3_0" is also in there with `#ifdef TFT70_V3_0` instead of `#ifndef` although it's not the default screen. I think it's easier to read and understand with `#ifdef` and it follows the already exising coding standard of TFT70_V3_0.

radry:
New printing screen  

printing_new

digant73:
@radry I would update the pictures in the main section with the final layouts.
In the Printing menu I would also move the progress bar just above the Pause/More/Stop buttons just to separate the live icons / stats from the standard control buttons

radry:
Progress Bar on top now. 
Looks better than dividing the icons and easier to implement than having it between bottom row and Babystep button.
printing_new_1

guruathwal:
@radry Nice work., but I am afraid to say that this implementation will be incompatible with the new menu API. Hardcoded menu layouts are being replaced with an easy-to-use grid system for all standard menus in my upcoming PRs.

radry:
@guruathwal 
Then you should implement the vertical GUI in the new menu API as well. 
I don't know how that new API works but since the hardcoded vertical layout follows the same rules and coding style as the landscape one, it should be easy to translate to your new API. I might be able to help.

TFT/src/User/Menu/StatusScreen.c
```diff
@@ -340,7 +340,7 @@ void drawStatus(void)
   GUI_SetTextMode(GUI_TEXTMODE_NORMAL);
   GUI_SetColor(GANTRYLBL_COLOR);
   GUI_SetBkColor(infoSettings.status_xyz_bg_color);
-  sprintf(tempstr, XYZ_STATUS, coordinateGetAxisActual(X_AXIS), coordinateGetAxisActual(Y_AXIS),
+  sprintf(tempstr, XYZ_STATUS, coordinateGetAxisActual(X_AXIS), coordinateGetAxisActual(Y_AXIS),
```
digant73, line 343: remove the trailing space in this line

digant73 (changes requested):

radry:
Build Test Fails because #2281 broke this PR. I already resolved the conflicts but haven't added #2281 to my fork.

@bigtreetech Please let me know what else you require to merge this PR.

TFT/src/User/Menu/Move.c
```diff
@@ -28,17 +28,23 @@ void storeMoveCmd(AXIS xyz, int8_t direction)
 
 void drawXYZ(void)
 {
-  char tempstr[20];
+  char tempstr[20];
```
digant73, line 31: remove trailing space

TFT/src/User/Menu/StatusScreen.c
```diff
@@ -323,13 +337,12 @@ void drawStatus(void)
     drawLiveText(3, &lvIcon, &SpeedItems[currentSpeedID]);
   #endif
 
-  GUI_SetTextMode(GUI_TEXTMODE_NORMAL);
+  GUI_SetTextMode(GUI_TEXTMODE_NORMAL);
```
digant73, line 340: remove trailing space

TFT/src/User/API/menu.c
```diff
@@ -3,14 +3,52 @@
 #include "ListItem.h"
 #include "Notification.h"
 
+
```
digant73, line 6: remove this empty line

TFT/src/User/API/menu.c
```diff
@@ -448,7 +709,7 @@ void busyIndicator(SYS_STATUS status)
 {
   if (status == STATUS_BUSY)
   {
-    GUI_SetColor(MENU_BUSY_DOT_COLOR);
+    GUI_SetColor(YELLOW);
```
digant73, line 712: restore GUI_SetColor(MENU_BUSY_DOT_COLOR);

digant73 (changes requested):

TFT/src/User/Menu/Move.c
```diff
@@ -44,7 +45,7 @@ void drawXYZ(void)
     sprintf(tempstr, "Z:%.2f  ", coordinateGetAxisActual(Z_AXIS));
     GUI_DispString(START_X + 3 * SPACE_X + 3 * ICON_WIDTH, (ICON_START_Y - BYTE_HEIGHT) / 2, (uint8_t *)tempstr);
   #endif
-
+
```
digant73: remove trailing space

TFT/src/User/Menu/StatusScreen.c
```diff
@@ -337,13 +337,12 @@ void drawStatus(void)
     drawLiveText(3, &lvIcon, &SpeedItems[currentSpeedID]);
   #endif
 
-  GUI_SetTextMode(GUI_TEXTMODE_NORMAL);
-  GUI_SetColor(GANTRYLBL_COLOR);
-  GUI_SetBkColor(infoSettings.status_xyz_bg_color);
+  GUI_SetTextMode(GUI_TEXTMODE_NORMAL);
```
digant73, line 340: remove trailing space

digant73 (changes requested):

Issue #2272, opened 2021-11-28
[FR] Minor UI changes


xPew:
To better suit my theme, I would like to ask for some minor UI changes but would be a nice touch.
Here, I would like to change the status screen text color to white. Can it be added as an option to the config.ini? My theme has a color scheme due to how our web interface looks, so I don't want to change it to a lighter background if it's possible to change the text color.
I guess nobody wants to use fancy colors here, so only a black and white option is enough, not to consume a lot of memory.
![image](https://user-images.githubusercontent.com/81440138/143775887-f3626ad7-8a9d-42cb-8a46-ea62852bca4d.png)

My other request is to @kisslorand . I really like your PR's, especially the status bar you added back then. May I ask you about the possibilities of its customization? I am not a programmer, and I guess it all comes down to memory issues, but I would like to also tweak this part of the UI to better suit my theme.  Could some color choices be added for this part of the UI too? Better in config.ini, but if it's not possible maybe some guidance about what to change in case of self-compiling to be able to customize this.

Thank you in advance! :)

radek8:
All you need is an adjustment

![image](https://user-images.githubusercontent.com/46979052/143781558-0adde1ed-7fba-45c2-9d61-3d6b202951e8.png)

kisslorand:
> My other request is to @kisslorand . I really like your PR's, especially the status bar you added back then. May I ask you about the possibilities of its customization? I am not a programmer, and I guess it all comes down to memory issues, but I would like to also tweak this part of the UI to better suit my theme. Could some color choices be added for this part of the UI too? Better in config.ini, but if it's not possible maybe some guidance about what to change in case of self-compiling to be able to customize this.
> 
> Thank you in advance! :)

Thank you for your appreciation but the status bar/icon is not my work. Anyhow What customization do you have in mind? What color choices and where do you have in mind?

xPew:
> > My other request is to @kisslorand . I really like your PR's, especially the status bar you added back then. May I ask you about the possibilities of its customization? I am not a programmer, and I guess it all comes down to memory issues, but I would like to also tweak this part of the UI to better suit my theme. Could some color choices be added for this part of the UI too? Better in config.ini, but if it's not possible maybe some guidance about what to change in case of self-compiling to be able to customize this.
> > Thank you in advance! :)
> 
[...]

PR #2275, opened 2021-11-29
Progress bar color

kisslorand:
### Description

Color choices added for progress bar during print.
The choices are: 
 - Orange
 - Yellow
 - Red
 - Green
 - Blue
 - Cyan
 - Magenta
 - Purple
 - Lime
 - Gray

The color of the progress bar can be set at compile time in the "Configuration.h" by setting `PROGRESS_BAR_COLOR`.

### Related Issues

#2272

### DIFF
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -149,13 +149,6 @@
 #define TERMINAL_COLOR_SCHEME 0  // Default: 0
 
 /**
- * Progress bar color
- * The color of the progress bar during print
- *   Options: [Orange: 0, Yellow: 1, Red: 2, Green: 3, Blue: 4, Cyan: 5, Magenta: 6, Purple: 7, Lime: 8, Gray: 9]
- */
-#define PROGRESS_BAR_COLOR 0  // Default: 0
-
-/**
  * Notification Style For ACK Messages
  * Set the notification style to use for displaying the ACK messages which start with "echo:".
  *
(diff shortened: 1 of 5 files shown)


xPew:
@kisslorand Thank you for this PR. It changes the color of the progress bar indeed. A config.ini option would be really nice, but it all comes down to memory :/
Also changes this line what radek suggested and it solves my other problem. Do you know a more permanent fix for this one by any chance? Maybe it consumes less memory?
image
I am willing to listen to the programmers, I just ask these for the basic users :D

TFT/src/User/Configuration.h
```diff
@@ -148,6 +148,13 @@
  */
 #define TERMINAL_COLOR_SCHEME 0  // Default: 0
 
+/**
+ * Progress bar color
+ * The color of the progress bar during print
+ *   Options: [Orange: 0, Yellow: 1, Red: 2, Green: 3, Blue: 4, Cyan: 5, Magenta: 6, Purple: 7, Lime: 8, Gray: 9]
+ */
+#define PROGRESS_BAR_COLOR 0  // Default: 0
+
```
digant73, line 157: you should move that setting after TERMINAL_KEYBOARD_LAYOUT (so in the param configurable at compile time).
Also please restore //#define MARKED_PROGRESS_BAR  // Default: commented (disabled). IOn this case // is not used for a comment but to disable a macro (or in general an instruction)

TFT/src/User/SanityCheck.h
```diff
@@ -51,6 +51,18 @@ extern "C" {
...
+    #error "PROGRESS_BAR_COLOR cannot be greater than 9"
+  #endif
+
+  #if PROGRESS_BAR_COLOR < 0
+    #error "PROGRESS_BAR_COLOR cannot be less than 0"
+  #endif
+#else
+  #define PROGRESS_BAR_COLOR 0
+#endif
+
```
digant73, line 65: same as in Configuration.h. It should be moved after TERMINAL_KEYBOARD_LAYOUT

digant73 (changes requested):

TFT/src/User/Configuration.h
kisslorand: Done.

TFT/src/User/SanityCheck.h
kisslorand: Done.

PR #2276, opened 2021-12-01
Optional value change on the status screen

### LINKED ISSUE #2272: [FR] Minor UI changes

radek8:
Added the ability to change the color of the values on the live icons on the status screen in the configuration.h file.
Some topics require the inverse display of color values on live status screen icons.
For example: THEME_Rep Rap Firmware Dark
resolves #2272

### DIFF
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -1267,10 +1267,10 @@
 /**
  * The color of the values on the status screen
  * Some topics require the inverse display of color values on live status screen icons.
- * For example: THEME_Rep Rap Firmware Dark  - options 1
- *   Options: [ Black: 0, White: 1 ]
+ * Uncomment for example: THEME_Rep Rap Firmware Dark 
+ * Comment for standard color
  */
-#define VALUE_COLOR 0 // Default: 0
+//#define VALUE_COLOR // Default: commented (disabled)
 
 /**
  * Progress Bar Layout (Printing menu)
--- a/TFT/src/User/Configuration.h
+++ b/TFT/src/User/Configuration.h
@@ -1265,6 +1265,14 @@
 #define PROGRESS_BAR_COLOR 0  // Default: 0
 
 /**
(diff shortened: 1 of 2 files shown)


TFT/src/User/Configuration.h
```diff
@@ -1264,6 +1264,14 @@
...
 #define PROGRESS_BAR_COLOR 0  // Default: 0
 
+/**
+ * The color of the values on the status screen
+ * Some topics require the inverse display of color values on live status screen icons.
+ * Uncomment for example: THEME_Rep Rap Firmware Dark 
+ * Comment for standard color
+ */
+//#define VALUE_COLOR // Default: commented (disabled)
+
```
kisslorand, line 1274: I would put this before PROGRESS_BAR_COLOR or after MARKED_PROGRESS_BAR so progress bar stuff remains together.

kisslorand (changes requested):

TFT/src/User/Configuration.h
radek8: OK

kisslorand: Thx

xPew:
Nice! Appreciate it!

PR #2281, opened 2021-12-04
Customizable hard coded colors moved to common file LCD_Colors.h

digant73:
All customizable hard coded colors used in menus moved to common file LCD_Colors.h.
This will allow users to customize colors applying changes only on that file (see it as the equivalent of Configuration.h used only for colors).

**PR STATE:** ready for merge


xPew:
Neat!

PR #2284, opened 2021-12-09
Configurable common file extension suppression + bugfixes

digant73:
**IMPROVEMENTS:**
* Configurable suppression (hide) for common gcode file extension: new param "filename_extension" added in config.ini. If set to 0 (false), any file extension starting with ".g" or ".G" (e.g. ".g", ".gco", ".gcode" etc...) will be hidden in both List Mode / Icon Mode menu. That setting can also be changed from Screen->UI menu. It implements #2074

**BUGFIXES:**
* Fixed missing display of filenames in sub folders when navigating from onboard SD on latest Marlin fw. It fixes #2182
* Minor bug fixes and optimization

resolves #2074
fixes #2182

**PR STATE:** ready for merge

Issue #2215, opened 2021-10-10
[FR] MULTI_VOLUME for onboard SD & onboard USB Support


Guilouz:
It will be interesting to add support for multi volume to choose which volume (SD or USB) to mount.

Usefull for new boards like SKR 2, GTR, Octopus etc

It added to Marlin menu with this PR : https://github.com/MarlinFirmware/Marlin/pull/22004

Multi volume are working in Marlin mode but not on TFT mode.
In Marlin mode print menu gives access to the ability to browse onboard SD storage and onboard USB storage on the motherboard.
In TFT mode we have only access to the onboard USB storage with OnboardSD icon but not to SD storage when MULTI_VOLUME is enabled in Marlin.

kisslorand:
How is this different from FR #2005?

kisslorand:
Multivolume is available only in Marlin mode. In touch mode there is no other way to communicate with Marlin than with gcodes. For the moment there is no gcode in Marlin for switching from onboard SD to onboard USB.

Guilouz:
> Multivolume is available only in Marlin mode. In touch mode there is no other way to communicate with Marlin than with gcodes. For the moment there is no gcode in Marlin for switching between onboard SD and onboard USB.

I understand, but there is some issue when USB Support is enabled, TFT always reports that there is no SD card.

kisslorand:
I just tested it today. By default USB is used, you have to switch to SD to access it.

Guilouz:
Yes USB is working but there is always "no Sd card" notification which repeats in TFT mode but not in Marlin mode. It's quite boring.

kisslorand:
I didn't have that "No SD card" when I tested today onboard USB.

kisslorand:
@Guilouz 
I started working on this. For the moment I implemented M19 in Marlin and with that gcode I can switch between SD and USB on the motherboard.
M19 S - switch to SD card
M19 U - switch to USB

Guilouz:
Ah good news ! I think a 2nd icon onboardUSB would be good, to keep the continuity of the screen ports.
Yes I think an option in config.ini to enable it because I think there is no option to parce if MULTI_VOLUME is enabled or not.

kisslorand:
You just gave me an idea, to make Marlin report its MULTI_VOLUME capability in case SDCARD is enabled.

kisslorand:
https://github.com/MarlinFirmware/Marlin/pull/23780

[...]

PR #2291, opened 2021-12-18
File extension -> more; bugfixes, improvements

### LINKED ISSUE #2074: [FR] hide/remove file extension (.gcode)
Since most screens are relatively small, the file names can get very long.
I do not see any benefit of displaying the file type/extension.

I think it would look even cleaner, if the extention **`.gcode`** was hidden.

file name before   |   |  file name after
-|-|-
*example.gcode*|**=>**|*example*

### LINKED ISSUE #2295: [BUG] (short description)
### Steps to reproduce

1. Try print with Sd TFT port
2. Try print with Isb TFT port

**Expected behavior**

**Actual behavior**

### Additional Information
* Hello,

first i'm sorry for my english but i'm french and again sorry if i'm in the wrong section but this is the first time i've used a request on github.

I recently installed on my FlSun super Racer a Skr v2 card and a Tft70 from Biqu. However since that time I have had bugs in my printing sometimes the printer stops with the M21 error or the M1 error without having configured a pause in the Slicer. It has also happened to me that printing stops without error on the TFT. I have tried printing with the SD card as well as the USB from the TFT but I have the same concerns however I have no problem securing the USB stick on the motherboard. I have the latest version of TFT as well as the latest version on the motherboard would you have any idea of ​​the problem thanks in advance.

### LINKED ISSUE #2319: [BUG] Strange behavior when setting the fan.
I am using fan controllers (M710)
The fan is set to 66%.
If I open the fan menu, the fan (CtA) displays the speed 66/00. It does not respond when you click to lower the values. By clicking on the values increase, the values are set to 1/1. And the fan stops

If I edit the value (F0) before editing the value of the controller fan (CtA), then I go to the value (CtA) it still behaves badly.
if I leave the menu and return again, values (CtA) 66/66 is displayed and editing works normally.

The fan setting (F0) also does not work properly.
If I set the fan speed via Gkod (M106 S128)
The value (F0) 50/0 is displayed in the fan menu.
The value cannot be decreased and when increased, it is calculated from the beginning of 1/1. And the fan stops.

For clarity, I enclose a link to the video.

https://youtu.be/qP6O3oVJ3eM
https://youtu.be/cs8mnnaL1KQ

PS:
The print speed menu works as expected.
After opening the menu, the speed 100/100 is displayed and can be increased or decreased

kisslorand:
### Requirements
BTT or MKS TFT

### Description
- Takes #2284 further and applies file extension suppression (if configured so) in printing menu too
- Removed the lowercase conversion function defined in the FW and switched to the one that is already present in the "string" library.
- Fixes a bug that made some list type menu's statusbar not to refresh (ex. file list menu).
- Fixed a bug with M0, M1 when printing from TFT (SD or USB). In case of M0 or M1 present in the gcode file, upon resume it would dive the nozzle into the bed - and make a nice tattoo there
- Fixed a bug in "printAbort()" function, a loop avoidance was implemented there but it was inactive
- Fixed Fan menu odd behaviour #2319
- Fixes a GUI bug when aborting a print from onboard SD (explained here: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291#issuecomment-1013685622)
- Added long folder name when printing from onboard SD
- Fixes a bug in the file list when selecting files from onboard SD. Some files, even thou they had long file names, the short file name was displayed instead (example here: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291#issuecomment-1013867722)
- Fixes a bug in the file list when selecting files from onboard SD. A file could show up in a folder where it didn't belong to. (example here: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291#issuecomment-1013867722)
- Fixes sending more commands at once without knowing if there are available slots for all of them in the motherboard's buffer. It guards against buffer overflow on the motherboard side and loss of commands from the TFT.
- Most probably fixes issue #2295 (thx to @radek for his hint about M220/M221)
- Some degree of code reductiom

### Related issues, feature requests

Fixes #2074
Fixes #2319
(most probably) Fixes #2295

### Prebuilt binaries

Prebuilt binaries with this PR are ready and can be downloaded here: https://mega.nz/folder/Hx4mXKjY#WFvcTZWweq_VRW9pR7nr_g
They are aligned to master as of 22.1.2022.
Any feedback from real life usage is welcome.


digant73:
I think you didn't consider the goal of PR #2284 and all the possible covered use cases avoiding any possible error/crash (wrong access to a string, use of extra byte for extension check, what happens in case of missing extension, make hiding and restoring totally transparent reducing the code etc...). The suppression of extension is needed only in the Print menu. You can have files with .gcode extension but possibly they are not gcode file and the opposite (long file names provided by Marlin can have missing extensions), etc... Furthermore, instead of hiding as much as possible any checks, all the checks are now duplicated in the code.
Please, consider to rollback those changes. I understand it's frustraiting for you to receive such kind of request but it's even more annoying for me to see someone changing the code correctly made in something that is not better.

kisslorand:
Can you give me please one single real-life example where the changes made will fail?
It seems you didn't fully checked what was done and because of it you make false assumptions. Until you have an example where these changes will fail and they wouldn't without these changes, please stop with the theories.


> The suppression of extension is needed only in the Print menu.

Who and when decided it? I am not aware of such regulations made public.

> You can have files with .gcode extension but possibly they are not gcode file and the opposite (long file names provided by Marlin can have missing extensions), etc...

If you have .gcode files that are not actually Gcode files nothing new will happen, everything will be the same as before this PR.  If Marlin has missing extensions, nothing new will happen, everything will be exactly the same as before this PR, those files will not be present in the filelist of the TFT's menu.


> Furthermore, instead of hiding as much as possible any checks, all the checks are now duplicated in the code.

Exactly! It's easier to debug and easier to understand. At runtime the number of checks haven't been increased.


> it's even more annoying for me to see someone changing the code correctly made in something that is not better

I am sorry you feel that way. I wrote it as clear as possible that the methods used were not suitable for the new tasks. I do not consider the new methods any better, just different and suitable for the tasks needed. Period.

I really hope a real example can and will be provided to show the supposed flaws of this PR so I can correct them.
Please, please, please no more theories...

digant73:
Marlin/RRF already filters the list provided to TFT (maybe even based on file content instead of file extension) so the TFT does not need to apply any additional filter (that means you can have filenames with an extension not matching the ones provided by TFT).
Furthermore, long filename can have no extension at all (no "." present) while its short name could not have it or it could have it but it can differ than the extension supported by TFT (.g, .gco etc...).
What happens when function restoreFileExtension is invoked for a filename if hideFileExtension previously didn't remove any extension? All the checks currently present in the existing hideFileExtension/restoreFileExtension are needed to cover all the sources for the filename and long filename (if any). You removed all the checks on the "extension flag" introduced by PR #2284.
I hope it's clear now

kisslorand:
> Marlin/RRF already filters the list provided to TFT (maybe even based on file content instead of file extension) so the TFT does not need to apply any additional filter (that means you can have filenames with an extension not matching the ones provided by TFT).
> Furthermore, long filename can have no extension at all (no "." present) while its short name could not have it or it could have it but it can differ than the extension supported by TFT (.g, .gco etc...).

PR #2284 doesn't allow extension-less files to be displayed, hence they cannot be selected. This PR does the same.

> What happens when function restoreFileExtension is invoked for a filename if hideFileExtension previously didn't remove any extension?

Where is `restoreFileExtension` invoked without `hideFileExtension` beeing invoked beforehand? I can't find any such instance of  `restoreFileExtension`. Would you be so kind to point me to that direction so I can remedy it?

> All the checks currently present in the existing hideFileExtension/restoreFileExtension are needed to cover ALL the sources for the filename and long filename (if any).

You are right, they are probably necessary in the current master but it is no need for that in this PR. If I am wrong please provide me a real example (not a theoretical one) so I can rectify it.

> Please, duplication of code here and there never makes the code more easy to debug and understand.

I guess obfuscated codes are better than... Sorry, I beg to differ. 

The code is there, I do not see why I have to explain it. For all the questions the answer is in the code. So, once again, please provide real scenario examples not theoretical ones. 
"This is the file source, this is the filename, I press this and this and this and this happens".

I cannot stress enough that this PR is rock solid until proven otherwise with real example and not unrelated theories.

digant73:
> > Marlin/RRF already filters the list provided to TFT (maybe even based on file content instead of file extension) so the TFT does not need to apply any additional filter (that means you can have filenames with an extension not matching the ones provided by TFT).
> > Furthermore, long filename can have no extension at all (no "." present) while its short name could not have it or it could have it but it can differ than the extension supported by TFT (.g, .gco etc...).
> 
> PR #2284 doesn't allow extension-less files to be displayed, hence they cannot be selected. This PR does the same.
> 

see function scanPrintFilesGcodeFs

> > What happens when function restoreFileExtension is invoked for a filename if hideFileExtension previously didn't remove any extension?
> 
> Where is `restoreFileExtension` invoked without `hideFileExtension` beeing invoked beforehand? I can't find any such instance of `restoreFileExtension`. Would you be so kind to point me to that direction so I can remedy it?
> 

Try this:
1) consider a long filename provided by Marlin with no extension (e.g. "my_long_filename_with_no_extension")
2) apply hideFileExtension
3) apply restoreFileExtension
4) what is the filename after restoreFileExtension?

> > All the checks currently present in the existing hideFileExtension/restoreFileExtension are needed to cover ALL the sources for the filename and long filename (if any).
> 
> You are right, they are probably necessary in the current master but it is no need for that in this PR. If I am wrong please provide me a real example (not a theoretical one) so I can rectify it.
> 
> > Please, duplication of code here and there never makes the code more easy to debug and understand.
> 
> I guess obfuscated codes are better than... Sorry, I beg to differ.
> 
> The code is there, I do not see why I have to explain it. For all the questions the answer is in the code. So, once again, please provide real scenario examples not theoretical ones. "This is the file source, this is the filename, I press this and this and this and this happens".
> 
> I cannot stress enough that this PR is rock solid until proven otherwise with real example and not unrelated theories.

I gave above the real examples and all the theory behind #2284

kisslorand:
> Extension-less was also present before #2284

Really?

LongFilename
FileList


> Try this:
> 
> 1. consider a long filename provided by Marlin with no extension (e.g. "my_long_filename_with_no_extension")
> 2. apply hideFileExtension
> 3. apply restoreFileExtension
> 4. what is the filename after restoreFileExtension?

I can't. Marlin doesn't provide long filename without extension

Any other "real" examples?

digant73:
Here an example

[very_very_very_log_long_long_very_very_very_long_long_long_file.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/7741212/very_very_very_log_long_long_very_very_very_long_long_long_file.zip)

Furthermore, Marlin provides truncated long filename for a lot of files (even if the filename is pretty short).
I also have the latest Marlin fw.

As far as I see, "Very Long File TEST3.g" is already a real example of truncated filename.

Isn't it a real example? isn't it better to have a more robust code for all use cases and for all users (with current and old Marlin fw) instead of trying to make something more weak that will not work in some cases (as it happened with the missing files on subdirectories when printing from onboard)?

kisslorand:
> Here an example
> 
> [very_very_very_log_long_long_very_very_very_long_long_long_file.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/7741212/very_very_very_log_long_long_very_very_very_long_long_long_file.zip)

Aha! Now that's an example I didn't thought about! Good catch!
Back to work... :)

I have to think of a good check for previously hidden extension check. The one used in PR #2284 (`if (filename[strlen(filename) + 1] != 0)  // check extra byte for filename extension check. If 0, no filename extension was previously hidden`) doesn't stand.
With the last example in the ZIP file this check will horribly fail.

kisslorand:
@digant73 
Check the new approach pls.

digant73:
The extension flag is provided on arrays infoFile.file[] and infoFile.longFile[]. It is not present on infoFile.title because infoFile.title is created by EnterDir/ExitDir (that uses strcat).
The extension flag is never set on hideFileExtension/restoreFileExtension (it remains always set to 0) while in your hideFileExtension you set something to 1 followed by 0
The reason why I didn't apply hide/restore in the rest of the code is because it is better to see the complete name. For example we can have two files that differ only for the extension (e.g. one is .g and the other is .gcode). At least we can identify the proper file in the the dialog box asking to confirm to proceed with the print. Removing the extension everywhere makes impossible to distinguish which file we are going to print. This is also the approach used on any OS (in the resource explorer you can suppress the extension but when you double click the file the complete name is reported (e.g. on Excel, Word etc...)

kisslorand:
I thought the same but it was past 2AM and I was already wasted. 
FInally made some changes that adds only the optional extensionless filename only to the printing screen. File confirmation and print summary is back to always show the extension.

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
 }
 
 // check if filename provides a supported filename extension
-char * isSupportedFile(char * filename)
```
digant73, line 135: not my usual coding style but in those file it was used the style providing a space for the type (e.g. char * filename instead of char *filename). I would restore (maintain the same style as it was before) the previous style (so char * isSupportedFile(...) instead of char *isSupportedFile(...)

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
+char *isSupportedFile(char * filename)
```
digant73, line 135: same as previous comment

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
 {
-  char * extPos = strstr(filename, ".g");  // support "*.g","*.gco" and "*.gcode"
+  char *extPos = strrchr(filename, '.');  // check last "." in the name where extension is supposed to start
```
digant73, line 137: same as previous comment

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
 
-  if (extPos == NULL)
-    extPos = strstr(filename, ".G");  // support "*.g","*.gco" and "*.gcode"
+  if (extPos != NULL && extPos[1] != 'g' && extPos[1] != 'G')
+  {
```
digant73, line 140: I would prefer to avoid the use of {} with a single line. but this is only my preferred style

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
+  }
 
   return extPos;
 }
-
-char * hideFileExtension(uint8_t index)
+/**
+  * @brief  Temporarily remove extension 
+  * @param  filename: targeted filename
+  * @param  maxLength: the max length of the "filename" parameter as it was defined.
```
digant73, line 149: there is no more param maxLength

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
+  *         It is needed to check if extension hide flag can be added.
+  * @retval pointer to the resulting name
+  */
+char *hideFileExtension(char * filename)
```
digant73, line 153: same as one of previous comment

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
-  char * filename = infoFile.file[index];
   char * extPos;
 
-  if (infoSettings.filename_extension == 0)  // if filename extension is disabled
+  extPos = isSupportedFile(filename);
+  if (extPos != NULL)  // if filename provides a supported filename extension
+  {
+    filename[extPos - filename] = 0;  // temporarily hide filename extension
+  }
+  else
```
digant73, line 162: as reported before, no need (and it could cause a bug) to set extension flag here. The "else" statement must be removed

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
+  
+  return filename;
+}
 
-    if (extPos != NULL)  // if filename provides a supported filename extension
-      filename[extPos - filename] = 0;  // temporary hide filename extension
+/**
+  * @brief  Restore the temprarily hidden extension 
+  * @param  filename: targeted filename
+  * @param  maxLength: the max length of the "filename" parameter as it was defined.
```
digant73, line 173: there is no more param maxLength

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
+  *         It is needed to check if extension hide flag can be retrieved.
+  * @retval pointer to the resulting name
+  */
+char *restoreFileExtension(char *filename)
```
digant73, line 177: same as one of previous comment (also for param *filename)

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
+  if (filename[strlen(filename) + 1] != 0)  // check extra byte for filename extension check. If 0, no filename extension was previously hidden
+  {
+    filename[strlen(filename)] = '.';       // restore filename extension
   }
 
-  if (infoMachineSettings.longFilename == ENABLED && infoFile.source == BOARD_SD)
+  return filename;
+}
+
+char *hideFileIndexExtension(uint8_t index)
```
digant73, line 187: same as one of previous comment

TFT/src/User/API/Vfs/vfs.c
```diff
@@ -132,63 +132,88 @@ bool IsRootDir(void)
...
+      filename = infoFile.longFile[index];
+      hideFileExtension(filename);
     }
   }
 
   return filename;
 }
 
-char * restoreFileExtension(uint8_t index)
+char *restoreFileIndexExtension(uint8_t index)
```
digant73, line 205: same as one of previous comment

TFT/src/User/API/Vfs/vfs.h
```diff
@@ -50,9 +50,11 @@ void resetInfoFile(void);
 bool EnterDir(char * nextdir);
 void ExitDir(void);
 bool IsRootDir(void);
-char * isSupportedFile(char * filename);
-char * hideFileExtension(uint8_t index);
-char * restoreFileExtension(uint8_t index);
+char *isSupportedFile(char * filename);
```
digant73, line 53: same as on of previous comment

TFT/src/User/Fatfs/myfatfs.c
```diff
@@ -116,7 +116,6 @@ bool scanPrintFilesFatFs(void)
       fileDate[infoFile.fileCount] = (finfo.fdate * 100000) + finfo.ftime;
       // copy file name
       strcpy(infoFile.file[infoFile.fileCount], finfo.fname);
-      infoFile.file[infoFile.fileCount][len] = 0;  // set to 0 the extra byte for filename extension check
```
digant73, line 119: same as one of previous comment

TFT/src/User/API/menu.c
```diff
@@ -819,7 +819,16 @@ void menuDrawTitle(const uint8_t *content)
   GUI_SetBkColor(infoSettings.title_bg_color);
   if (content)
   {
-    GUI_DispLenString(10, start_y, content, LCD_WIDTH - 20, true);
+    if (MENU_IS(menuPrinting) && infoSettings.filename_extension == 0)
+    {
+      hideFileExtension((char *)content);
```
digant73, line 824: you cannot use hideFileExtension/restoreFileExtension on the general function menuDrawTitle. "content" is a general string so it does not provide any extension flag

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -119,7 +119,6 @@ bool scanPrintFilesGcodeFs(void)
           break;
         }
         strcpy(infoFile.longFile[infoFile.fileCount], Pstr_tmp);
-        infoFile.longFile[infoFile.fileCount][strlen(Pstr_tmp) + 1] = 0;  // set to 0 the extra byte for filename extension check
```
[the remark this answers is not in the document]
kisslorand: Done it, my way. No code increase.

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -150,13 +149,12 @@ bool scanPrintFilesGcodeFs(void)
       }
 
       strcpy(infoFile.file[infoFile.fileCount], file);
-      infoFile.file[infoFile.fileCount][strlen(file) + 1] = 0;  // set to 0 the extra byte for filename extension check
```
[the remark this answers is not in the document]
kisslorand: Done it, my way. No code increase.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Done.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Done.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Done.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Done.

TFT/src/User/API/Vfs/vfs.c
kisslorand: You can avoid it, no problem.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Fixed.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Done.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Fixed.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Done.

TFT/src/User/API/Vfs/vfs.h
kisslorand: Done.

TFT/src/User/API/Vfs/vfs.c
kisslorand: Done.

TFT/src/User/Fatfs/myfatfs.c
kisslorand: Done it, my way. No code increase.

TFT/src/User/API/menu.c
kisslorand: > you cannot use hideFileExtension/restoreFileExtension on the general function menuDrawTitle. "content" is a general string so it does not provide any extension flag.

There's a conditional there. Where does "counter" points when the conditional is true? To what "string"?

> If you want to hide the file extension reported in the title bar (e.g. on Print or PrintingMenu) you can simply make a copy of infoFile.title first and then apply hideFileExtension but not restoreFileExtension. E.g.:
> 
> char copy_of_title[MAX_PATH_LEN];
> 
> strcpy(copy_of_title, infoFile.title); // infoFile.title has the file extension and no extension flag hideFileExtension(copy_of_title); // copy_of_title has no more file extension and no extension flag menuDrawTitle((uint8_t *)copy_of_title);

Thanks, there are more ways to... peal an orange.

> // in any case don't call restoreFileExtension(copy_of_title) or restoreFileExtension(infoFile.title) because it will access/check a missing extension flag

Sure, thx.

> * rollback the changes made in menu.c

Sorry, I won't. There is some bugfix too.

> * remove restoreFileExtension on vfs.h (it will be used only internally by restoreFileIndexExtension).

Nope

> * restore the previous names for hideFileIndexExtension/restoreFileIndexExtension ("Index" doesn't mean anything for me)
> * according to previous bullet, rename hideFileExtension (e.g. hideExtension)

Done.

TFT/src/User/API/menu.c
```diff
@@ -819,7 +819,16 @@ void menuDrawTitle(const uint8_t *content)
   GUI_SetBkColor(infoSettings.title_bg_color);
   if (content)
   {
-    GUI_DispLenString(10, start_y, content, LCD_WIDTH - 20, true);
+    if (MENU_IS(menuPrinting) && infoSettings.filename_extension == 0)
+    {
+      hideExtension((char *)content);
```
digant73, line 824: function menuDrawTitle is a general function to display any kind of text in the title bar (not only a filename). E.g. it is called also by ListManager, ScreenSetting, Speed etc... None of those is passing a string with an additional extension flag at the end, so the restoreExtension will provide an out of buffer read in those use cases

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -112,14 +112,14 @@ bool scanPrintFilesGcodeFs(void)
...
+        infoFile.longFile[infoFile.fileCount] = malloc(strlen(Pstr_tmp) + 2);  // "+ 2": space for terminating null character and the flag for filename extension check
         if (infoFile.longFile[infoFile.fileCount] == NULL)
         {
           clearRequestCommandInfo();
           break;
         }
-        strcpy(infoFile.longFile[infoFile.fileCount], Pstr_tmp);
-        infoFile.longFile[infoFile.fileCount][strlen(Pstr_tmp) + 1] = 0;  // set to 0 the extra byte for filename extension check
+        // copy file name and set the flag for filename extension check
+        strncpy(infoFile.longFile[infoFile.fileCount], Pstr_tmp, strlen(Pstr_tmp) + 2);  // "+2":  terminating null character and the flag for filename extension check
```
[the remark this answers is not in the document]
kisslorand: > this is not correct. you are not setting the extension flag.

Than how do you explain that this code in "myfatfs.c"
```
      infoFile.file[infoFile.fileCount][len] = 1;  // make sure file extension flag is NOT set 
      strncpy(infoFile.file[infoFile.fileCount], finfo.fname, len + 1);
```
and this code in "vfs.c"
```
char * hideExtension(char * filename)
{
  char * extPos = isSupportedFile(filename);

  if (extPos != NULL)  // if filename provides a supported filename extension
  {
    if (filename[strlen(filename) + 1] == '\0')
        filename[extPos - filename] = '*';  // temporarily hide filename extension
  }
  
  return filename;
}
```

resulted this: 
TerminalChar

Magic? Luck? I don't think so...
You probably do not understand how `strncpy` works. A quick search on Google yields this:
"The C library function `char *strncpy(char *dest, const char *src, size_t n)` copies up to n characters from the string pointed to, by src to dest. **In a case where the length of src is less than that of n, the remainder of dest will be padded with null bytes.**"

You can also look up `strncpy`  in the library itself.



> it is very dangerous to try to copy more data than Pstr_tmp really have (the max number of byte to copy must always be <= strlen(Pstr_tmp)

It's not how `strncpy` works, since "Pstr_tmp" is the source, not the destination.

TFT/src/User/Fatfs/myfatfs.c
```diff
@@ -114,9 +114,8 @@ bool scanPrintFilesFatFs(void)
 
       // copy date/time modified
       fileDate[infoFile.fileCount] = (finfo.fdate * 100000) + finfo.ftime;
-      // copy file name
-      strcpy(infoFile.file[infoFile.fileCount], finfo.fname);
-      infoFile.file[infoFile.fileCount][len] = 0;  // set to 0 the extra byte for filename extension check
+      // copy file name and set the flag for filename extension check
```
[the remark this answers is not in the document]
kisslorand: You mistake "index" with "size". For the rest check the answer above.

TFT/src/User/API/menu.c
kisslorand: You missed the point that `hideExtension((char *)content)` and `restoreExtension((char *)content)` is called only if the conditional ` if (MENU_IS(menuPrinting) && infoSettings.filename_extension == 0)` is true. Can you guess where does "content" points in that case?

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -140,23 +140,22 @@ bool scanPrintFilesGcodeFs(void)
...
       {
         if (infoFile.longFile[infoFile.fileCount] != 0)
           free(infoFile.longFile[infoFile.fileCount]);
 
         break;
       }
-
-      strcpy(infoFile.file[infoFile.fileCount], file);
-      infoFile.file[infoFile.fileCount][strlen(file) + 1] = 0;  // set to 0 the extra byte for filename extension check
+      // copy file name and set the flag for filename extension check
```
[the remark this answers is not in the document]
kisslorand: Check above.

Randwic:
Hello,

first i'm sorry for my english but i'm french and again sorry if i'm in the wrong section but this is the first time i've used a request on github.

I recently installed on my FlSun super Racer a Skr v2 card and a Tft70 from Biqu.  However since that time I have had bugs in my printing sometimes the printer stops with the M21 error or the M1 error without having configured a pause in the Slicer.  It has also happened to me that printing stops without error on the TFT.  I have tried printing with the SD card as well as the USB from the TFT but I have the same concerns however I have no problem securing the USB stick on the motherboard.  I have the latest version of TFT as well as the latest version on the motherboard would you have any idea of ​​the problem thanks in advance.

kisslorand:
@Randwic 
Yeah, you are right, this is a very wrong place to write about the SKR v2 problems. Right here we talk about a pull request, not issues.

Even if we were talking about issues, it would be about issues of the TFT's firmware, not the motherboard.
If it's any comfort for you, this M1 and M21 bug is more and more present for those who have SKR v2 (including myself, I also have this board with the same M1, M21 problems) but no one has any hint what causes it.

Randwic:
@kisslorand 

Sorry this is my first time on GitHub thanks for your reply

TFT/src/User/API/menu.c
digant73: You missed both my previous explanations. Even the Printing menu can pass a "standard" (no extension flag) string. And you also made a change in your PR where this happens. I suggested you to apply the hideExtension function directly in Printing menu (in the only point where you need it). It will be completely safe and logical and the code is even reduced in size (no need for extra checks, no need to invoke restoreExtension).
Sorry, but I cannot spend other time in such kind of infinite discussions

kisslorand: > Even the Printing menu can pass a "standard" (no extension flag) string. And you also made a change in your PR where this happens.

The only time menuDrawTitle() is called in printing menu is when menuDrawPage() is called in the printing menu. There menuDrawPage() is impossible to pass a pointer to "standard" (no extension flag) char array because "printingItems.title.address" will always point to "infoFile.longFile[infoFile.fileIndex]" or "infoFile.file[infoFile.fileIndex]". Both of them already are extension flagged strings.

> I suggested you to apply the hideExtension function directly in Printing menu (in the only point where you need it). It will be completely safe and logical and the code is even reduced in size (no need for extra checks, no need to invoke restoreExtension).

You might have suggested it but it  implies further complications. If I only apply the hideExtension function than infoPrintSummary will  have the filename without extension.  To avoid that than extension must be restored after menuDrawPage() is called. This has another drawback, when menuReDrawCurTitle() is called than it would display again the file with extension. 

> I cannot spend other time in such kind of infinite discussions

No need to. It seems all your valid worries and usable suggestions are settled now.

TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -592,7 +592,7 @@ void menuPrinting(void)
     if (infoMachineSettings.longFilename == ENABLED && infoFile.source == BOARD_SD)
```
digant73, line 592: As it was done in Print.c

in PrintingMenu.c:

```
    if (infoMachineSettings.longFilename == ENABLED && infoFile.source == BOARD_SD)
      printingItems.title.address = (uint8_t *) infoFile.longFile[infoFile.fileIndex];
    else
      printingItems.title.address = (uint8_t *) infoFile.file[infoFile.fileIndex];

```
is equivalent to:

`    printingItems.title.address = hideFileExtension(infoFile.fileIndex);`

in Printing.c

```
  if (infoMachineSettings.longFilename == ENABLED && infoFile.source == BOARD_SD)
    sprintf(infoPrintSummary.name,"%." STRINGIFY(SUMMARY_NAME_LEN) "s", infoFile.longFile[infoFile.fileIndex]);
  else
    sprintf(infoPrintSummary.name,"%." STRINGIFY(SUMMARY_NAME_LEN) "s", infoFile.file[infoFile.fileIndex]);
```

is equivalent to:

`  sprintf(infoPrintSummary.name,"%." STRINGIFY(SUMMARY_NAME_LEN) "s", hideFileExtension(infoFile.fileIndex));
`

As already said, in both cases there is no need to call restoreFileExtension and the original menuDrawTitle can be restored. Also hideExtension and restoreExtension will be no more exposed for public use

TFT/src/User/API/Gcode/mygcodefs.c
digant73, line 122: // "+ 2": space for terminating....     instead of  // "+2":  terminating....

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -140,23 +140,22 @@ bool scanPrintFilesGcodeFs(void)
...
+      strncpy(infoFile.file[infoFile.fileCount], file, strlen(file) + 2);  // "+2":  terminating null character and the flag for filename extension check
```
digant73, line 152: // "+ 2": space for terminating....     instead of  // "+2":  terminating....

TFT/src/User/Menu/Print.c
```diff
@@ -310,7 +310,7 @@ void menuPrintFromSource(void)
       }
       else
       { // title bar is also drawn by listViewCreate
-        listViewCreate((LABEL){.address = (uint8_t *)infoFile.title}, NULL, infoFile.folderCount + infoFile.fileCount,
+        listViewCreate((LABEL){.index  = LABEL_DYNAMIC, .address = (uint8_t *)infoFile.title}, NULL, infoFile.folderCount + infoFile.fileCount,
```
digant73, line 313: index = LABEL_DYNAMIC (one space before =)

TFT/src/User/API/Gcode/mygcodefs.c
kisslorand: Done

TFT/src/User/API/Gcode/mygcodefs.c
kisslorand: Done

TFT/src/User/Menu/Print.c
kisslorand: Done

TFT/src/User/Menu/PrintingMenu.c
kisslorand: This suggestion is wrong in many ways. 
- Hiding an extension for an already hidden file name can lead to unpredictable results (ex: "Cover.gembird.gcode"). Added an extra check to not hide extension successively.
- Hiding extension for print summary is the opposite of the intention. It was a consensus to show it.
- Leaving the filename in an unknown state is anything but safe. Better hide it only when needed and restore it so no surprises down the line. (See cause of bug #2300)

kisslorand:
Added two new bugfixes in Stage 2.8 commit:
- Fixed a bug with M0, M1. In case of M0 or M1 present in the gcode file, upon resume it would dive the nozzle into the bed and make a nice tattoo there
- Fixed a bug in "printAbort()" function, a loop avoidance was implemented there but it was inactive

Guilouz:
@kisslorand Do you have find the issue for M0/M1 ?

kisslorand:
Sorry, I didn't.

Guilouz:
Oh I thought your last stage was fixed that, shame.

kisslorand:
> Oh I thought your last stage was fixed that, shame.

Check https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/2295#issuecomment-1009441835

kisslorand:
Stage 2.9 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/a011780e6d6741989fa8916297a77e0beed194e6) fixes bug #2319.

radek8:
> Stage 2.9 (commit [a011780](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/a011780e6d6741989fa8916297a77e0beed194e6)) fixes bug #2319.

Thank you
The fix works

kisslorand:
Stage 2.9.1 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/36f1f310029d3d6f199c46039952a49331819caa) fixes a GUI bug when aborting a print from onboard SD.  The splash screen after the print abort confirmation had fake buttons. Those buttons were leftovers from the print abort confirmation popup.

Fake_Buttons

After the fix looks like this: 

No_Fake_Buttons

kisslorand:
Stage 3 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/510e4445d234dc2eaeed340fc5d1b8bca462030f) brings the display of the long name of folders when selecting files from onboard SD.

kisslorand:
Stage 3.1 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/c0168216e7b7e3ac6e9a25f3c4d6238f246e428c) extends the folder long names to subfolders too and does some memory optimization avoiding unnecessary memory usage when no long name exists for a folder.

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -172,11 +194,24 @@ bool scanPrintFilesGcodeFs(void)
...
-        infoFile.folder[infoFile.folderCount] = malloc(len);
+        infoFile.folder[infoFile.folderCount] = malloc(strlen(folder) + 1);
+
         if (infoFile.folder[infoFile.folderCount] == NULL)
           break;
-        strcpy(infoFile.folder[infoFile.folderCount++], folder);
+          
+        strcpy(infoFile.folder[infoFile.folderCount], folder);
+
+        if(longFolder != NULL)
```
digant73, line 204: "if (", not "if("

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -172,11 +194,24 @@ bool scanPrintFilesGcodeFs(void)
...
+        {
+          infoFile.longFolder[infoFile.folderCount] = malloc(strlen(longFolder) + 1);
+          if (infoFile.longFolder[infoFile.folderCount] != NULL)
+            strcpy(infoFile.longFolder[infoFile.folderCount], longFolder);
+        }
+        else
+          infoFile.longFolder[infoFile.folderCount] = NULL;
+
+        infoFile.folderCount++;
+
```
digant73, line 214: remove this empty line

TFT/src/User/Configuration.h
```diff
@@ -998,13 +998,13 @@
  * Generic Debug
  * Uncomment/Enable to enable arbitrary debug serial communication to SERIAL_DEBUG_PORT defined in board specific Pin_xx.h file.
  */
-//#define DEBUG_SERIAL_GENERIC  // Default: commented (disabled)
+// #define DEBUG_SERIAL_GENERIC  // Default: commented (disabled)
```
digant73, line 1001: I suspect you committed this file by mistake

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -188,13 +195,17 @@ bool scanPrintFilesGcodeFs(void)
...
-        infoFile.longFolder[infoFile.folderCount] = malloc(strlen(longFolder) + 1);
+        if (longFolder != NULL)  // avoid allocating space if there's no long folder name
+          infoFile.longFolder[infoFile.folderCount] = malloc(strlen(longFolder) + 1);
 
         if (infoFile.folder[infoFile.folderCount] == NULL)
           break;
           
         strcpy(infoFile.folder[infoFile.folderCount], folder);
-        strcpy(infoFile.longFolder[infoFile.folderCount], longFolder);
+        if(longFolder != NULL)
```
digant73, line 204: if (

TFT/src/User/API/Gcode/mygcodefs.c
kisslorand: Will be fixed in an upcoming bugfix.

TFT/src/User/API/Gcode/mygcodefs.c
kisslorand: Will be fixed in an upcoming bugfix.

TFT/src/User/Configuration.h
kisslorand: Yes, by mistake. Will be fixed in an upcoming bugfix.

TFT/src/User/API/Gcode/mygcodefs.c
kisslorand: It will be fixed in an upcoming bugfix.

Guilouz:
I see you are in onbardSD fixes, do you think it's possible to implement MULTI_VOLUME support to have the choice to mount SDCard or USB drive with compatible motherboards ?

Like MKS firmware : https://www.facebook.com/watch/?v=1576346382757020

kisslorand:
Stage 3.2 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/865f7575a8cca73de2386499573e09953dba40c4) fixes a bug in the file list when selecting files from onboard SD. Some files, even thou they had long file names, the short file name was displayed instead.

Before the fix:
NoLongFileName

After the fix:
WithLongFileName

kisslorand:
> I see you are in onbardSD fixes, do you think it's possible to implement MULTI_VOLUME support to have the choice to mount SDCard or USB drive with compatible motherboards ?
> 
> Like MKS firmware : https://www.facebook.com/watch/?v=1576346382757020

There's one more bug to fix and than I will look into it.

J0hanss0n:
Just a short general question because it applies to this PR from my point of view:
Wouldn't it be much more convenient to have one PR per (unrelated) Issue/Feature/Bug instead of putting it all together? This could help maintainers merge urgent/obvious fixes without being blocked by larger changes and corresponding discussions. 
Regards, Johannes

kisslorand:
Yes and no. Some bugfixes need the previous ones. Just take a look for how long this PR is opened. Now multiply that time with the number of fixes and you can estimate how long it would take to have them all merged into master.
What you are saying has a perfect reasoning, however it has been a common practice for everyone  to have only one open PR.

kisslorand:
Stage 3.3 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/8f68e75b72ee28dab81cd1c5cc0c6f56395ff61e) fixes a bug in the file list when selecting files from onboard SD. A file could show up in a folder where it didn't belong to.

Example:
Folder structure: 
Folder

Before the fix:
FileList_nOK

After the fix:
FileList_OK

TFT/src/User/API/Gcode/mygcodefs.c
```diff
@@ -71,16 +71,22 @@ bool scanPrintFilesGcodeFs(void)
       line++;            // (use a common file path) e.g. "/sub_dir/cap.gcode" -> "sub_dir/cap.gcode"
 
     // "line" never has "/" at the beginning of a path (e.g. "sub_dir/cap.gcode") while "infoFile.title" has it
-    // (e.g. "bSD:/sub_dir"), so we skip it during the check of current directory match (index 5 used instead of 4)
-    if (strlen(infoFile.title) > 5 && strstr(line, infoFile.title + 5) == NULL)  // if "line" doesn't include current directory
-      continue;
+    // (e.g. "bSD:/sub_dir"), so we skip it during the check of current folder match (index 5 used instead of 4)
+    Pstr_tmp = strstr(line, infoFile.title + 5);  // "+ 5" skips the 5 bytes related to prefix "bSD:/" in infoFile.title
```
digant73, line 75: you are accessing infoFile.title + 5 before checking the length for infoFile.title!

kisslorand: Thank you. Fixed here: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/0d5297c5411ec741f5b842bc800584e8cdbe78bf

kisslorand:
> I see you are in onbardSD fixes, do you think it's possible to implement MULTI_VOLUME support to have the choice to mount SDCard or USB drive with compatible motherboards ?
> 
> Like MKS firmware : https://www.facebook.com/watch/?v=1576346382757020

Check https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/2215#issuecomment-1014884823
In the video that you posted the link for it can be seen that the TFT is connected in Marlin mode only.

kisslorand:
Stage 3.4 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/13f4e7ba57b881fbedb4e8b46d822699619884e0) partially eliminates sending multiple commands at once. On some boards Marlin expects only one command at a time. It is against what ADVANCED_OK stands for but that's the way it is Marlin at this point.
This commit most probably fixes issue #2295 (based on user's feedback).

After/if this PR is merged I will ask @digant73 to separate some commands that left unchanged. It's about MBL settings, it's his area.

digant73:
@kisslorand As far I see Marlin reads a stream of bytes (multiple commands separated by \n) putting each command in a queue before processing the queue

J0hanss0n:
@digant73 
Since the stream is received asynchronously to the queue loop, and I assume the loop running much faster than the bytes are received, Marlin starts processing as soon as the first newline is detected. That means, @kisslorand theory is still plausible, because the processing/answer can interfere with receiving the second command.

digant73:
@J0hanss0n Marlin loops on reading the stream until no more data are available. So in case of multiple commands on same stream, Marlin will reads all the commands (the command separator is \n or \r) and put them on the queue (having BUFSIZE entries).
Only when the stream was read, Marlin will start to process the commands in the queue (e.g. M200 and M221). See function get_serial_commands on file Marlin/src/gcode/queue.cpp

J0hanss0n:
@digant73 
> Marlin loops on reading the stream until no more data are available.

Yes but I can't see a intercharacter timeout being checked (or do I miss it?). And since the loop works faster than character times (about 100us @115200baud) there's a chance the first command is processed during rx of second command.

The c < 0 code you refer to, is just an exception catch for the case the serial implementation reports available data (serial_data_available) but then returns  -1 or other invalid data.

digant73:
the TFT sends "M220\nM221\n" as a unique stream. According to the issue, it seems that only M221 is then considered (sooner or later) by Marlin as M1, M21 etc... There are single commands that are longer than "M220\nM221\n" but it seems that they are not affected by any reading issue. So it seems to me that the issue is more in Marlin when it has a queue with more than 1 entry. The piece of code I reported is the one I would expect to handle a problem reading from the serial port (the character will be skipped so M221 could be read as M21 etc...).

J0hanss0n:
> the TFT sends "M220\nM221\n" as a unique stream.

There are no "unique streams" that are somehow atomic in serial communication. Each byte received is separated by the time it takes to transmit it.
As soon as you process the single bytes faster than they are sent you will break out the hadData loop, and this is likely the case.

digant73:
The only delay is on Marlin (when buffering each command from serial before moving to read the next one, if present) and it works on a slower mainboard as mine (so no problem also on skr 2). The scenario you reported would affect any command sent by TFT and that is not the case

J0hanss0n:
There's a delay between each byte received because it is transmitted bitwise in the configured baudrate. The faster the mainboard the more likely it is that the bytes are processed faster than sent. So it makes sense that slower boards are less prone to the error. 
Single commands never do harm because during processing and answering by the mainboard no new bytes will be transmitted from TFT which can get lost.
If there's really some sort of half duplex implementation because of dma use for serial communication present, all of @kisslorand observations make sense.

V1EngineeringInc:
This simple change has been tested in the real world as well on many combinations and has a dramatic effect on usability. I highly recommend merging this. If anything, why not? The other option is to figure out why we can't use these TFT's, but no other screens are having the issue on Marlin. 

Even if it was a Marlin issue, it needs to get figured out and fixed (PR request) from this side because it only effects these TFT's. The beauty of open source is adding to it, not telling them to change things your use case breaks.

I am here to test because my understanding of coding is not as in depth as you all are, I have also been working with and contributing to Marlin for a considerable amount of time.

oldman4U:
Not to forget that it seems that only mainboards with a certain cpu are affected. We had this before with other commands

J0hanss0n:
> Marlin is enqueuing both M220 and M221 received as "M220\nM221\n".

No it doesn't. Because there's a break between each character. The stream is not atomic. The loop is left after each received byte because the loop processing and command enqueuing in case of EOL is faster than reception of a new byte. That means M220 gets enqueued, and since the first byte of M221 is not yet there, it gets processed outside the rx loop. Then the answer of M220 interferes according to the half duplex implementation of serial with the reception of the M221, leading to lost bytes.
Real world observations confirm this.

digant73:
@kisslorand if you have more than two serial ports enabled in Marlin, simply enable only the TFT port and it will probably solve the issue

kisslorand:
I have only two serials enabled, USB and TFT.

digant73:
The problem should be on the number of serial ports

kisslorand:
I think it is more than obvious why neither of them cannot be disabled.
Without USB Marlin doesn't even compile, without TFT... :)

kisslorand:
I could disable USB serial. Checking it right now.

digant73:
that should be better.

digant73:
> I could disable USB serial. Checking it right now.
> 
> So basically what you want is to stay on a serial port until there's no more data on that serial and then check the other serial(s), right?

yes, but other changes should be applied to GCodeQueue::get_serial_commands(), so at the moment it is more easy to enable only one port. Probably the problem is on the USB port and that affects the TFT

kisslorand:
> Which is the port value for USB? 1?

SERIAL_PORT 1
SERIAL_PORT_2 -1 - this is how my config was

I just switched so 
SERIAL_PORT -1    (this is for USB)
SERIAL_PORT_2  1   (this is for the TFT)

Until now it was the other way around.

digant73:
ok, the second solution should be ok. GCodeQueue::get_serial_commands() should acces only port 0 (TFT).
With your first configuration, GCodeQueue::get_serial_commands() tried to access port 1 that is not existing in your mainboard and that could be the reason of the issue

digant73:
> > Which is the port value for USB? 1?
> 
> SERIAL_PORT 1 SERIAL_PORT_2 -1 - this is how my config was
> 
> I just switched so SERIAL_PORT -1 (this is for USB) SERIAL_PORT_2 1 (this is for the TFT)
> 
> Until now it was the other way around.
> 
> Nah, it doesn't change anything, after 10 seconds I got this:
> 
> ```
> Marlin >> wait
> TFT    >> M220
> TFT    >> M221
> Marlin >> X:0.000 Y:0.000 Z:0.000 E:0.000 Count X:0 Y:0 Z:0
> Marlin >> FR:100%
> Marlin >> ok P15 B15
> TFT    >> M118 P2 This is a very long message to test some things.
> Marlin >> echo:E0 Flow: 100%
> Marlin >> ok P15 B15
> Marlin >> This is a very long message to test some gs.
> ```

that is not the scenario for M21 in Marlin side. This is a scenario than can happen on TFT side. So it is not a problem. Try to see if you will get the issue from Marlin side (so you will see a print stop or M21)

kisslorand:
> > ```
> > Marlin >> wait
> > TFT    >> M220
> > TFT    >> M221
> > Marlin >> X:0.000 Y:0.000 Z:0.000 E:0.000 Count X:0 Y:0 Z:0
> > Marlin >> FR:100%
> > Marlin >> ok P15 B15
> > TFT    >> M118 P2 This is a very long message to test some things.
> > Marlin >> echo:E0 Flow: 100%
> > Marlin >> ok P15 B15
> > Marlin >> This is a very long message to test some gs.
> > ```
> 
> that is not the scenario for M21 in Marlin side. This is a scenario than can happen on TFT side

You forgot the I use HW monitoring, not SW monitoring.
"TFT      >>" is the communication wire TX from the TFT
"Marlin >>" is the communication wire TX from Marlin

digant73:
I know but this is a scenario that can happen, while you should not see on Marlin side any print stop or any M21 ack towards the TFT

kisslorand:
The testing I am doing now are with `speedQueryWait = (storeCmd("M220\nM221\nM118 P2 This is a very long message to test some things.\n"));`, not with separated commands.

digant73:
yes and that is ok. You can also restore the original "M220\nM221\n" and see if the issue (print stop with M1 or M21 message) will be present or not. If not the issue was due to the use in Marlin of a wrong port (1)

kisslorand:
Switching serial ports TFT<->USB:
```
Marlin >> wait
TFT    >> M220
TFT    >> M221
Marlin >> FR:100%
Marlin >> ok P15 B15
Marlin >> echo:No SD card
Marlin >> ok P15 B15
```

digant73:
see also the changes needed in Marlin. The swap of ports does not change anything. You simply need to replace a line of code in Marlin to scan only port 1 (and not also 0)

kisslorand:
Stage 3.5 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/6fd1853d9f88213a35cf8bd7d02742faddebf75f) is a global patch to avoid sending multiple commands to the command queue.
The amount of free buffer is not known to the TFT, it can lead to lost commands if more command is sent than the Marlin's can take,
Later if/when ADVANCED_OK will be implemented in the TFT, the command queue handling procedure could take care of sending multiple commands at once.

digant73:
there is no need to change storeCmd. If you want to split a multicommand you can use mustStoreScript as already done in the code or split the commands directly were needed.

kisslorand:
> there is no need to change storeCmd. 

Yes, there is, explained under:

>If you want to split a multicommand you can use mustStoreScript 

That's exactly what I did respecting the non-urgent statue of the `storCmd()`contrary to the urgent/must atribute of `mustStoreScript()`

> or split the commands directly were needed

I explained in comments why it is not desired. Some command groups must remain together and get a feedback for them being sent together. What's the feedback for the whole group if only one is not sent? False? True? What if only one is sent and the rest not? True? False? Will you send all of them again? None of them?
The new logic will keep them separated and send them all or none. In case of none a retry of sending all will be attempted.

digant73:
you can control that at the source (splitting the commands or not). You simply complicated a function making it longer and slower. Remember that it is used also during the print where each line is made by one gcode

kisslorand:
This is complicated? I am sorry you feel this way.
So you're suggesting to complicate it at each place where multiple commands need to be sent. Check at each place the availability of place in the command queue? What about the feedback? When is it true? When is it false?
So you're suggesting implementing everything in each place will make it faster and less complicated? Hmmm
It is open source, you can make anytime a "real, stable, perfect, logical, optimized, simple, fast, easy" PR.

digant73:
There is only one multiple command in the whole code (M220\nM221\n) + one for MBL setting (for which no split is needed). There are thousands of commands in the code already split and sent one after another. There is no need to change a function that is doing what it is requested to do. Furthermore, you consider only Marlin as printer fw.

kisslorand:
Why would you send multiple commands if you do not know that there's enough room or not in the motherboard's buffer?

kisslorand:
Prebuilt binaries with this PR are ready and can be downloaded here: https://mega.nz/folder/Hx4mXKjY#WFvcTZWweq_VRW9pR7nr_g
They are aligned to master as of today 21.1..2022

digant73:
@kisslorand we all know that you like to simply create useless discussions. I was referring to all the gcodes sent with sendCmd function in the code. With the exception of M220\nM221\n and 1 MBL setting, all the gcodes are already invoked separately, so buffered separately and so sent separately by TFT. That new function is simply totally useless, a degradation of space and performance. But we know you always find a not compliancy here and there. Marlin has an incoming queue shared with all the serial ports enabled so you cannot even be sure the gcode sent by TFT is not discarded on mainboard due to its queue reaching its limit

kisslorand:
Stage 3.6 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/f13c45420e4799f0e7e67cd352d63c7412605e00) eliminates the usage of partial heat related commands in the command queue and reduces code.

kisslorand:
Stage 3.7 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/e0bf90b8f15e8dcaf11ba67ea87274369bbbfaaa) speeds up the separation (if needed) of the commands before they're sent to the command queue.

kisslorand:
Stage 3.8 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/683a003ecdb8af47bc51d135d26eadeda9540e65) speeds up the evaluation of messages received from the motherboard.. Code size is reduced.
Stage 3.8.1 (commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2291/commits/d350708eb9855491952a8c58201d50b9703ba651) speeds up parsing "config.ini" file and speeds up evaluating the commands to be sent to the motherboard. Code size is reduced.

Prebuilt binaries here: https://mega.nz/folder/Hx4mXKjY#WFvcTZWweq_VRW9pR7nr_g

bigtreetech:
Hey  Buddy. There are still some things I don't understand in this PR, so I decided to divide this PR into several different PR according to different repair features. I will merge the PR I confirmed to be OK first(For example #2348). When I divide all the features, I will discuss with you again what I don't understand. And thank you for your hard work

PR #2325, opened 2022-01-17
Fix duplicate serial data pushed to debug serial port

guruathwal:
- Fix a bug where serial data received from the printer was being pushed to debug serial port twice when any of the debug options are enabled in the **configuration.h** file.

### DIFF
--- a/TFT/src/User/API/parseACK.c
+++ b/TFT/src/User/API/parseACK.c
@@ -1242,13 +1242,18 @@ void parseACK(void)
       ack_port_index = PORT_1;  // reset ACK port index to avoid wrong relaying (in case no more commands will
                                 // be sent by interfaceCmd) of any successive spontaneous ACK message
     }
-    #ifdef SERIAL_PORT_2
+    #if defined(SERIAL_PORT_2) || defined(SERIAL_PORT_3) || defined(SERIAL_PORT_4)
       else if (!ack_seen("ok") || ack_seen("T:") || ack_seen("T0:"))  // if a spontaneous ACK message
       {
         // pass on the spontaneous ACK message to all the supplementary serial ports (since these messages come unrequested)
         for (SERIAL_PORT_INDEX i = PORT_2; i < SERIAL_PORT_COUNT; i++)
         {
-          if (infoSettings.serial_port[i] > 0)  // if serial port is enabled
+    // forward data ony if serial port is enabled
+          if (infoSettings.serial_port[i] > 0
+              #ifdef SERIAL_DEBUG_PORT
+                && serialPort[i].port != SERIAL_DEBUG_PORT // do not forward data to serial debug port
+              #endif
+              )
           {
             Serial_Puts(serialPort[i].port, dmaL2Cache);  // pass on the ACK message to the port
           }
@@ -1263,13 +1268,18 @@ void parseACK(void)
   }
 }
 
-#ifdef SERIAL_PORT_2
+#if defined(SERIAL_PORT_2) || defined(SERIAL_PORT_3) || defined(SERIAL_PORT_4)
 
 void parseRcvGcode(void)
 {
   for (SERIAL_PORT_INDEX i = PORT_2; i < SERIAL_PORT_COUNT; i++)  // scan all the supplementary serial ports
   {
-    if (infoSettings.serial_port[i] > 0)  // if serial port is enabled
+    // forward data ony if serial port is enabled
+    if (infoSettings.serial_port[i] > 0
+        #ifdef SERIAL_DEBUG_PORT
+          && serialPort[i].port != SERIAL_DEBUG_PORT // do not forward data to serial debug port
+        #endif
+        )
     {
       while (syncL2CacheFromL1(serialPort[i].port))  // if some data are retrieved from L1 to L2 cache
       {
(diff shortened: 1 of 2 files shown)


TFT/src/User/API/parseACK.c
```diff
@@ -1242,13 +1242,18 @@ void parseACK(void)
...
     }
-    #ifdef SERIAL_PORT_2
+    #if defined(SERIAL_PORT_2) || defined(SERIAL_PORT_3) || defined(SERIAL_PORT_4)
       else if (!ack_seen("ok") || ack_seen("T:") || ack_seen("T0:"))  // if a spontaneous ACK message
       {
         // pass on the spontaneous ACK message to all the supplementary serial ports (since these messages come unrequested)
         for (SERIAL_PORT_INDEX i = PORT_2; i < SERIAL_PORT_COUNT; i++)
         {
-          if (infoSettings.serial_port[i] > 0)  // if serial port is enabled
+    // forward data ony if serial port is enabled
```
digant73, line 1251: is that comment not indented properly?

TFT/src/User/API/parseACK.c
```diff
@@ -1242,13 +1242,18 @@ void parseACK(void)
...
+          if (infoSettings.serial_port[i] > 0
+              #ifdef SERIAL_DEBUG_PORT
+                && serialPort[i].port != SERIAL_DEBUG_PORT // do not forward data to serial debug port
```
digant73, line 1254: usual two spaces for comment?

TFT/src/User/API/parseACK.c
```diff
@@ -1263,13 +1268,18 @@ void parseACK(void)
...
 
 void parseRcvGcode(void)
 {
   for (SERIAL_PORT_INDEX i = PORT_2; i < SERIAL_PORT_COUNT; i++)  // scan all the supplementary serial ports
   {
-    if (infoSettings.serial_port[i] > 0)  // if serial port is enabled
+    // forward data ony if serial port is enabled
+    if (infoSettings.serial_port[i] > 0
+        #ifdef SERIAL_DEBUG_PORT
+          && serialPort[i].port != SERIAL_DEBUG_PORT // do not forward data to serial debug port
```
digant73, line 1280: usual two spaces for comment?

digant73:
@Gurmeet Something not related to this PR that could be fixed in Terminal menu.
When you enter in the Terminal menu the history (up to 10 pages) of input/output is collected but when you send a gcode and the Terminal menu is open, only the current page and the following will be available. It means that part of the output sent by mainboard is lost. For example, if I send M43 (PINS_DEBUGGING enabled in Configuration_adv.h) only the last part of the output is displayed. It should be better to improve the terminal starting from the current page but allowing also to browse the previous pages.

TFT/src/User/API/parseACK.c
digant73, line 1251: "only" instead of "ony"

TFT/src/User/API/parseACK.c
```diff
@@ -1263,13 +1268,18 @@ void parseACK(void)
...
 
-#ifdef SERIAL_PORT_2
+#if defined(SERIAL_PORT_2) || defined(SERIAL_PORT_3) || defined(SERIAL_PORT_4)
 
 void parseRcvGcode(void)
 {
   for (SERIAL_PORT_INDEX i = PORT_2; i < SERIAL_PORT_COUNT; i++)  // scan all the supplementary serial ports
   {
-    if (infoSettings.serial_port[i] > 0)  // if serial port is enabled
+    // forward data ony if serial port is enabled
```
digant73, line 1277: "only" instead of "ony"

guruathwal:
> @gurmeet Something not related to this PR that could be fixed in Terminal menu. When you enter in the Terminal menu the history (up to 10 pages) of input/output is collected (in background) but when you send a gcode and the Terminal menu is open, only the current page and the next ones will be available. It means that part of the output sent by mainboard could be lost (making the use of the terminal useless). For example, if I send M43 (PINS_DEBUGGING enabled in Configuration_adv.h) (providing 7-8 pages of output) only the last part of the output is displayed. It should be better to improve the terminal starting from the current page but allowing also to browse the previous pages.


@digant73 The current terminal system is designed to allocate memory to pages according to the space needed by each page. this gives more flexibility to have more pages for the same memory size against pre-allocating memory to each page because most of the time one page is drawn on 1/3 of the screen. But because the text from the printer is heavily padded with spaces for M43, one page will be drawn on the whole screen resulting in fewer pages displayed. Even if we change it to not reset after switching to the terminal it will still end up displaying the last parts of the text.

digant73:
@Gurmeet Ok but is should be better to enable the history handling just before the gcode is sent and not when menuTerminalWindow is loaded. In that case the entire history will be available even if menuTerminalWindows is loaded when part of the output from mainboard was already received (e.g. 5/8 pages for M43). If you have a fast mainboard and/or slow TFT, it is currently quite easy to loose part of the output.
It seems that in menuKeyboardView we could simply avoid to execute loopBackEnd() in case a gcose is sent. In that case, menuTerminalWindow will be loaded first (in menuTerminal) and will start to handle the history properly

rondlh:
I also found some issues with the console in relation to the M43 command, resulting in truncated output and all data turning orange.
If padding with spaces is the problem then I would suggest the console to filter the data to replace 4 or more repeated spaces with 3 spaces. This way most formatting will be fine, but heavily padded data, that doesn't fit in 1 row anyway, would not take too much space.

PR #2341, opened 2022-01-22
Fix Neopixel LED light

Guilouz:
- Fixed LED status after exiting print summary.
- Fixed state of the LED after exiting the PID menu.

### Benefits

LED are no longer off after exiting print summary (LED change from green to white) and LED are no longer off after exiting the PID menu.


digant73:
I was planning to make the RGBWPI components configurable in Configuration.h in a further PR fixing some bugs.

PR #2361, opened 2022-01-28
Idea for custom lib for marlin for touchmode over hardwareserial/lcdserial 

lnfactionfreddy:
an idea for a library for Marlin for better support in Touch mode. Also on heating wait. Can also be made compatible with Marlin mode.


lnfactionfreddy:
Yes, I can describe it more.
I only use the definition of Lcd_serial but it has nothing to do with the DWIN display.
Basically emulate one of the default serial port from marlin like the MKS TFT V1.0 for esp8266 module (wifi module).
So you can send via the lcd_serial gcode that will be passed on to the ring buffer and also send custom command (text).
Without Marlin being in some kind of wait loop and not answering.

TFT => send_gcode => tft_serial => filter whether gcode => send to ringbuffer => reply send to tft =>tft_serial=>TFT 

TFT => send_custom => tft_serial => filter whether gcode => executes custom function => reply send to tft =>tft_serial=>TFT 

marlin query for event => reply send to tft => tft_serial=> TFT

Unfortunately, the responses from the standard serial ports cannot be redirected to the lcd_serial, so you would have to write an extra response for this.
Such as for M115 M503 M502 M500 M92 M114....

advantage if the Marlin is waiting (heating up and.. ) the function btt_tft_api::update();  still working and can also send and receive.

Sorry if my english is so bad.
Please let us know if you have any further questions.

radek8:
As for the fact that Marlin does not respond while waiting for M109 and M190, all you have to do is turn on M109 and M190 emulation on the TFT.
Then there is no problem that Marlin does not respond while waiting for the temperature to reach.

lnfactionfreddy:
what about the remoteprint ui with the stop and pause buttom.

lnfactionfreddy:
Okey

radek8:
I'm not saying it's bad. I'm just curious :-)
You don't have to close your PR because of my question.
I'm not a programmer so I can evaluate your work.
I'll be happy to test it when you're done.

lnfactionfreddy:
In principle it already works for M105 and M503 (as a dummy) 
 The effort is in principle huge for programming for the whole answer of the Gcodes
 The Remoteprint (Octopi) was also triggered via it, as with the DWIN display.
 and also use the buttom stop and pause in the remote print.
That was just my idea behind it.
So it's only about the function to test for G Commands 
Temp report and Homing work.

lnfactionfreddy:
You could also think about integrating the Marlin mode into the library.
I would have suggested it to Marlin, but they had no interest in it.

PR #2365, opened 2022-01-30
Avoid sending multiple commands at once

kisslorand:
### Requirements

BTT or MKS TFT.

### Description

   This PR eliminates the sending of multiple commands at once. At the moment of sending commands from the queue it is not known if there are more than one slot free in the mainboard's buffer so sending multiple ones at once can lead to missed/not received commands.
   A new procedure was implemented that deals with these command groups that need to be sent together. It puts them into the command queue one by one so they will be sent one by one. If there is no room for all of them none is sent. There's a feedback of success if all of them were queued or a feedback of failure if there was an error, in which case none of them is sent. With the existing procedure it was impossible to send all of them or none of them if there was no room in the command queue. 
   A lot of approaches were tested to find the right balance between speed, code size and memory usage. There were techniques that gained fraction of nanoseconds but used more code or memory and there were others with less code but three times slower and so on.

### Related issues

Sending commands separately could avoid the bug discussed in issue #2295.

### Notes

   For those concerned about speed here are some numbers on a BTT TFT35 v3:
 Sending "M220\nM221\n" with the new procedure (5.534 microseconds) is 3.32% slower than sending separated "M220\n" and "M221\n" with the old procedure (5.356 microseconds). This means 0.178 microseconds more to put "M220\nM221\n" into the command send buffer. Just as an observation, it takes around 400 microseconds to send "M220\nM221\n" to the mainboard on UART at 250000 b/s.

### DIFF
--- a/TFT/src/User/API/MachineParameters.c
+++ b/TFT/src/User/API/MachineParameters.c
@@ -406,7 +406,7 @@ void sendParameterCmd(PARAMETER_NAME name, uint8_t elementIndex, float Value)
 {
   char tempCmd[30];
   sprintf(tempCmd, "%s %s", parameterCode[name], parameterCmd[name][elementIndex]);
-  storeCmd(tempCmd, Value);
+  storeCmdScript(tempCmd, Value);  // storeCmdScript() used because there are elements in parameterCmd[] with extra command
 }
 
 void saveEepromSettings(void)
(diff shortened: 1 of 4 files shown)


V1EngineeringInc:
My extruder is not showing up with a temp, even though it does work?

V1EngineeringInc:
Holding up great for me more than 24 hours on two printers with SKR Pro's and TFT35 e3 v3. Makes the TFT side useable again. It would be great to see this merged.

kisslorand:
Commit https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2365/commits/bc755d8a73fa20e2258e122280f95b878139f665 is wrong. It defeats the purpose of the function to send all script commands or none and give back a relevant feedback. Please do not merge it.

PR #2367, opened 2022-01-30
Disable Home Offset for Delta printers

Guilouz:
Disabled Home Offset icon, M206 command isn’t available for Delta printers.

### DIFF
--- a/TFT/src/User/Menu/Tuning.c
+++ b/TFT/src/User/Menu/Tuning.c
@@ -42,9 +42,9 @@ void menuTuning(void)
 
       case KEY_ICON_2:
         #if DELTA_PROBE_TYPE == 0  // if not Delta printer
-        storeCmd("M206\n");
-        zOffsetSetMenu(false);  // use Home Offset menu
-        OPEN_MENU(menuZOffset);
+          storeCmd("M206\n");
+          zOffsetSetMenu(false);  // use Home Offset menu
+          OPEN_MENU(menuZOffset);
         #endif
         break;
 
--- a/TFT/src/User/Menu/Tuning.c
+++ b/TFT/src/User/Menu/Tuning.c
@@ -9,9 +9,9 @@ const MENUITEMS TuningItems = {
     {ICON_PID,                     LABEL_PID},
     {ICON_TUNE_EXTRUDER,           LABEL_TUNE_EXTRUDER},
     #if DELTA_PROBE_TYPE == 0  // if not Delta printer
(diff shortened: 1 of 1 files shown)


TFT/src/User/Menu/Tuning.c
```diff
@@ -8,7 +8,11 @@ const MENUITEMS TuningItems = {
   {
     {ICON_PID,                     LABEL_PID},
     {ICON_TUNE_EXTRUDER,           LABEL_TUNE_EXTRUDER},
+    #if DELTA_PROBE_TYPE == 0  // if not Delta printer
     {ICON_PROBE_OFFSET,            LABEL_H_OFFSET},
```
digant73, line 12: indent all the lines between #if statement by two characters

TFT/src/User/Menu/Tuning.c
```diff
@@ -8,7 +8,11 @@ const MENUITEMS TuningItems = {
...
+    #else
+    {ICON_NULL,                    LABEL_NULL},
```
digant73, line 14: as previous comment

TFT/src/User/Menu/Tuning.c
```diff
@@ -37,9 +41,11 @@ void menuTuning(void)
         break;
 
       case KEY_ICON_2:
+        #if DELTA_PROBE_TYPE == 0  // if not Delta printer
         storeCmd("M206\n");
```
digant73, line 45: as previous comment

Foxies-CSTL:
Hi,
I also use Deltas printers with a BTT-TFT35 screen. The remark of @Guilouz is right (I also work on it) but you should not disable the "H_OFFSET" icon in the machine tuning part but rather reassign it to "P_OFFSET" with a warning message. I made a mistake when I wanted to make a merge on Master instead of a PR, sorry.
Here is the example that works for me.
27661dcab83c59968fabd8707f00a38c44e5bb2c

Guilouz:
@Foxies-CSTL P_OFFSET is already available in ABL menu, it's not necessary to have it twice.

Foxies-CSTL:
@Guilouz 
I know that but I found it easier to group them in the "tuning" part than to go through several menus. I do the same with the different UBL modes but I did not publish it.

Guilouz:
But the P_Offset has no place in the Tuning menu, it's a leveling adjustment and there's no point in creating duplicates.
We don't do PID and extruder adjustment all the time, it has more place in the ABL part next to calibration and leveling.

PR #2380, opened 2022-02-05
Add support for updating icons, fonts, config and language files from USB flash drive. 

guruathwal:
- Add support for updating icons, fonts, config and Language files from USB flash drive.
  - For devices with USB Flash drive support, the icons, fonts, config and Language files can be updated/flashed from a USB flash dirve as it was done through an SD card.
  - This reduces the need for removing/inserting SD card when only icons, fonts, config or Language needs updating.
  - If both SD card and USB drive are present then both of them will be scanned for update files. 
  - The firmware still needs to be updated from SD card.
- Reduce code for comment pasring.


kisslorand:
Just an idea...
Copy at least "config.CUR" and "language.CUR" to the SD card if they are updated from USB.
USB drives come and go, so if config needs some changes have it at hand on the SD card and avoid using an old one.

README.md
```diff
@@ -415,29 +403,27 @@ See [BIGTREETECH-TouchScreenFirmware/releases](https://github.com/bigtreetech/BI
...
+**cancel:**<br>
+`M118 A1 P0 action:cancel`
+
+**pause:**<br>
+`M118 A1 P0 action:pause`
+
+**resume:**<br>
+`M118 A1 P0 action:resume`
+
+**time remaining progress:**<br>
```
digant73, line 421: In the current readme file there were some missing info (due to characters requiring escape char "\\" first) in these last two commands.
The right syntax for the commands is:

**time remaining progress:**
M118 A1 P0 action:notification Time Left \<XX\>h\<YY\>m\<ZZ\>s (e.g. 02h04m06s)
or
M117 Time Left \<XX\>h\<YY\>m\<ZZ\>s

**file data progress:**
M118 A1 P0 action:notification Data Left \<XXXX\>/\<YYYY\> (e.g. 123/12345)
or
M117 Data Left \<XXXX\>/\<YYYY\>

guruathwal:
> Just an idea... Copy at least "config.CUR" and "language.CUR" to the SD card if they are updated from USB. USB drives come and go, so if config needs some changes have it at hand on the SD card and avoid using an old one.

I am still using USB drives older than 2012 but can't find any old SD cards.
IMHO, USB drives are far more durable when it comes to removing/inserting multiple times in an hour for cases like these.
Micro SD cards are fragile and are not designed for this purpose, I have already broken 3 16gb Micro SD cards and 5 Micro SD adapters only while trying to debug the firmware. who knows, maybe I am clumsy at handling Sd cards.

README.md
guruathwal: Oh. ok.

kisslorand:
You misunderstood me. I just suggested to have those files on rhe SD card also (with "CUR" extension). I keep the SD card in my printer all the time, do not use it for any other purposes. As for rhe USB sticks, I do not keep them in the printer and use them for various purposes and yes, I very often loose them. :)
  If I would need only one parameter to change it would be hard to remeber after 2 months which USB stick I used for FW flashing and even thou I would remeber, it might have been deleted a long time ago by the USB stick being used  for many other purposes.
So it would be easy to just grab tbe file from the SD card and change that.

README.md
digant73: thanks for the changes. There is just a missing \<br\> after M118 A1 P0 action:notification Data Left \<XXXX\>/\<YYYY\> (e.g. 123/12345). It is instead properly present for M118 A1 P0 action:notification Time Left \<XX\>h\<YY\>m\<ZZ\>s (e.g. 02h04m06s)

guruathwal:
@kisslorand  ok I understand it now.

README.md
guruathwal: ok

radek8:
My observation.
I have the configuration stored in the computer and when I need to change the parameter, I copy the configuration to the SD disk.
I'm using multiple SD drives and I won't search for the last configuration used. I always have the current one in the Computer

PR #2404, opened 2022-02-16
Cleanup, fixes, code reduction

kisslorand:
### Requirements

BTT or MKS TFT

### Description

Just some cleanup, fixes and code reduction.

PR #2406, opened 2022-02-18
Fixed wrong title in Printing menu if printing from remote

### LINKED ISSUE #2317: [BUG] Case light panel in touch mode is not working correct
### Description

Case light panel for touch mode working in not the best way. To accept changes it is necessary to press the return arrow sometimes and enter to the control menu for case light again. In order to turn off the light after turning on also you need to play a bit with opening the menu with control of the light case, go out, press twice to turn off/on icon. Sometimes is working sometimes not.

Is it possible to make a 1% step not a 10% step for changing the light intensity? Or enter a value like for Hot End or Bed?

For Marlin mode, everything is fine 1% change, when you accept the value let say 55% that value is simply set. 

### Steps to reproduce

1. Enter the case light menu 
2. try to turn on and off the light a few times and observe how it is working 
3. change % value and try to turn off and turn on light

**Expected behavior**
Turning off/on the light when the icon is pressed 
Setting the % value with 1% step instead of 10% step 
when % value is set light should fluently adjust the intensity according to the wanted value

**Actual behavior**
hard to control case light in touch mode
changing of the % is difficult 
turning off/on is difficult 

### Hardware Variant
TFT35 e3 v3.0.27 x DEC 30.2021 & Marlin 02000801
TFT35 e3 v3.0.27 x DEC 30.2021 & Marlin 02000801 & SKR 1.4 Turbo

digant73:
**IMPROVEMENTS:**
* Added **Software Repository** section in README.md: It highlights to use the **Master** branch for both the source code and resources (precompiled firmwares, themes etc...).
* Added **Portrait Mode** section in README.md: It explains how to compile and install the firmware with Portrait Mode support.
* Reviewed CaseLight menu and API: Implemented #2317. Reduced RAM and FLASH size.
* Some code reduction and cleanup.

**BUGFIXES:**
* Fixed wrong title in Printing menu if printing from remote: The bug can cause a freeze. Modified function getFilename (in vfs.h / .c) in order to provide the correct filename also if printing from remote (remote onboard or remote host).
* Fixed #2317.

resolves #2317

**PR STATE:** ready for merge


kisslorand:
I would like to suggest some minor changes.
Get "getFilename()" back as it was and in "initPrintSummary()" change to this:
`sprintf(infoPrintSummary.name, "%." STRINGIFY(SUMMARY_NAME_LEN) "s", (infoFile.source <= BOARD_SD) ? getFilename(infoFile.fileIndex) : infoFile.title);  // get short/long filename or filename from offboard print source`

"initPrintSummary()" is used only once in a print but "getFilename()" is/can be used more times.
These changes would not affect functionality, program size or speed but it would help avoiding coliision with PR #2404.
Thank you for your understanding.

digant73:
I would do that but there are also other collisions. As also explained to bigtreetech, current hideFilenameExtension and RestoreFilenameExtension functions apply the changes to both short and long filenames just to be consistent even with the case where only short filename is present (otherwise you could get a short filename with extension if long is present and one without extension if only short is present). For that reason there are also the helper (hidden) functions hideFilename and restoreFilaname (originally provided in one of your PR). The following functions (requiring an index) are used only in Print menu (the menu that really needs to use index to navigate in the list menu):

getFoldername
hideFilenameExtension (this is a sort of getFilename removing also extension)
restoreFilenameExtension (this is a sort of getFilename restoring also extension)

Once the print file is selected (by Print menu or by remote), the other menus (e.g. Printing menu) simply needs to retrieve the print filename. For that reason I provided getPrintFilename (no more requiring any index).

If getFilename() will be required in the future no problem to add it again. But currently it was improperly used only in initPrintSummary (just before hiding the extension) and Printing menu (to set the title without extension) and replaced by getPrintFilename.

kisslorand:
Thank you for the explanations, I understand your concern. This thing with filename handling got out of hand, it constantly needs a patch over a patched patch of a patch...

BTW, did you check my suggestion? If yes, what would be it's shortcoming?

Regarding the hiding of extension for both short and long file name, my opinion is that it is not needed because of the simple fact that if long file name is present than the short file name without extension is not needed anywhere in the code. Hiding the extension of the short file name in case the long file name is present has zero benefits.

digant73:
short filename is used also in Print.c. It is needed for the preview (and there was also a bug on that). Then we provided getFoldername and hide/restoreFilenameExtension returning the long name if present. But the values returned are used only as title to be displayed in the menus (so only estetic). I think that getPrintFilename is better to be used (it hides to Printing menu the checks to understand which is the info providing the filename). We defined getFoldername and hide/restoreFilenameExtension also for that reason (check and provide long name if present etc...). Of course those functions must be all consistent (return long name or short name, hide extension for long and short name etc...). You could also not consider the value returned by hide/restoreFilenameExtension and use the short names and that will work while a different result will be observed if the changes will be applied only to long names

README.md
```diff
@@ -80,15 +81,28 @@ Only the TFTs listed below are currently supported. Trying to install the firmwa
...
 Firmware source: https://github.com/MarlinFirmware/Marlin/releases
 
-Minimum Marlin firmware version: 2.0.8.1
+Minimum Marlin firmware version: **2.0.8.1**
 
-Distribution date: 2021-05-15
+Distribution date: **2021-05-15**
 
-In order the TFT firmware is able to provide all of its functionalities/features, the following options must be enabled in Marlin firmware.
+</br>In order the TFT firmware is able to provide all of its functionalities/features, the following options must be enabled in Marlin firmware.
```
kisslorand, line 105: "In order for the TFT's firmware to be able to provide..."

digant73: Ok thanks. Applied the change also to config.ini file

digant73:
Most of the time, rebase does not work. I need to close this PR and open it again

Issue #2398, opened 2022-02-15
Embedded Gcode thumbnails


trizmark:
I am using PrusaSlicer to generate my Gcode files and unable to get thumbnails to display on the BTT TFT35.

I have modified the PrusaSlicer configuration and enabled the G-code thumbnails and set it to 95x95.
I have changed Configuration.h and set THUMBNAIL_PARSER to 2. After compilation I can see that the RAM usage goes from 20028 bytes to 63940 bytes (which is in line with what's required for the Base64 decoding).

When I browse the files I can see a split second wait (I assume trying to process the embedded thumbnail), but no thumbnail is displayed.

Do I need to do anything else to get thumbnails working?

sarvenn:
Definately there is one way else and it's working perfect!
https://github.com/effgarces/Biqu-Thumbnail-Generator

I use the thumbnail parsel 0 (default-classic). I use this post processing script (it is now compatible with PS2.4 also) , it's flawless.

Download this ( https://github.com/effgarces/Biqu-Thumbnail-Generator/releases/download/0.03/biqu_convert_p24.zip ) and copy the standalone script to your prusaslicer.exe folder. Then add the script name to PS just like this:
![image](https://user-images.githubusercontent.com/66178082/154125276-0e24eb90-0326-4bcf-bd02-4a85e4efe2e8.png)

sarvenn:
News,
Hopefully it would be avaible in Superslicer without the need of script usage.
Merill is working on it.
https://github.com/supermerill/SuperSlicer/issues/887

trizmark:
Finally I had some time to look into what's going on with the firmware. After a bit of debugging, I realised that the only reason why the base64PNG thumbnail is not displayed is because the code is looking for it in the wrong place.

Let me elaborate:
First I set the THUMBNAIL_PARSER to 2 to ensure that we're looking for base64PNG thumbnails
After generating a gcode file with PrusaSlicer (with thumbnail size set to 95x95) I expected the firmware to do the following:
- look for base64PNG,
- look for RGB565 bitmap in comments
- look for RGB565 bitmap at fixed location

Looking at the debug logs, the order is exactly the opposite.
[...]

PR #2408, opened 2022-02-18
Fixed base64PNG thumbnail search

trizmark:
### Requirements

### Description

This is a fix for the embedded base64PNG thumbnail search.
Due to the nature of the thumbnail search order, base64PNG thumbnails were never displayed. While we try to locate an RGB565 thumbnail in the comments, the file pointer moves to the end of the MAX_THUMBNAIL_SEARCH_BLOCKS. The original code tried to locate the base64PNG thumbnail from this position onward, which meant that it was never found.

The fix is to simply seek back to the beginning of the gcode file at the start of the modelFileSeekToThumbnailBase64PNG function.

### Benefits

Base64PNG thumbnails (e.g. PrusaSlicer generated ones) are now displayed correctly.

### Related Issues

#2398

### DIFF
--- a/TFT/src/User/API/UI/ui_draw.c
+++ b/TFT/src/User/API/UI/ui_draw.c
@@ -280,6 +280,9 @@ uint32_t modelFileSeekToThumbnailBase64PNG(FIL *fp, uint16_t width, uint16_t hei
   snprintf(buf, sizeof(buf), "; thumbnail begin %hux%hu ", width, height);
   dbg_print("Start search\n");
 
+  // Seek to the beginning of the file as the file pointer was moved during the RGB565 thumbnail search
+  f_lseek(fp, 0);
+
   if (!modelFileFind(fp, buf))
     return 0;
 


digant73:
Please, don't forget to test also all the other scenarios (THUMBNAIL_PARSER set to Classic and RGB565)

trizmark:
@digant73 The changed code is only executed if THUMBNAIL_PARSER is set to PARSER_BASE64PNG:

The changed function is `modelFileSeekToThumbnailBase64PNG` in `ui_draw.c`, which is only called from `model_DirectDisplay_Base64PNG`, which is only called from `model_DirectDisplay`, where it's only called if THUMBNAIL_PARSER is set to PARSER_BASE64PNG:

```
#if (THUMBNAIL_PARSER == PARSER_BASE64PNG)
    if (model_DirectDisplay_Base64PNG(pos, &gcodeFile))
      return true;
#endif
```

I have tested the other two methods of THUMBNAIL_PARSER and they work fine (as they're called before the base64PNG parser).

PR #2410, opened 2022-02-18
No more partial cmd in queue, less code

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR eliminates partial or incomplete commands in the command queue and reduces code size.


TFT/src/User/API/interfaceCmd.c
```diff
@@ -867,15 +861,8 @@ void sendQueueCmd(void)
...
-            uint8_t i = cmd_seen('T') ? cmd_value() : heatGetCurrentHotend();
-            if (cmd_seen('S'))
-            {
-              heatSyncTargetTemp(i, cmd_value());
-              heatSetSendWaiting(i, false);
-            }
-          }
-          break;
+            synchNoWaitHeating(cmd_seen('T') ? cmd_value() : heatGetCurrentHotend());
+           break;
```
digant73, line 865: one char less of indent

TFT/src/User/API/parseACK.c
```diff
@@ -1078,7 +1078,9 @@ void parseACK(void)
         infoMachineSettings.autoReportTemp = ack_value();
 
         if (infoMachineSettings.autoReportTemp)
-          storeCmd("M155 ");
+        {
```
digant73, line 1081: it should be better to not use {} in an "if" with only one instruction (as done in all the rest of the file)

digant73:
According to your previous changes, the "else if" block at line 845 in interfaceCmd should be removed and so also the function updateCmd could be removed because no more used

kisslorand:
> According to your previous changes, the "else if" block at line 845 in interfaceCmd (M155 case) should be removed and so also the function updateCmd could be removed because no more used

There are no such things left behind.

TFT/src/User/API/interfaceCmd.c
kisslorand: Fixed.

TFT/src/User/API/parseACK.c
kisslorand: Fixed.

kisslorand:
@bigtreetech
Any reason this remains unmerged for so long?

kisslorand:
Closing because it's also implemented in PR #2519.

PR #2413, opened 2022-02-19
Add two settings for 'Show Bootscreen' and 'Alert that the Heaters are still on'

Diegovd:
### Description

Since:
- I personnaly do not need to see the BTT Bootscreen when powering on the printer, and there is a fixed 3 seconds waiting time added to the startup process because of the way it is programmed,
- I do not find the "Alert: the heaters are still on" message convenient
=> I decided to add two extra settings to control those behaviours.

Changes were implemented in:
- Configuration.h
- config.ini

An 'if' condition was added to:
- TFT/src/User/API/ModeSwitching.c
- TFT/src/User/Menu/common.c
in order to control the bahaviour.

Language files were updated accordingly.

The bevahivour can be thus set in the Configuration.h, the config.ini, or directly changed on Screen by the User (toggle button).

### Benefits
Allows Users that do not compile themselves the code to be able to change those settings by:
- Feature screen Menu (LCD)
- Config.ini File

I left the settings as the are presently programmed since the january-february releases:
- Bootscreen is Enabled (Default)
- Heaters Alert is Enabled (Default)

### Related Issues
No issues. Compilation, flashing, testing went OK.
This was tested successfully with a BTT TFT35 E3 V3 and a BTT SKR E3 Turbo Mainboard.

### Testing
(0.- - File name and settings for 'show bootscreen' / 'heaters alert' )

1.- - Configuration.h: 1 / 1 + config.ini: 1 / 1
=> assert after reboot, check in Features, change in Features, test heating with menus, test menus without heating, test after changing  'show bootscreen' / 'heaters alert' in Features and after reboot.

2.- - Configuration.h: 1 / 1 + config.ini: 0 / 0
=> assert after reboot, check in Features, change in Features, test heating with menus, test menus without heating, test after changing  'show bootscreen' / 'heaters alert' in Features and after reboot.

### Risks
Tests were not full coverage (not tested with Configuration.h: 0 / 0, since that is not the Default value).
Testing was done on only one TFT LCD device (mentioned above).

### Personal message
I am many things (software tester, Java programmer, Cypress programmer) but I am not a C++ programmer.
This is my first contribution to BTT.
Please let me know if the way I implemented the changes are correct or not.
Thank you for your patience.

Best regards, Diego von Deschwanden.


kisslorand:
I am sorry to burst your bubbles but both of the things you've changed are not taking into account that this FW goes eventually in a commercial product sold all over the world.
1. It's unlikely that Bigtreetech would like the user to be able to remove the logo on startup, it's advertising and I guess they are also proud to show who made the product in question. Just turn on a Samsung TV... Can you remove the startup logo from the menu as an option? :)
2. A reminder that heaters are on is a safety feature. In Europe (at least) it is mandatory for a product to have safety features to get approval of importing and selling. If a user removes those features himself by compiling his own firmware is one thing, but selling a product with the ability to let the user disable safety features is not the best idea.

Diegovd:
Hello @kisslorand , thank you for your feedback.

That is why I left the current configuration for both Items on 'On' in both Configuration.h and config.ini.

In the best-case scenario BTT can just acknowledge this PR as Feedback from a User (I have many many BTT products).
In the worst-case scenario, I can work with my forked version. So, no biggie :)

1.- About the Bootscreen: for me it helps to shave 3 seconds for the waiting on Ready Status, since the initial synchronization between LCD and mainboard is a bit long. (long is obviously very subjective ;) )

2.- About the Heaters alert: when preparing the machine and filament to be ready to print: why would I want to turn off the heaters ? I have never had to click Yes.
The cooldownTemperature() routine that calls that message is called when exiting: 
-- \TFT\src\User\Menu\BedLeveling.c
-- \TFT\src\User\Menu\LoadUnload.c
-- \TFT\src\User\Menu\Extrude.c

In my case these 3 Menus are used before printing.
Maybe LoadUnload.c can be useful at the end of a print, or when stopping a print. So in that case the message would make appropriate sense.
My End-Gcode takes care of unloading an enough lenght of filament that I only need to remove it manually from the extruder, without heating the nozzle again.

Obviously, if there is a regulatory policy or a company policy that would not allow this PR to continue, I will not be dissapointed. I fully respect the fact that any change has to be in the benefit for all persons involved.

Thanks
Best regards, Diego von Deschwanden.

Guilouz:
As much as having the logo doesn't matter to me, but actually this popup which asks if you want to stop the heating or not is rather annoying when you leave a menu. If I decided to heat it is indeed to launch a print.

digant73:
@Diegovd If needed by someone, that kind of settings (in particular bootscreen) can be made configurable at compile time (as made for many others) instead of wasting FLASH size. FLASH size is almost 85%. IMHO it is better to save space for adding/improving features

kisslorand:
> @Diegovd If needed by someone, that kind of settings (in particular bootscreen) can be configured at compile time

That one is already configurable at compile time.
Hint: "BTT_BOOTSCREEN_TIME"

digant73:
yes, I was referring to SHOW_BTT_BOOTSCREEN to disable bootscreen instead of make it configurable at runtime.
The heater message was provided in particular for the bed leveling menu. You can prefer to heat bed and nozzle during BL process, validation, probe offset etc... Leaving the BL menu it is asked to switch off. That code is shared just to reduce the FLASH size. So it is used also by other few menus

Diegovd:
Hello @kisslorand  @digant73, thank you for your comments.
Not many people compile the binaries themselves.
For you it might be very simple. Many users have no idea. I am in many 3D Support groups and I see that everyday.


1.- Can the Showbootscreen 'fixed 3 seconds loop' run parallel to the boot sequence?
  Otherwise the way that function is implemented is making people wait 3 seconds more on an already quite long boot and synchronization sequence.
------------------------------------------------------------------------------------------------------------------------
 TFT/src/User/API/ModeSwitching.c (original)

 29     #ifdef SHOW_BTT_BOOTSCREEN
        if (modeFreshBoot)
        {
          uint32_t startUpTime = OS_GetTimeMs();

          heatSetUpdateSeconds(TEMPERATURE_QUERY_FAST_SECONDS);
          LOGO_ReadDisplay();
          updateNextHeatCheckTime();  // send "M105" after a delay, because of mega2560 will be hanged when received data at startup

          **while (OS_GetTimeMs() - startUpTime < BTT_BOOTSCREEN_TIME)  // display logo BTT_BOOTSCREEN_TIME ms
          {
            loopProcess();
          }**

          heatSetUpdateSeconds(TEMPERATURE_QUERY_SLOW_SECONDS);
          modeFreshBoot = false;
        }
      #endif
      break;
      
      ----------------------------------------------------------------------------------------------------

2.- Can someone please explain to me what is the point of the Heaters alert please? Is there a regulatory requirement?
     That function is not user-friendly, and to me and @Guilouz do not make sense.

As I said previously, the PR keeps the actual behaviour and settings set by BTT.
It lets the User choose what is more convenient for him.


Thank you.
Best regards, Diego

kisslorand:
The heaters alert is a safety feature. User friendly or not, it is still what it is.
When you hit "Stop" during a print, there's also a not user friendly popup that asks you if you really want to stop the print. Should we remove that also or make it optional?
When you select a gcode file for printing there's also another non-user-friendly popup that shows you again the selected file and asks you if you want to print that file. Should we remove that also or make it optional?
When you want to extrude/load/unload filament there's again a unfriendly popup that alerts you if the nozzle temperature is low. Should that be removed or made optional too?
Etc, etc, etc...

I know you want to make a point and stand by your preferences, but it's not me or @digant who pull the strings around here. We expressed our opinions so did you. Someone else is calling the shots.
As for the regulatory requirement, I am not entering into the dispute who, why, when, how certificates are needed and obtained, certificates that without them you cannot import and sell goods.

Diegovd:
Hi @kisslorand , what about code review, did I do a good/clean job? Are the if/then statements OK on a C++ standard point of view? As I said, I am a complete newbie. But tried to do it as expected at this level.

Thank you very much.
Regards, Diego

traffic-light:
@Diegovd it's C. Not CPP ;-)
Didn't take a look. But I hope you used C functions ;-)

The "stop heating?" Question can be indeed anoying. It's not really a safety option be easy you should turn on the timeout function in marlin itself isteas of relying on an external input, like the screen.
I wouldn't make this configuratie available in the screen settings. Only in file would be enough.

Still. I don't think it's useful enough to consume memory.


I think the boodscreenpart isn't really anoying because the mainboard is never alive, at least for, within 3 seconds.

There is a good instruction in the readme for compiling.. And in my opinion it's needed to know how to compile software. For example, most users using the scree are also using a bigtreetecht mainboard. For this mainboard you should also compile marlin according to the machine....

digant73:
#2430 added SAFETY_ALERT property in Configuration.h just to enable/disable safety alert popup messages used by some menus (e.g. heaters still on when leaving some menus). That property could be used in the future for any other kind of message considered as a "safety alert".
If disabled (commented out), no extra code is also present in the binary.

PR #2420, opened 2022-02-22
Implemented #2317 + some code reduction

### LINKED ISSUE #2317: [BUG] Case light panel in touch mode is not working correct

digant73:
**IMPROVEMENTS:**
* Reviewed CaseLight menu and API: Implemented #2317. Reduced RAM and FLASH size.

resolves #2317

**PR STATE:** Ready for merge


TFT/src/User/API/parseACK.c
```diff
@@ -27,15 +27,17 @@ const ECHO knownEcho[] = {
...
+  {ECHO_NOTIFY_TOAST, "echo:EEPROM can"},         // M503
   {ECHO_NOTIFY_NONE, "Cap:"},                     // M115
   {ECHO_NOTIFY_NONE, "Config:"},                  // M360
   {ECHO_NOTIFY_TOAST, "Settings Stored"},         // M500
   {ECHO_NOTIFY_TOAST, "echo:Bed"},                // M420
   {ECHO_NOTIFY_TOAST, "echo:Fade"},               // M420
   {ECHO_NOTIFY_TOAST, "echo:Active Extruder"},    // Tool Change
   {ECHO_NOTIFY_NONE, "Unknown command: \"M150"},  // M150
+  {ECHO_NOTIFY_TOAST, "echo:No SD card"},         // onboard SD card
+  {ECHO_NOTIFY_TOAST, "echo:SD card"},            // onboard SD card
```
kisslorand, line 40: I am not sure suppressing "No SD card" and "SD card" is a good idea.
On some motherboards you cannot access the onboard SD from a host (from PC by USB) as a storage device (to upload FW for ex.) card until you unmount id with M22. After that you have to mount it with M21 to be able to print from it.
If those messages are suppressed the user has no feedback of his actions (success or failure) regarding the mount and unmount of the SD Card. With my SKR v1.4 Turbo I used M21 and M22 a lot (custom commands).

digant73: they are not suppressed but displayed as toast messages. Otherwise I always get a popup message every time I plug in/out an SD card. Furthermore, notification are sent "Media inserted" or "Media removed". So the messages are even too much

kisslorand: Strange behaviour, I have these popups only if I mount or unmount the onboard SD and I do not have any message about media inserted or removed if I manipulate the onboard SD.
When I operate my printer from the PC I am not necessary right near my printer so a toast message would be long gone before I had the chance to check the printer. I would really hate to not have a confirmation on the TFT about the SD card being mounted or unmounted on the motherboard.
Just an idea, maybe you should check your Marlin config to not mount and unmount the SD card every time you insert or remove it. Another idea would be to keep personalized changes to your printer only, not for a public release.
Thank you for understanding this.

digant73: it' not a problem to remove those toast for me. so I will remove from my PR

rondlh (approved):
Tested, works fine

PR #2422, opened 2022-02-23
Add support for Delta Diagonal Rod Trim

Guilouz:
- Added support for Delta Diagonal Rod Trim adjustments in machine parameters settings for Delta printers.
- Updated French language file.


TFT/src/User/API/Language/language_keywords.h
```diff
@@ -66,6 +66,7 @@ extern "C" {
 #define LANG_KEY_STEALTH_CHOP                 "label_stealth_chop:"
 #define LANG_KEY_DELTA_CONFIGURATION          "label_delta_configuration:"
 #define LANG_KEY_DELTA_TOWER_ANGLE            "label_delta_tower_angle:"
+#define LANG_KEY_DELTA_DIAGONAL_ROD            "label_delta_diagonal_rod:"
```
digant73, line 69: remove a space here. IMHO, we should avoid to add other params. This kind of code takes a lot of FLASH (labels, code), it cannot be optimized and it's not even used by the majority of users.

Guilouz: Fixed. This params is used by all Delta printers users.

Guilouz:
@bigtreetech can you merge it please ?

PR #2446, opened 2022-03-16
Fix untimely mode switching while Marlin Mode is active

### LINKED ISSUE #2432: [BUG] TFT35 switches mode automatically when print is started, then nothing happens
### Description
In Marlin Mode I start a print from onboard media on the TFT35, it switches automatically to Touch Mode. After G34 it begins to heat the bed and it stops right before auto bed levelling parking the nozzle for no apparent reason. Tried to launch a print directly from Touch Mode and the result is the same. Trying to cancel the print causes the TFT to freeze.

### Steps to reproduce

1. Select a file from media
2. Start the print
3. Wait until auto bed levelling procedure

Expected behavior
Auto bed levelling, start printing.

Actual behavior
Nozzle is parked and heaters shutted down.

### Hardware Variant
TFT35 V3 + SKR E3 Turbo
Latest firmware build BIGTREE_TFT35_V3.0.27.x.bin + Marlin 2.0.9.3

### Additional Information
Attached video and photos of the issue. Included the config.ini file and Marlin Configuration.h

https://user-images.githubusercontent.com/82800736/156929551-ad96bccb-d485-4d52-8429-02fc22660754.mp4

### ZIP files
[config.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/8192875/config.zip)
[Configuration.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/8192888/Configuration.zip)

guruathwal:
- Fix mode switching while Marlin Mode is active.
  Prevent mode switching when serial is always on and print is started from other remote hosts. #2432.
- Misc clean-up

Fixes #2432.


TFT/src/User/API/parseACK.c
```diff
@@ -682,7 +682,7 @@ void parseACK(void)
                infoFile.source >= BOARD_MEDIA && infoFile.source <= BOARD_MEDIA_REMOTE &&
                ack_seen("SD printing byte"))  // if printing from (remote) onboard media
       {
-        setPrintResume(false);
+        setPrintResume(true);
```
digant73, line 685: that should remain setPrintResume(false). setPrintResume(true) is called in hostActionCommands function. The opposite happens for setPrintPause() function (true in parseAck function and false in false in hostActionCommands function). Basically, for resume we wait the **last** confirmation from hostActionCommands function. For pause, we wait the **first** confirmation from parseAck function.

TFT/src/User/API/Printing.c
```diff
@@ -374,6 +381,9 @@ void printRemoteStart(const char * filename)
 {
   infoHost.printing = true;  // always set (even if printing from onboard media)
 
+  if (MENU_IS(menuMarlinMode))  // do not process any printing info if Marlin Mode is active
```
digant73, line 384: mmh, just because parseAck will receive and handle also "Not SD printing" and "SD printing byte" and so it will call setPrintPause() and setPrintResume() (that will access infoPrinting, infoFile.source etc..), it is probably better to let printRemoteStart() initialize those data structures and simply avoid to switch to the Printing menu (so it is enough to move your  `if (MENU_IS(menuMarlinMode))...` just before `infoMenu.cur = 1;`. There should also no need to apply any check/change in printEnd (or printComplete). We will simply hear a beep at the end of the print (or on print cancel). But that should be (for me) a good thing to have

guruathwal: I would not recommend handling any print information at all apart from the status that a print is going on outside of the TFT. No print information should be processed while in Marlin Mode. 

>There should be also no need to apply any check/change in printEnd (or printComplete). We will simply hear a beep at the end of the print (or on print cancel). But that should be (for me) a good thing to have

This should be avoided because most of the hardware is required to be deinitialized for Marlin Mode to work properly like the buzzer and SD card (on some boards) is also used by Marlin. if the buzzer is not deinitialized by Touch mode, it will not work with Marlin Mode. This is also true for Encoder active signal to avoid encoder reading by Marlin mode when it should not.

digant73: Yes that's also true. So if (in printRemoteStart()) we do not set infoPrinting, infoFile etc..., infoPrinting.printing will always remains set to false (e.g. by infoPrinting definition) so functions printEnd() will do nothing (it exits immediately) and there is no need to check if we are in Marlin Mode, setPrintResume will not set infoHost.printing to true (because infoPrinting.printing is set to false).

TFT/src/User/API/parseACK.c
guruathwal: The true-false argument for setPrintResume is redundant.

TFT/src/User/Hal/LCD_Encoder.c
```diff
@@ -4,21 +4,19 @@
 
 #if ENC_ACTIVE_SIGNAL
 
-void LCD_Enc_InitActiveSignal(bool marlinType_LCD12864)
```
digant73, line 7: all the files in HAL folder have been made independent from infoSettings, machineSettings etc.... For that reason, that param was passed as an argument.

guruathwal: Ok noted

digant73: thanks for the changes. Now it's also better than the old version

TFT/src/User/API/Printing.c
```diff
@@ -36,7 +36,7 @@ void setExtrusionDuringPause(bool extruded)
 
 bool isHostPrinting(void)
 {
-  return infoHost.printing;
+  return (infoHost.status == HOST_STATUS_PRINTING);
```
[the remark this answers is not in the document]
guruathwal: Yes you are right.

TFT/src/User/API/Settings.h
```diff
@@ -64,6 +64,15 @@ enum
   sign_count
 };
 
+typedef enum
+{
+  HOST_STATUS_IDLE,
```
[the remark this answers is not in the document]
guruathwal: enums by default starts with 0.

TFT/src/User/API/Printing.c
```diff
@@ -742,7 +735,7 @@ void setPrintPause(HOST_STATUS hostStatus, PAUSE_TYPE pauseType)
 
   // in case of forcing update or printing from Marlin Mode (infoPrinting.printing set to "false") or
   // in case of printing from remote host or infoSettings.m27_active set to "false", always force to "false"
-  if (infoPrinting.printing && (infoFile.source >= BOARD_MEDIA_REMOTE || !infoSettings.m27_active))
+  if (!infoPrinting.printing && (infoFile.source >= BOARD_MEDIA_REMOTE || !infoSettings.m27_active))
```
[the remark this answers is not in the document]
guruathwal: That makes sense.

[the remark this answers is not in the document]
guruathwal: My earlier logic was kinda correct.

TFT/src/User/API/Vfs/vfs.h
```diff
@@ -18,12 +18,12 @@ extern "C" {
 
 typedef enum
 {
-  TFT_SD,
-  TFT_USB_DISK,
-  BOARD_MEDIA,
-  BOARD_MEDIA_REMOTE,
-  REMOTE_HOST
-} FS_SOURCE;
+  FS_TFT_SD,
```
digant73, line 21: my current PR already renamed those following the naming we provided in README.MD and config.ini files (we reported everywhere TFT SD, TFT USB, (REMOTE) ONBOARD MEDIA. Also FS_ is repetitive/ambiguous (FS_SOURCE was for File System source)

guruathwal: Adding a common prefix to a set of enum makes them easier to find and use.  If `FS_` looks ambiguous then it can be replaced with `FS_SOURCE`. Although for these and other places `FS` is being used for the same meaning which makes it not ambiguous.
I am unable to understand how this makes the whole cleanup in your PR invalid? the only cleanup relating to these enums is renaming `TFT_USB_DISK` to `TFT_USB` and replacing `BOARD` with `ONBOARD`.

TFT/src/User/API/Printing.c
```diff
@@ -412,7 +415,7 @@ bool printStart(void)
 
   switch (infoFile.source)
   {
-    case REMOTE_HOST:         // present just to make the code robust. It should never be executed
+    case REMOTE_HOST:  // present just to make the code robust. It should never be executed
```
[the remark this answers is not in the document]
guruathwal: I am sorry, I did not get what you are trying to say.

TFT/src/User/API/Vfs/vfs.h
digant73: hmm I preferred to avoid to use prefixes on the core files (Configuration.h, vfs, parseAck and interfaceCmd, Printing) because it is more compact and readable (of course we have to avoid to provide generic names such as RESUMED). When we use this def here and there it seems to me too much verbose to repeat a prefix. Also when we have variables already self explaining like infoFile.source it seems redundant to provide a prefix like FS_ when the rest of the def is not ambiguous (e.g. TFT_USB means any media on a TFT USB interface (a USB disk, pen drive, dongle etc.), ONBOARD_MEDIA means a media on a not known ONBOARD interface (could be SD, USB etc...). It seems to me also too much verbose HOST_STATUS_PRINTING. The prefix HOST_ was already clear and enough (e.g. HOST_PRINTING, HOST_PAUSED already indicate a status). Sure the type HOST_STATUS is correct

TFT/src/User/API/Printing.c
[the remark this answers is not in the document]
digant73: Forget it. Changing the order to:

```
    case ONBOARD_MEDIA_REMOTE:  // present just to make the code robust. It should never be executed
    case REMOTE_HOST:

```

already solved the problem. I meant that the comment was applied to both the two cases so I aligned it two chars after the longer one (so two chars after ONBOARD_MEDIA_REMOTE:)

PR #2449, opened 2022-03-21
Code reduction + cleanup

### LINKED ISSUE #2450: [BUG] Minor correction needed in README.md
### Description

Noticed a very minor formatting error in README.md to be corrected.
Under "Options to fully support Bed Leveling menu:" the line for g26_Mesh_Validation says (in configuration.h) twice.
First as part of the variable, with the gray background (this one needs to be removed as redundant) and then again afterwards as with the rest of the list of options.

### Steps to reproduce

1. look at the README.md on the main github page for the repository.
2. Notice the formatting error.

### LINKED ISSUE #2455: [BUG] Manual PID tune commands always result in a "Process aborted!" popup, even if succesfull
### Description
Running a manual PID tune (e.a. M303 S100 C2) will always result in a "Process aborted!" popup, even if the PID tune completes successfully. Also a M150 command is given, even if event LEDs are disabled in config.ini.

### Steps to reproduce
1. Run a PID tune via the console or ESP3D (M303 S100 C2)
2. Wait for the command to complete (back out of the console)

**Expected behavior**
The TFT can ignore the PID tune or inform about the actual results of the PID tune

**Actual behavior**
The TFT always gives a "Process aborted!" popup, independent of the actual PID process outcome.
Also an M150 is sent to the motherbord, even if event LEDs are disabled in the config.ini

### Hardware Variant
BTT TFT35V3.0

Latest TFT firmware (March 21st)
Latest stable Marlin 2.0.9.3 (March 13th, MKS Monster8 V1.0)

### Additional Information
I can follow the process in PID.c, the pidSucceeded is false at start and is never set to true if a manual PID tune is started (even if pidUpdateStatus is called with succeeded is true). The calls to LED_SetEventColor are not guarded with the event_led parameter in config.ini resulting in a M150 call, which is not supported in my case.

digant73:
**IMPROVEMENTS:**
* Code reduction: Code reduction in parseAck.c, interfaceCmd.c and other files. Code sharing is applied. This allowed to reduce the FLASH code space by about 2KB. 
* Code cleanup: No need to make configurable at runtime bootscreen and safety alert messages when they are already configurable at compile time (SHOW_BTT_BOOTSCREEN, BTT_BOOTSCREEN_TIME and SAFETY_ALERT). There is also no advantage on removing bootscreen if Marlin fw is not ready to reply to TFT. About 500 bytes of FLASH now reduced (plus the space for the extra labels).
* Cleanup: Some definitions in Printing.h/c and some labels reviewed in order to use the same naming convention.
* Conditions in switch statement in Printing.c ordered from TFT_SD to REMOTE_HOST (as they are in VFS.c).

**BUGFIXES:**
* Fixed the typo reported in #2450.
* Fixed #2455.
* Fixed freeze on print abort when printing from onboard: A bug introduced in one of the last merged PR.
* Fixed missing ":" for "label_onboard_usb" in language_keywords.h.

fixes #2450
fixes #2455


Guilouz:
Can you add a small fix in language_keywords.h, missing ":"

#define LANG_KEY_ONBOARDUSB                   "label_onboardusb"

digant73:
> Can you add a small fix in your PR, in language_keywords.h, missing ":"
> 
> #define LANG_KEY_ONBOARDUSB "label_onboardusb"



> Can you add a small fix in your PR, in language_keywords.h, missing ":"
> 
> #define LANG_KEY_ONBOARDUSB "label_onboardusb"

sure, done.
For FR language I also set STRING_TFT_USB_NOT_DETECTED   "Aucune clé USB détectée." is that correct?

digant73:
> Can you add a small fix in your PR, in language_keywords.h, missing ":"
> 
> #define LANG_KEY_ONBOARDUSB "label_onboardusb"



> Can you add a small fix in your PR, in language_keywords.h, missing ":"
> 
> #define LANG_KEY_ONBOARDUSB "label_onboardusb"

sure, done.
For FR language I also set STRING_TFT_USB_NOT_DETECTED   "Aucune clé USB détectée." is that correct?

Guilouz:
Yes it’s correct.

TFT/src/User/config.ini
```diff
@@ -27,8 +28,8 @@
 # Distribution date: 2021-05-15
 # Firmware source: https://github.com/MarlinFirmware/Marlin/releases
 #
-# In order the TFT firmware is able to provide all of its functionalities/features,
-# ensure that the following options are enabled in Marlin firmware.
+# In order for the TFT's firmware to be able to provide all of its functionalities/features,
+# the following options must be enabled in Marlin firmware.
 #
```
guruathwal, line 33: It will be much better to keep it straightforward like:
`To use all the features & functionalities supported by the TFT, the following options must be enabled in Marlin firmware.`

TFT/src/User/my_misc.c
```diff
@@ -145,7 +145,7 @@ const char *stripHead(const char *str)
 {
   // example: " :/test/cap2.gcode\n" -> "test/cap2.gcode\n"
 
-  while (*str == ':' || *str == '/' || *str == ' ')
+  while (*str == ' ' || *str == ':' || *str == '/')
```
kisslorand, line 148: In my opinion it's best to put the most probably frequent conditionals first to avoid unnecessary checks

digant73: this function is typically used for such kind of strings:
`M20 SD:      /my_filder/my_file.gco`
where the keyword `M20 SD:` was previously removed so the string passed to this function is `      /my_filder/my_file.gco`. So yes, ` ` is the only character that will be repeated and the first one that will appear. Then one `/` and finally `:` (that will probably never be present). Also this function is called when the system is on idle (not printing).

TFT/src/User/my_misc.c
```diff
@@ -140,12 +140,12 @@ double stringToDouble(char *str, char **endptr)
   return val * sign;
 }
 
-// strip out any leading " ", ":" or "/" character that might be in the string
+// strip out any leading " ", "/" or ":" character that might be in the string
 const char *stripHead(const char *str)
 {
-  // example: " :/test/cap2.gcode\n" -> "test/cap2.gcode\n"
+  // example: ":    /test/cap2.gcode\n" -> "test/cap2.gcode\n"
```
kisslorand, line 146: How can " " appear there?

digant73: see section "Printing from Host" in README.MD. There are also other examples

kisslorand: OK, thank you. It's clear now. :)

PR #2483, opened 2022-03-31
Bugfix  Start/Resume Icon not updated 

### LINKED ISSUE #2481: [BUG] Start/Resume Icon not updated
After a M600 (popup to purge and continue) or filament runout and print restarted, resume button is not updated, it's always in resume status.

On this photo machine is actually in printing mode.

IMG_3146

I have also notified after print is finished, LED Event go to green but after exiting resume menu Event LED turns off instead of white.

digant73:
**BUGFIXES:**
* Fixed #2481: That bug was a side effect of PR introducing infoHost.status (bug already existing but emerged with that new variable). Another bug emerged with that variable was a freeze aborting a print from remote onboard (e.g. started by ESP3D and then aborted from Printing menu).
* Some cleanup: Global types and variables moved from settings.h (previously defined here just to avoid a compile error) and includes.h to main.h + other minor cleanup.

fixes #2481


TFT/src/User/API/Printing.c
```diff
@@ -714,9 +711,10 @@ void setPrintAbort(void)
 
 void setPrintPause(HOST_STATUS hostStatus, PAUSE_TYPE pauseType)
 {
-  if (infoHost.status == HOST_STATUS_IDLE)  // in case print was aborted or completed, nothing to do
+  // in case print was aborted or completed or printAbort() is aborting the print, nothing to do
+  if (infoHost.status <= HOST_STATUS_STOPPING)
```
kisslorand, line 715: infoHost.status can never be equal to HOST_STATUS_STOPPING at this point

fiveorders:
Hello, I am using Octopus Pro with Marlin 2.9.3 and TFT70. Filament runout workflow works as normal but it doesn't automatically resume. Nozzle goes to its place and stays there until we press the resume button on the screen. If we press the resume button earlier it says Busy Processing

Previous firmware was resuming auto.

digant73:
@fiveorders that logic is correct (and it is there since a lot of time). Once the filament is put in the filament sensor you can press on resume button to resume the print.

fiveorders:
Actually not, because it doesn't let us press resume until it goes to its position and stays there. This is causing blobs. Normally it was going to its place and starting printing immediately. We just press the continue button on the screen after the filament is loaded and extruded. It was like this with the previous BTT firmware 27.

digant73:
are you printing from onboard (marlin's SD slot) or TFT (TFT's SD or USB)?

fiveorders:
Thank you for your detailed reply. I print through USB, Actually, I am on 2.9.3 and I was on the same FW with the previous TFT firmware. Now I have both the GD chip version of TFT70 and they both do the same thing. I didn't change anything on Marlin. 

It purges over the printed part for a second until I press resume. At least purges 2-3cm

digant73:
Ok. It could be possible that your issue is related to this PR #2466 that was temporary merged by BTT. I have no GD version and everything is working properly (from TFT SD, USB and onboard. Even printing from ESP3D).

fiveorders:
I have both gd version and stm version both do the same. Is it the icon update problem? Because it also extrudes some filament over the starting point. I will do some more tests but I don't know where to start. 

do you think I should install a bugfix version of Marlin? I don't think this is related to marlin though because it was working

digant73:
if you are not on the current TFT fw (which fixed also the icon update issue), update your fw. If the issue is present when printing from TFT's USB, you don't need to update Marlin

fiveorders:
> if you are not on the current TFT fw (which fixed also the icon update issue), update your fw. If the issue is present when printing from TFT's USB, you don't need to update Marlin

https://github.com/digant73/BIGTREETECH-TouchScreenFirmware/tree/bugfix-2022-03-31

I have downloaded this bugfix. Is this not the firmware you are mentioning? I normally modify the menus and config for my purpose. I did nothing on this, directly downloaded and flashed. It still does the same thing. And also Z axis goes up 10mm up when you turn on the printer.

digant73:
Absolutely NO!
You must always download the official BTT fw from https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware from the **master** branch

fiveorders:
Dear @digant73 I have tried that as well. The same thing happening M600 emulation on or off doesn't matter. Everything works as expected until it goes back to the point. It doesn't start printing without me pressing Resume. But on top ACK it says "Printing" although it is not. Any ideas are appreciated.

fiveorders:
@digant73 how do you tell which version is the firmware? the one which was working #define CONFIG_VERSION 20210605 has this date on the Configuration.h file.

digant73:
@fiveorders Do you have the runout sensor connected to mainboard or TFT?

fiveorders:
> @fiveorders Do you have the runout sensor connected to mainboard or TFT?

Sensor connected to the Octopus Pro board. Problem solved. Thanks!

https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2494

PR #2494, opened 2022-04-09
Auto resume by host action command

Msq001:
### Description

When filament runout sensor is inserted to motherboard, And printing start from TFT SD or USB, TFT sends `M75` to the mainboard to notify the motherboard to start detecting filament runout sensor.

After the motherboard detects the filament runout, it will notifies TFT by `host action command` and TFT printing will be paused by `setRunoutAlarmTrue()`.  since `infoHost.status` of TFT is always `HOST_STATUS_IDLE`, printing cannot be resumed automatically by  `setPrintResume()`

### DIFF
--- a/TFT/src/User/API/Printing.c
+++ b/TFT/src/User/API/Printing.c
@@ -715,11 +715,11 @@ void setPrintPause(HOST_STATUS hostStatus, PAUSE_TYPE pauseType)
   // in case print was completed or printAbort() is aborting the print,
   // nothing to do (infoHost.status must be set to "HOST_STATUS_IDLE"
   // in case it is "HOST_STATUS_STOPPING" just to finalize the print abort)
-  if (infoHost.status <= HOST_STATUS_STOPPING)
-  {
-    infoHost.status = HOST_STATUS_IDLE;  // wakeup printAbort() if waiting for print completion
-    return;
-  }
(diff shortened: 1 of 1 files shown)


Msq001:
@digant73 
Hello, Please check this problem. is `if (infoHost.status <= HOST_STATUS_STOPPING)` necessary ?

digant73:
@bigtreetech yes it is needed to avoid any freeze in case we abort a print from onboard. See line 541 in parseAck.c:

```
      // wait until infoHost.status is set to "HOST_STATUS_IDLE" by setPrintPause()
      loopProcessToCondition(&isHostPrinting);

```

However we can easily fix the issue with runout on mainboard simply switching the check on HOST_STATUS_STOPPING after the change on infoPrinting.pause

So doing something like:

```
  // no need to check it is printing when setting the value to "false"
  infoPrinting.pause = false;

  // in case print was completed or printAbort() is aborting the print,
  // nothing to do (infoHost.status must never be changed)
  if (infoHost.status <= HOST_STATUS_STOPPING)
    return;

```

digant73:
@bigtreetech I didn't understand if you were asking me to provide the fix. In that case it was included in #2496.

digant73:
@bigtreetech I didn't understand if you were asking me to provide the fix. In that case it was included in #2496.

PR #2497, opened 2022-04-12
Added X and Y coordinates in MBL and Mesh Editor menus

### LINKED ISSUE #2385: [FR] Manual Mesh Bed Leveling Screen X&Y Values and I&J (grid point number) Values Should be Shown on the Screen
I want to see the X&Y coordinates while probing the bed, this will let me adjust the probing margin accordingly. I can optimaze the margin to probe as much as off the bed but if I don't know the probing point coordinates I can not decide how to tune the probing margin from the edges. Also that would be better to show the grid point numbers (e.g. for a 3x3 grid I:0 J:0 is the first point I:2 J:2 is the last point).
I don't mean the mesh grid view screen. That screen also have it but only I and J values of the point.

So my FR requests to solve this issue are:
-Add Point I&J values on the MBL probing screen (nice to have)
-Add Point X&Y values on the MBL probing screen (must to have)
-Add Point X&Y values on the MBL grid point screen (must to have)

a1
a2
a3
a4

### LINKED ISSUE #2387: [Q] FYSETC SD-WIFI
Is there anyone that manage FYSETC SD-WIFI card to get working with TFT35? I have one and I got mostly sd card init failed. It seems there is an issue maybe with timing or something. Not sure it is quite TFT35 related. I managed to fix fw of SD-WIFI so that WebDav interface works but it works only if it is not inserted into TFT35 card slot. It appears that TFT blocks somehow the card, so ESP is unable to read anything from the real card. Any ideas?

digant73:
**IMPROVEMENTS:**
* Added X and Y coordinates in MBL and Mesh Editor menus: It implements #2385.

resolves #2385

IMG_20220412_160842

IMG_20220412_161454

### DIFF
--- a/TFT/src/User/API/coordinate.c
+++ b/TFT/src/User/API/coordinate.c
@@ -105,6 +105,11 @@ void coordinateSetAxisActual(AXIS axis, float position)
   curPosition.axis[axis] = position;
 }
 
+void coordinateGetAllActual(COORDINATE *tmp)
+{
+  memcpy(tmp, &curPosition, sizeof(curPosition));
+}
+
 void coordinateQuerySetWait(bool wait)
 {
   coordinateQueryWait = wait;
--- a/TFT/src/User/API/coordinate.h
+++ b/TFT/src/User/API/coordinate.h
@@ -49,6 +49,7 @@ float coordinateGetExtruderActual(void);
 void coordinateSetExtruderActualSteps(float steps);
 float coordinateGetAxisActual(AXIS axis);
 void coordinateSetAxisActual(AXIS axis, float position);
+void coordinateGetAllActual(COORDINATE *tmp);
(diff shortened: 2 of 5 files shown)


Guilouz:
While you are in the Mesh Editor, do you think it is possible to gray out the non-accessible points on the bed for delta printers?

digant73:
I don't see a way to know the unreachable points

Guilouz:
Non reachable points stay at 0.0. Maybe add a color (grey) for points are at 0.0 no?

digant73:
The not reachable points, if any, should depend on the probe position, printer hw etc. I cannot make any assumption about that

sarvenn:
Hi @digant73,
Thnx for the PR. Will Mesh Tuner update be available at ABL(BLTouch) as well?

digant73:
@sarvenn yes

PR #2500, opened 2022-04-20
Bug fixes on Printing menu + cleanup

digant73:
**BUG FIXES:**
* Elapsed print time stopped during a print pause/filament runtout etc.: The elapsed print time must not be stopped (e.g. as during the heating phase). It measures the total time to reach the print completion not the active time spent to print. Fix done in function updatePrintTime().
* Progress bar on Printing menu not fully supporting a switch from file based progress to Slicer based progress: As also supported by Marlin, the progress bar is now (as it was before one of the last merged PR) properly supporting the switch from file based to slicer based progress (it means the possibility the progress can be reduced). The fix is done in function reDrawProgress() (a function that is expected to be called only 100 times in a whole print). In general, the Printing menu must make no assumption about the Printing API and the Remote Printing API listed in **Printing from Remote Host** section in README.MD. Any value set by the API must be simply displayed by the Printing menu.
* Fixed wrong progress display in case a print is aborted and the progress was controlled by slicer: The progress was set to 0% instead of 100%. Fix done in function printComplete().
* Fixed not 100% progress display when a print is completed: It can happen (in particular with files not big or printing from onboard) that the progress is moved to 100 just when the print is also completed (print status changed to **not printing**). In that case, the latest merged PR will not update the progress to 100% on screen (progress bar and value not updated. E.g. it will remain to 99%). Fix done in function menuPrinting().
* File based progress formula restored to linear function **file_pos(t) / file_size** as also used by any known mainboard firmware (e.g. Marlin). It didn't make any sense to provide a function not linear and variable during the print removing the comments contribution only when they are read from file (all the comments contribution should be removed offline before starting the print (file_size should be constant during the entire print)). The result of that formula was wrong even in case the print is based on a repeated pattern (e.g. 100 bytes of gcode + 100 of comments on a print duration of 100 minutes will provide a 33% (instead of 50%) of progress after 50 minutes while the linear function will always provide the correct value (50% at 50% of the print, 75% at 75% of print etc...).

**MINOR IMPROVEMENTS:**
* Some cleanup:
  * clearInfoFile() not needed when exiting from Printing menu (it was already called in initMenuPrinting() just before starting the print).
  * exitFolder() not needed in Printing API (it is needed by menuPrintFromSource() (if any was used to start the print) so it is called when exiting from Printing menu. It is compliant (it works) with any kind of print (TFT, onboard, remote)).
  * Usual code style review.
* Minor code reduction.

### DIFF
--- a/TFT/src/User/Menu/PrintingMenu.c
+++ b/TFT/src/User/Menu/PrintingMenu.c
@@ -10,11 +10,11 @@ const GUI_RECT printinfo_val_rect[6] = {
   {PS_ICON_VAL_X, PS_ICON_VAL_Y, PS_ICON_VAL_SM_EX, PS_ICON_VAL_Y + BYTE_HEIGHT}
 };
 
-#define PROGRESS_BAR_RAW_X0   (START_X)                             // X0 aligned to first icon
+#define PROGRESS_BAR_RAW_X0   (START_X)                                 // X0 aligned to first icon
 #ifdef PORTRAIT_MODE
-  #define PROGRESS_BAR_RAW_X1 (START_X + 3*ICON_WIDTH + 2*SPACE_X)  // X1 aligned to last icon
+  #define PROGRESS_BAR_RAW_X1 (START_X + 3 * ICON_WIDTH + 2 * SPACE_X)  // X1 aligned to last icon
 #else
-  #define PROGRESS_BAR_RAW_X1 (START_X + 4*ICON_WIDTH + 3*SPACE_X)  // X1 aligned to last icon
+  #define PROGRESS_BAR_RAW_X1 (START_X + 4 * ICON_WIDTH + 3 * SPACE_X)  // X1 aligned to last icon
 #endif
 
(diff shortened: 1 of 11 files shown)


TFT/src/User/Menu/PrintingMenu.c
```diff
@@ -701,7 +697,6 @@ void menuPrinting(void)
         }
         else
         { // Main button
-          clearInfoFile();
```
kisslorand, line 704: Do not leave garbage in memory. From this point on  the data is not needed anymore, it is just a waste of precious memory space. Yes, it will be cleared in the next print but why keep it in the memory???
Set it free! :)

digant73: it was already cleared at the beginning of this print (not the next one) in initMenuPrinting().

kisslorand: Yes, you're right. In this case it should be removed from "menu.c" (line 1232~1233) also.

PR #2504, opened 2022-04-22
Improved GUI performance

digant73:
**IMPROVEMENTS:**
* Improved performance on GUI_DispOne() function: With the exception of text requiring transparency (GUI_TEXTMODE_TRANS), each character in a text is now displayed in a char by char mode (an LCD window is used) instead of pixel by pixel mode (an LCD window made by only one pixel). The improved performance are present in any menu displaying text (e.g. Printing menu) and on any TFT variant.

### DIFF
--- a/TFT/src/User/API/UI/GUI.c
+++ b/TFT/src/User/API/UI/GUI.c
@@ -108,7 +108,7 @@ void GUI_DrawPoint(uint16_t x, uint16_t y)
 void GUI_FillRect(uint16_t sx, uint16_t sy, uint16_t ex, uint16_t ey)
 {
   uint16_t i=0, j=0;
-  LCD_SetWindow( sx, sy, ex-1, ey-1);
+  LCD_SetWindow(sx, sy, ex-1, ey-1);
   for (i=sx; i<ex; i++)
   {
     for (j=sy; j<ey; j++)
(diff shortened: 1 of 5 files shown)


TFT/src/User/API/UI/GUI.c
```diff
@@ -606,49 +606,89 @@ void GUI_DispOne(int16_t sx, int16_t sy, const CHAR_INFO *pInfo)
 {
   if (pInfo->bytes == 0) return;
 
+  uint8_t w = pInfo->pixelWidth;
+  uint8_t h = pInfo->pixelHeight;
+  uint16_t bitMapSize = (h * w / 8);
+  uint8_t font[bitMapSize];
```
guruathwal, line 612: It is not an optimization when you duplicate a variable knowing that it can be read from an already existing variable. Now for each character, the width and height are copied to a new variable instead of directly accessing the variable resulting in unnecessary CPU cycles.

digant73: Yes, but they are used as conditions in loops (e.g. y < **h**;) or repeated more than one time in the code (e.g. sx + **w**, y * **w**). For that reason they have been loaded in new variables (e.g. avoiding access with **->** in loops etc...). Otherwise, we should be sure the compiler is able to optimize the code (e.g. loading and maintaining pInfo->pixelWidth in a register before it is used in the whole function).

digant73:
@bigtreetech If function LCD_WR_16BITS_DATA() could be also used just to move to next display address without any write of pixel color on the display, then the optimization could be also extended to text with transparency. I tried to make a custom LCD_WR_16BITS_DATA() but I was not able to force only the display address change without also writing a pixel color on the screen.

The idea was something like that:

if (buf[i] == transPixel)  // if it is a transparency pixel, move to next display address
  LCD_WR_16BITS_NO_DATA();  // move to next display address
else
  LCD_WR_16BITS_NO_DATA(buf[i]);  // write pixel color and move to next display address

ETE-Design:
@digant73 Have an Sidewinder X1 - Have looked at your Thingiverse... And your firmware work great when I use it... But would like to compile my own... What ENVS have you used to Compile? Also is it Bitmap Folder for TFT28 or TFT35 you use?

digant73:
@ETE-Design I use vscode.

ETE-Design:
> @ETE-Design I use vscode.

Same here... But what ENVS have you set in PlatformIO.ini so it work with MKS TFT28 on an X1?

digant73:
It is MKS_TFT28_V1_0_ARTILLERY

digant73:
https://github.com/digant73/BIGTREETECH-TouchScreenFirmware/tree/artillery_tft_fw_1.27.x_patch_8
The env is already set as default

ETE-Design:
@digant73 I see... Do you have it with newest changes anywhere?

ETE-Design:
@digant73 That worked on my X1... Funny thing is, in the past I updated my X2 TFT Firmware with yours, see the picture :-) Now it won't update when trying to use your newest version in Thingiverse... It only try to update from the Config.ini nothing else, can you help again?
IMG_20220523_212039

digant73:
closing due to no interest from BTT

oldman4U:
😥 but understandable.

No version release for more than one year.

PR #2505, opened 2022-04-23
Increase value display length after decimal in list view

### LINKED ISSUE #2423: [BUG] Rounding of parameter values
In machine settings the 3rd digit after the decimal point is always rounded up or down.

For some parameters (like diagonal rod, diagonal rod trim, etc) we need to have precise values to 4 digits after the decimal point.

In example here we need to adjust settings with precise values (in red background).

Capture d’écran 2022-02-23 à 16 21 25
 
Currently if I enter the value 314.8381, the result saved by TFT is 314.84 and they are saved in the EEPROM as is.

guruathwal:
- Increase value display length after the decimal point in list view.
- increase float precision for Delta configurations in **MachineParameters.c** and Numpad.c

Fixes #2423


digant73:
Please add also the changes in MachineParameters.c proposed in #2423 so #2423 is fully fixed.

TFT/src/User/API/UI/ListItem.c
```diff
@@ -33,18 +35,20 @@ char * getDynamicTextValue(uint8_t i)
 // save dynamic value (upto 7 digits) ( i : index of the value position, value:float value)
 void setDynamicValue(uint8_t i, float value)
 {
-  if (value < 1.0f)
-    sprintf(dynamic_text_value[i], "%.3f", value);
+  if (value < 10.0f)
```
digant73, line 38: maybe it should be considered abs(value) and the number of decimal digits should be reduced by 1 in case the value is negative (there is a - sign)

TFT/src/User/API/UI/ListItem.c
```diff
@@ -33,18 +35,20 @@ char * getDynamicTextValue(uint8_t i)
...
+    sprintf(dynamic_text_value[i], "%.5f", value);
+  if (value < 100.0f)
+    sprintf(dynamic_text_value[i], "%.4f", value);
   else if (value < 1000.0f)
-    sprintf(dynamic_text_value[i], "%.2f", value);
+    sprintf(dynamic_text_value[i], "%.3f", value);
   else
     sprintf(dynamic_text_value[i], "%.1f", value);
```
digant73, line 45: maybe it's .2f

Guilouz:
Tested, list display 4 digits after the decimal point.

Numpad also needs to be updated, it only shows 3 digits after the decimal point and rounds the last digit.

Guilouz:
@digant73 Do you know where I can change it in Numpad ?

Guilouz:
@digant73 Thanks it's working and rounded value is gone now.

PR #2517, opened 2022-05-05
Improved Leveling Corner menu

### LINKED ISSUE #2516: Auto bed leveling / L Corner controls configuration
I am still getting used to all the options the firmware has now compared with the old version I was used to use, thanks to the developers for this amazing work. 
I found the "L corner" option under the Auto Bed Level menu 

Level01

and I see that the range for defining the edges is from 60 (min) to 100 (max). 

Level02

Where can I change that range to a much lower minimum?

I searched on the config.ini file and I only find reference to this setting, which was already low, i.e. 10:

```
## Leveling Edge Distance (Manual Leveling)
#   Inset distance from bed edges. This distance is added to minimum X & Y bed coordinates and
#   subtracted from maximum X & Y bed coordinates to calculate manual leveling points.
#   Unit: [distance in mm]
level_edge_distance:10

```

I also searched in the configuration.h file and I found two references there but none of them was set to the minimum of 60 that I see on my TFT. One under leveling settings:
`#define LEVELING_EDGE_DISTANCE    20  // (mm) Inset distance from bed's edge for calculating leveling point location (Default: 20)
`
and the other under manual leveling:

```
// Manual Leveling
// Move to four corner points to Leveling manually (Point 1, Point 2, Point 3, Point 4).
#define LEVELING_EDGE_DISTANCE_DISPLAY_ID "X/Y"
#define LEVELING_EDGE_DISTANCE_MIN          0                    // Default: 0
#define LEVELING_EDGE_DISTANCE_MAX         100                    // Default: 100

```

SKR 1.3
TFT35 V3.0

digant73:
**IMPROVEMENTS:**
* Improved Leveling Corner menu: Edge distance can be set in the range 0-100 (as in Manual Leveling menu). In case the edge distance is below `max(probe offset X, probe offset Y) + 1` then a warning popup message is displayed asking to use the safe min value given by `max(probe offset X, probe offset Y) + 1`. The default edge distance (set when entering in the menu) is given by `max(level_edge_distance, max(probe offset X, probe offset Y) + 1)` where `level_edge_distance` is set in `config.ini`. Furthermore, when exiting from the menu the original edge distance value (for Manual Leveling menu) is restored.

resolves #2516


radek8:
@digant73, your edit is working, but the newly set value will not be saved.
If I leave the LevelCorner menu, the value will be overwritten by the original value.

digant73:
But the feature has nothing to save (even in the past). There was an issue that was fixed by

radek8:
OK, it's working anyway. Thank you.
I made my own adjustment for myself.

Issue #2323, opened 2022-01-16
[BUG] Incorrect spelling in config.ini and config_rrf.ini


rondlh:
### Description
Not really a bug, but something that could be improved. The spelling of the following words in config.ini and config_rrf.ini is incorrect.
Note that some of these words are used in keyword, so if they would be updated then this will have consequences elsewhere. At least updating the comments would make sense in my view.
Here the list of words I found:
baudrate --> baud rate
english --> English
fullname --> full name
speeed --> speed
encorder --> encoder
etc --> etc.
Backgroud --> Background
minimun --> minimum
Fullscreen --> Full screen
fullscreen --> full screen
characthers --> characters
invertion --> inversion
usefull --> useful
reacheable --> reachable 
miliseconds --> milliseconds
conjuction --> conjunction
Filamanent --> Filament
begining --> beginning
Feedrate --> Feed rate
feedrate --> feed rate

digant73:
Thanks for the checks. Fixed on my PR.
I didn't changed the following words. They should be ok also as they are (commonly used).
baudrate
english
fullname
fullscreen
feedrate

PR #2519, opened 2022-05-12
Machine parameter menu fix, temp set fix, numpad fix

kisslorand:
This PR fixes some menu related bugs:
 1. Exiting from machine parameter settings menu, responding to save or not the settings to eeprom, the menu didn't exit
 2. Setting the target temperature of any heater was erratic, especially when using the numpad. Many times the new value wasn't retained. When using the +- buttons sometimes the values skipped back to some random/previous value. It was so frustrating to prepare the bed temperature for a print and after 10 or so minutes to realize that it is not heated at all...
 3. Pressing "0" in numpad when it was the first key pressed or value shown was already "0" the acoustic feedback was one that indicated erroneous input. It's not wrong to press "0" in any case.

   With this PR after responding to "Save settings to the EEprom" when leaving the machine parameter settings menu, it now exits.
   The target temperature setting for the heaters is rewritten from an open-loop approach to a closed-loop approach. Every phase of it is properly tracked and handled. Setting the target temperature from the TFT is not perturbed anymore by external sources, when selecting rapidly with many successive "+" or "-" buttons, the transmission is not cluttered anymore, only just a few intermediate values are sent, not all of them and than - eventually - the last one.
   PR #2410 is included also as it was needed for the proper and simpler implementation of the new closed-loop approach for the target temperature setting of the heaters and for whatever reason it was delayed for so long, that reason doesn't exist anymore.

Also, there's a slight RAM and Flash size reduction too.

#### A few days later...
  Fixed a typo reported in #2323


Guilouz:
@kisslorand Just try your Pr and there is issue with it, print start without waiting for temperature.

kisslorand:
Can you please attach gcode file?

Guilouz:
@kisslorand Here. :
[Minimal_Surface_Fruit_Bowl2.gcode.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/8703775/Minimal_Surface_Fruit_Bowl2.gcode.zip)

kisslorand:
Yes, I saw that, it was the key to find the bug. I use M105 between them, that's why I couldn't replicate your issue.

kisslorand:
> @kisslorand Just try your Pr and there is issue with it, print start without waiting for temperature.

@Guilouz 
The problem is fixed, did you try it?

Guilouz:
Yes it's working.

Guilouz:
@kisslorand But have a strange issue with temperature when using prepare menu, only hotend heat and bed stop to heat, see this video : 
https://drive.google.com/file/d/1oHNFn26sXpfIisUq8oOqpQkoFSHZqQqY/view?usp=sharing

kisslorand:
@Guilouz 
Thank you for your feedback. It was a problem indeed, a stupid mistake on my part.
It is fixed now, please check it out if you can.

Guilouz:
@kisslorand Yes it's ok now.

PR #2521, opened 2022-05-13
Fix GitHub Actions

thisiskeithb:
### Description

Various Actions cleanup and fixes in response to [CVE-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/) that broke the auto-building & committing of updated biniares & config.

*Note: The Build Binaries action will run after the next PR merge/commit since the `.github` folder is excluded.*

### Benefits

Binaries & config will update.

### Related Issues

None. Discovered broken actions while checking out recent updates.

### DIFF
--- a/.github/workflows/buildBinary.yml
+++ b/.github/workflows/buildBinary.yml
@@ -17,85 +17,127 @@ jobs:
     if: github.repository == 'bigtreetech/BIGTREETECH-TouchScreenFirmware'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
       - name: Setup Python
         uses: actions/setup-python@master
         with:
           python-version: '3.x'
-      - name: Install Platform IO
+
+      - name: Install PlatformIO
         run: |
           python -m pip install --upgrade pip
           pip install -U platformio
+
       - name: Build TFT24 V1.1
         run: platformio run --environment BIGTREE_TFT24_V1_1
+
       - name: Build TFT28 V1.0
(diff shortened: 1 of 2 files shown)


.github/workflows/buildBinary.yml
```diff
@@ -17,85 +17,127 @@ jobs:
...
-        run: git config user.email "${GITHUB_ACTOR}@users.noreply.github.com" && git config user.name "${GITHUB_ACTOR}" && find "Copy to SD Card root directory to update/" -name 'config.ini' -exec git add {} \;
+        run: |
+          find "Copy to SD Card root directory to update/" -name 'config.ini' -exec git add {} \;
+
       - name: Push
-        uses: actions-x/commit@v2
+        uses: actions-x/commit@v6
         with:
+          email: 38851044+bigtreetech@users.noreply.github.com
+          name: bigtreetech
```
thisiskeithb, line 141: Use full GH account details since `${GITHUB_ACTOR}` was not working correctly.

.github/workflows/buildTest.yml
```diff
@@ -8,81 +8,105 @@ on:
...
     - '**/*.md'
 
-  push:
-    paths-ignore:
-    - Bootloader/**
-    - Copy to SD Card root directory to update/**
-    - images/**
-    - readme/**
-    - '**/*.md'
-
```
thisiskeithb, line 18: This action no longer needs to run on push/direct commits since it is handled under the Build Binaries action.

thisiskeithb:
I would *highly recommend* moving to Github's Releases feature instead of bundling built binaries in up in the source code, but that's for another discussion/PR 🙂

PR #2538, opened 2022-06-09
Hotend MPC (model predictive control) tune

kisslorand:
With the recent major Marlin release (v2.1.x) we have now a new temperature control for the nozzle(s), the MPC (Model Predictive Control). This PR adds the ability to make use of it, you can MPC tune your nozzle(s) and/or you can edit the input parameters for it (heater power, filament heat capacity).
Let's say you change filament from PETG to nylon. It's enough to change the filament heat capacity and you're good to go, no need for any new calibration.
The input parameters are marked with the letters used by the M306 command:
 - P: Heater power in watts
 - H: Filament Heat Capacity (joules/kelvin/mm)

The MPC capability of Marlin is automatically detected, no config or user intervention is needed for it to be active.

More about MPC here: https://marlinfw.org/docs/features/model_predictive_control.html
and here: https://marlinfw.org/docs/gcode/M306.html

The PID tune from TFT has some slight rework done. It doesn't continue to tune all the heater elements if an error occurs. The fan is not started automatically anymore, for heat losses to part cooling fan there is a separate parameter in Marlin ("PID_FAN_SCALING"), there are tuning procedures for it.
This PID rework also reduces its flash size.

Both PID and MPC tuning menus use a blocking splash screen that let's the user know that the printer is busy. No touch input is processed until the process is finished with success or failure.

It worth mentioning that MPC is only available for the nozzle(s), the heated bed is still PID controlled. MPC and PID cannot coexist for the nozzle(s) so if MPC is enabled in Marlin PID tuning will be available, if exists, only for the heated bed.

Issues/limitations:
Unlike PID tuning from TFT, MPC tuning from the TFT is done one by one, you have to do it separately for all the nozzles you might have.

Notes:
Due to the high precision of filament heat capacity the precision of list view and numpad input is increased. It is now up to 5 digits after the decimal point.
Unlike in a case of regular popup, with the splash on, the menu's header is updated with actual values.


bryan065:
Any chance you can add the console report as a popup? It'd be nice to be able to see the reported numbers without having to manually go into the terminal to run `M306 T`.

    MPC Autotune finished! Put the constants below into Configuration.h  
    MPC_BLOCK_HEAT_CAPACITY 18.42  
    MPC_SENSOR_RESPONSIVENESS 0.2176  
    MPC_AMBIENT_XFER_COEFF 0.0664
    MPC_AMBIENT_XFER_COEFF_FAN255 0.0998

Bapp1:
ive just installed and updated my tft firmware but it asks for icons update required - i cant find them and printer now wont get away from this screen so i cant do anything even after a power off or pressing reset

PR #2541, opened 2022-06-14
Enabling Leveling in RRF

### LINKED ISSUE #2537: ABL Buttons Missing with RRF
After updating to recent firmware on my E3 RRF/TFT35 E3 V3 system, I seem to have lost the ABL buttons on the TFT. Has something changed so that these buttons are now disabled?

radek8:
Enabling Leveling in RRF

After merging #2179, users with RRF no longer see the leveling icon.
This fix sets fixed ABL leveling for RRF.

resolves #2537

### DIFF
--- a/TFT/src/User/API/Settings.c
+++ b/TFT/src/User/API/Settings.c
@@ -250,7 +250,7 @@ void setupMachine(FW_TYPE fwType)
   else if (infoMachineSettings.firmwareType == FW_REPRAPFW)
   {
     infoMachineSettings.softwareEndstops = ENABLED;
-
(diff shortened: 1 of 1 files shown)


TFT/src/User/API/Settings.c
```diff
@@ -250,7 +250,7 @@ void setupMachine(FW_TYPE fwType)
   else if (infoMachineSettings.firmwareType == FW_REPRAPFW)
   {
     infoMachineSettings.softwareEndstops = ENABLED;
-
+    infoMachineSettings.leveling         = BL_ABL;
```
kisslorand, line 253: Do all Reprap FW printers have ABL???

PR #2552, opened 2022-07-06
CZ translation

radek8:
Editing CZ translation

### DIFF
--- a/TFT/src/User/API/Language/language_cz.h
+++ b/TFT/src/User/API/Language/language_cz.h
@@ -335,7 +335,7 @@
     #define STRING_MPC_TITLE              "MPC autokalibrace"
     #define STRING_PID                    "PID"
     #define STRING_PID_TITLE              "PID autokalibrace"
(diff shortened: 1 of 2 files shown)


TFT/src/User/API/Language/language_cz.h
```diff
@@ -335,7 +335,7 @@
     #define STRING_MPC_TITLE              "MPC autokalibrace"
     #define STRING_PID                    "PID"
     #define STRING_PID_TITLE              "PID autokalibrace"
-    #define STRING_TUNE_START_INFO        "Autokalibrace bude spuštěna. Bude to chvíli trvat.\nPokračovat?"
+    #define STRING_TUNE_START_INFO        "Spustit kalibraci vytápění? Bude to chvíli trvat.\nPokračovat?"
```
kisslorand, line 338: Double question? To which one is the user responding? What if the user doesn't want to run heater calibration but he wants to continue? :)

radek8: Yes you're right. They are two questions about the same thing, but it could be misunderstood. I will fix it.

PR #2554, opened 2022-07-12
Code reduction

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR is a minor code reduction . 
It restores some code reduction that was overwritten by PR #2505 (in "Listitem.c" , "setDynamicValue()") and restores the numpad precision to 5 which was also overwritten by PR #2505 to 4. The listing of parameters allows precision of 5 but the numpad was set back to a limit of 4.

### Benefits

Everyone

### Related Issues

This PR is not related to any reported issue.

### DIFF
--- a/TFT/src/User/API/Temperature.c
+++ b/TFT/src/User/API/Temperature.c
@@ -120,7 +120,7 @@ void heatCoolDown(void)
 // Is heating waiting to heat up
 bool heatGetIsWaiting(uint8_t index)
 {
-  return (heater.T[index].waiting != WAIT_NONE);
+  return (heater.T[index].waiting == true);
 }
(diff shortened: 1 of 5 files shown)


bigtreetech:
Hello, https://marlinfw.org/docs/gcode/M109.html
<html>
<body>


[R<temp>] | Target temperature (wait for cooling or heating).
-- | --



</body>
</html>
<html>
<body>


[S<temp>] | Target temperature (wait only when heating)
-- | --



</body>
</html>

This PR implements both the ‘S’ parameter and ‘R’ parameter of M109 according to the logic of 'R' parameter.

This change does not seem appropriate

PR #2570, opened 2022-08-07
Use Marlin 2.1.1 or Newer

thisiskeithb:
### Description

Use Marlin 2.1.1+ since 2.1 had some gnarly SD bugs in it.

### Benefits

Improved readme.

### Related Issues

None.

### DIFF
--- a/Copy to SD Card root directory to update/config.ini
+++ b/Copy to SD Card root directory to update/config.ini
@@ -29,14 +29,14 @@
 # Supported Marlin Firmware Versions
 #--------------------------------------------------------------------
 #
-# Minimum Marlin firmware version: 2.1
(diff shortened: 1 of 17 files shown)


digant73:
It should be updated also `config.ini` file in both `Copy to SD Card root directory to update` and `TFT\src\User` folders

thisiskeithb:
> It should be updated also config.ini file

Will do.

> in both Copy to SD Card root directory to update

No need to mess with that folder since the CI test will build & copy the config (and bins) over: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/7c816bec99f181868b66ec679b9eefcd2d2130af/.github/workflows/buildBinary.yml#L120-L135

PR #2577, opened 2022-08-16
Minor bugfixes + cleanup on configuration files

### LINKED ISSUE #2544: BTT UPS
Unsure exactly what this section does.. The first option is totaly confusing.. It ism enabled by default. Seems it should be disabled by default ?#### Power Loss Recovery Mode
# Enable power loss recovery.
# Disable to reduce the loss of TFT SD card or TFT USB disk.
#   Options: [disable: 0, enable: 1]
pl_recovery:1

#### Power Loss Recovery Homing
# Home before power loss recovery.
#   Options: [disable: 0, enable: 1]
pl_recovery_home:0

#### Power Loss Recovery Z Raise
# Raise Z axis on resume (on power loss with UPS).
#   Unit: [distance in mm]
#   Value range: [min: 0.0, max: 2000.0]
pl_recovery_z_raise:10.0

#### BTT Mini UPS Support
# Enable backup power/UPS to move Z axis on power loss.
#   Options: [disable: 0, enable: 1]
btt_mini_ups:0

### LINKED ISSUE #2545: Chamber temperature sensing causes bed temperature display malfunction
### Description

I am using the Octopus v1.1 board with the TFT35v3 display.  I have a thermistor connected and configured to the `TEMP_2_PIN` to monitor the chamber temperature, but no chamber heater (`HEATER_CHAMBER_PIN` is undefined).  This is just to monitor/display the chamber temperature even though it can't be controlled.

The chamber (and bed) temperatures display and function correctly in OctoPrint on a connected RPi, but the TFT35v3 display fails to properly alternate the bed and chamber temperatures.  It will sometimes stop displaying bed temperature at all (either actual or target), sometimes will show the chamber icon instead of the bed icon at times when it should be showing the bed icon, and will sometimes just show `0/0` for the temperature where it should show either the chamber temperature or the bed temperature, and generally misbehaves.

I am suspecting it's related to the fact that I have Marlin configured to read/report the Chamber temperature but not control it, since there is no chamber heater.

### Steps to reproduce

1. Configure Marlin for Chamber temperature sensor (am using a simple 10K thermistor since it's low temp compared to hotend):
```
#define TEMP_SENSOR_CHAMBER 4
```
2. Configure Marlin temperature sensor pin (leave `HEATER_CHAMBER_PIN` undefined):
```
#if TEMP_SENSOR_CHAMBER
  #define TEMP_CHAMBER_PIN		TEMP_2_PIN
  //#define HEATER_CHAMBER_PIN      P2_04   // Required heater on/off pin (example: SKR 1.4 Turbo HE1 plug)
  //#define HEATER_CHAMBER_INVERTING false
  //#define FAN1_PIN                   -1   // Remove the fan signal on pin P2_04 (example: SKR 1.4 Turbo HE1 plug)
```
3. I have tried with `HEATER_CHAMBER` enabled and disabled in the TFT35 config.  It fails either way with erratic behavior, but does it more often and quicker when enabled.  Try it both ways:
```
#define HEATED_CHAMBER 1  // Default: 0
```

**Expected behavior**
I expect it to correctly display the chamber temperature as my OctoPrint does.  And expect it to correctly show the bed temperature, both actual and target.  I would even be OK with it just not showing the chamber temperature at all, since I can see it via OctoPrint (which is why I've tried configuring the TFT35 both ways), but I need it to show the bed temperature correctly.  However, it fails either way.

**Actual behavior**
It gets into a mode where it stops showing the bed heater temperature (either current or target) and displays the chamber heater icon when it should be showing the bed heater icon.  When it gets like this, you can't see the current bed temperature anymore from the display (either on the information screen or at the temperature bar at the top on other menus).  Sometimes switching from touchscreen mode to Marlin mode and then back will reset it and let it show the bed temperature again, but it doesn't always work.  A hard reset will make it work again until it gets back out of sync and into this strange mode.

### Hardware Variant
TFT35v3

TFT Firmware is at 1e5955e76d764e830ca707508052275d31f2e95e

Marlin Firmware is at https://github.com/MarlinFirmware/Marlin/commit/8aca38351cd6d33a1202bf9a13890dc8e5eadd49

### Additional Information
* Configuration.zip is attached with both TFT35 and Marlin config [Configuration.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/8933939/Configuration.zip)

* At present, I don't have a picture available, since I haven't had a camera available at the time it occurred.  I will try to find a camera and get a picture the next time it happens.

digant73:
Some minor bugfixes + cleanup on configuration files.

fixes #2544
fixes #2545

### DIFF
--- a/TFT/src/User/API/parseACK.c
+++ b/TFT/src/User/API/parseACK.c
@@ -1223,7 +1223,7 @@ void parseACK(void)
       {
         infoMachineSettings.buildPercent = ack_value();
       }
-      else if (ack_seen("Cap:CHAMBER_TEMPERATURE:"))
+      else if (ack_seen("Cap:CHAMBER_TEMPERATURE:") && infoSettings.chamber_en == DISABLED)  // auto-detect only if set to disabled
       {
         infoSettings.chamber_en = ack_value();
       }
--- a/Copy to SD Card root directory to update/config.ini
+++ b/Copy to SD Card root directory to update/config.ini
@@ -29,14 +29,14 @@
 # Supported Marlin Firmware Versions
 #--------------------------------------------------------------------
 #
-# Minimum Marlin firmware version: 2.1
-# Distribution date: 2022-06-06
+# Minimum Marlin firmware version: 2.1.1
+# Distribution date: 2022-08-06
 # Firmware source: https://github.com/MarlinFirmware/Marlin/releases
 #
 # To use all the features and functionalities supported by the TFT,
 # the following options must be enabled in Marlin firmware.
 #
-# General options:
+# General options which MUST be always activated:
 #   EEPROM_SETTINGS (in Configuration.h)
 #   BABYSTEPPING (in Configuration_adv.h)
 #   AUTO_REPORT_TEMPERATURES (in Configuration_adv.h)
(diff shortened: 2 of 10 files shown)


TFT/src/User/API/parseACK.c
```diff
@@ -1223,7 +1223,7 @@ void parseACK(void)
       {
         infoMachineSettings.buildPercent = ack_value();
       }
-      else if (ack_seen("Cap:CHAMBER_TEMPERATURE:"))
+      else if (ack_seen("Cap:CHAMBER_TEMPERATURE:") && infoSettings.chamber_en == DISABLED)  // auto-detect only if set to disabled
```
kisslorand, line 1226: This is kind of counterintuitive. In English "disable" means to turn something that was functional into an inoperative state. 
I speak only for myself, but I surely do not want something that I disabled to be functional.

PR #2585, opened 2022-08-24
parseAck() speed increase

kisslorand:
### Requirements

MKS or BTT TFT

### Description

This PR increases the speed of parsing the messages from the mainboard.
The increase of speed is due to:
- the rewriting the of the functions that are searching a character or string in the message received from the mainboard
- replacing general searching functions with situation specific ones (ones that are faster)
- applying in some cases a methodic search (search first for a common leading substring and only if found continue for the remaining)

The functions "ack_seen()", "ack_continue_seen()" and "ack_starts_with()" (formerly "ack_cmp()") are heavily used in "parseACK()". They were rewritten for much higher speed thus they help parsing the messages received from the motherboard faster.
I made this speed increase 8 months ago and extended performance increase measurements were done at that time. Given the amount of time gone by I do not recall the exact data but roughly there's a speed increase between 3% and 30+%. And this is only from those 3 functions. The speed increase is the lowest at searching one single character that is at the beginning of the searched string and it increases as the number of characters to be searched increases and the further they are in the searched string.

   The rest of the optimizations only add to these numbers. One example would be the parsing of the capabilities of the printer. Before it searched each and every cycle for the existence of 18 capabilities. These information were very-very rare present yet the lookup for them was present even during printing. Now there's only one lookup of the presence of any of them (by the leading substring of "Cap:") and if evidence is found that there's a capability message  present, only than a lookup for the specific capability is done.

### Benefits

Everyone

### Related Issues

None


digant73:
@kisslorand ok for using ack_continue_seen in parseAck. But the changes you made in the parsing functions (in both config.c and parseAck.c) are slower (e.g. you used strlen) and they can produce a buffer overflow (that was the reason for so much freezing experienced by users in the past). We already discussed that with BTT and BTT also made performance test resulting in slower performance.

kisslorand:
@digant 
Introducing of "strlen" was a last-minute addition and I was already about to remove it.
Other than that there's no other bottleneck, the performance increase is real. I spent back than 2 whole days making tests and measurements. The time needed to execute those 3 functions is less than their original counterpart. As far as my logic goes, less time needed for execution means faster execution. Correct me if I am wrong...

P.S.: I think I know what BTT performance test are you talking about, Those test were using predefined search functions (strcmp, strstr, etc.) and indeed they perform worse. This time it's not the case anymore.

digant73:
a brief check from my side:
- key_seen() in config.c should be ok (no buffer overflow and more compact)
- ack_starts_with in parseAck.c is slower than the current one (strlen() present)
- ack_seen() in parseAck.c should be slower than the current one (as it is accessing `str[++i]` in the loop while the current code access a simple index variable)
- ack_continue_seen() in parseAck.c as for ack_seen(). But here the new check `if (max_index < ack_index)` (instead of current `if (max_len < 0)`) could speed up performance

I would apply the check `if (max_index < ack_index)` on the current code for ack_continue_seen() and verify if that is enough to provide an overall better result than the new parsing functions.

kisslorand:
> * ack_starts_with in parseAck.c is slower than the current one (strlen() present)

As stated before it is a last-minute addition that is about to be removed.



> * ack_seen() in parseAck.c should be slower than the current one (as it is accessing `str[++i]` in the loop while the current code accesses a simple index variable)

Actually, in our context,
```
   if (str[++i] == '\0')
   {
      ...
   }
```

is faster than 
```
   i++;
   if (str[i] == '\0')
   {
      ...
   }
```

This is not in theory, it was measured by me. There is `i++` in the original code's "for" cycle.

I repeat it again. I tested and measured many methods, theories. I had several variants, I even had ones that were even faster but I've chosen a balance between performance and size. When I say "faster" it means it was faster in practice, measured, not theoretically.

digant73:
I was referring that `if (i == str_len)`was replaced by `if (str[i] == '\0')` (not even considered `i++` that was moved from the `for` to `if`) that is slower. By the way, it's up to you to provide the code being sure you made proper testing and evaluation. As far as I see, I would make the code a lot more linear (with no performance degradation) avoiding the use of ack_continue_seen / ack_starts_with inside each `if ack_starts_with(...)` in the main `if then else` sequence because they are making the code format completely fixed. E.g. for M201, only `M201 Xxx Yyy Zxx` is parsed while `M201 Yyy Xxx Zzz` is not parsed properly although it is correct. Also there is no advantage to complicate the code for all the param belonging to M503 section because they are not used during a print but only when the system is on idle. It is enough to use outer ack_starts_with just tp speed up the skip case.

TFT/src/User/API/parseACK.c
```diff
@@ -538,789 +531,790 @@ void parseACK(void)
...
     #endif
   }
 }
 
 #ifdef SERIAL_PORT_2
 
-void parseRcvGcode(void)
-{
-  for (SERIAL_PORT_INDEX i = PORT_2; i < SERIAL_PORT_COUNT; i++)  // scan all the supplementary serial ports
+  void parseRcvGcode(void)
```
digant73, line 1301: for the functions in .c no indentation is used with `#ifdef`. Only for inline code or data structures / function definition, indentation is used. If it was depending on me, no indentation was used at all. Something similar to Marlin was used.

kisslorand: Good to know, thank you!

kisslorand:
> I was referring that `if (i == str_len)`was replaced by `if (str[i] == '\0')` (not even considered `i++` because it was moved from the `for` to `if`) that is slower.

"The proof is in the pudding."

> By the way, it's up to you to provide the code being sure you made proper testing and evaluation.

I would happily share my testing method after I see a shared method that shows that those 3 functions are slower than their original counterpart. Than a discussion can be made to be sure who made proper testing and evaluation. As far as the "evaluation" goes I do not see it as rocket science to compare two numbers and see which one is smaller. I am confident that I am pretty good at evaluating which one of two given numbers is the smaller one.

> As far as I see, there are now limitation on format parsing. I would make the code a lot more linear (with no performance degradation) avoiding the use of that kind of `ack_continue_seen` / `ack_starts_with` combination inside each `if ack_starts_with(...)` in the main `if then else` sequence because they are making the code format completely fixed (the code must be more flexible to support RRF, Marlin etc...). E.g. for M201, only `M201 Xxx Yyy Zxx` is parsed while `M201 Yyy Xxx Zzz` is not parsed properly although it is correct (even for Marlin). Also there is no advantage to complicate the code for all the param belonging to M503 section because M503 is not issued during a print but only when the system is on idle. The outer `ack_starts_with` is enough to speed up the skip case (e.g. M201). Inside each param block (e.g. M201) I would use ack_seen (code easy and more flexible as it is now)

I agree with this, the impact of rigorous checking order is minimal and better be prepared for the unknown (deviation from standards). A commit for this has been already submitted.

Issue #2493, opened 2022-04-09
[BUG] TMC driver settings bugs, bump sensitivity and current for Dual X/Y systems


rondlh:
### Description

Setting the TMC driver bump sensitivity and stepper current has some issues.

**---Setting bump sensitivity---**
Set X to 1, X2 to 2, Y to 3, Y2 to 4 results in:
X = 2, X2 = 1, Y = 4, Y2 = 3. (Values for X and X2 are swapped, same for Y and Y2)

### Steps to reproduce
1. Set the bump sensitivity for X and X1, or Y and Y1 to different values via the TFT
2. Leave the settings menu
3. Go back into the bump sensitivity menu to see the actually set values.

**Expected behavior**
Values as previously define.

**Actual behavior**
Set value for X and X1 are swapped (same for Y and Y1)

**---Stepper current---**
Setting the current for X will also change the current for X2, the same is true for Y.

### Steps to reproduce
1. Set the TMC driver current for X to a different value than X2 (or use Y and Y2)
2. Leave the settings menu
3. Go back into the driver current menu to check the set value.

**Expected behavior**
X as set, X2 unchanged

**Actual behavior**
X2 is changed to the set value for X. X set correctly.

### Hardware Variant
BTT TFT35 V3.0
MKS MONSTER8 V1.0

### TFT Firmware Version & Main Board Firmware details
Latest TFT FW, April 2nd 2022
Latest Marlin 2.0.9.3 release March 27th 2022

digant73:
It seems a bug in Marlin.
If you use pronterface and send M906 Z1052 and then M503 you should see Z2 was also set to 1052.

```
echo:  M906 X1050 Y1150 Z1052
echo:  M906 I1 Z1052

```

rondlh:
> It seems a bug in Marlin. If you use pronterface and send M906 Z1052 and then M503 you should see Z2 was also set to 1052.
> echo:  M906 X1050 Y1150 Z1052
> echo:  M906 I1 Z1052
@digant73
Yes, you are right, but the TFT should send "M906 I0 Z1052" when there are more drivers on an axis, then it works as expected.
Did you manage to confirm the bump sensitivity issue?

digant73:
No, because I only have X and Y for bump sensitivity. However, you can add `I0` for X, Y etc.. in MachineParameters.c in the following array:

[...]

PR #2586, opened 2022-08-25
TMC driver settings bugs, bump sensitivity and current for Dual X/Y s…

digant73:
**BUGFIXES:**
* As reported in #2493 by @rondlh, since Marlin 2.0.9.3 the status of the TMC commands in relation to the usage of the I-command is the following:
  - **M913** (Set TMC Hybrid Threshold Speed) and
**M914** (Set TMC stallguard sensitivity)
The meaning of the I-parameter is as follows:
No I-parameter or I0 means all drivers (X, X2)
I1 means driver 1 (X, Y, Z)
I2 means driver 2(X2, Y2, Z2)
  - **M906** (set driver current) and
**M569** (Set spreadcycle or stealthchop)
The meaning of the I-parameter is as follows:
No I-parameter or I-1 means all drivers (X, X2)
I0 means driver 1 (X, Y, Z)
I1 means driver 2 (X, Y, Z)

  This PR is fixing indexing for M906 (the only command that was no more compliant with current Marlin implementation (checked Marlin 2.1.1 code).
  Furthermore, as already reported by @rondlh, Marlin documentation for both M913 and M914 is incorrect. I reported the bug here https://github.com/MarlinFirmware/Marlin/issues/24678.

* Minor code reduction for M301/M304 handling in interfaceCmd.c and parseAck.c.
* Minor cleanup.

### DIFF
--- a/TFT/src/User/Menu/Pid.c
+++ b/TFT/src/User/Menu/Pid.c
@@ -9,7 +9,7 @@ const char *const pidCmdRRF[] = PID_CMD_RRF;
 static int16_t pidHeaterTarget[MAX_HEATER_PID_COUNT] = {0};
 static uint8_t curTool_index = NOZZLE0;
 static uint8_t degreeSteps_index = 1;
-uint32_t pidTimeout = 0;
+static uint32_t pidTimeout = 0;
 static PID_STATUS pidStatus = PID_IDLE;
 
 // called by parseAck() to notify PID process status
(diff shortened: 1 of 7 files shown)


TFT/src/User/Menu/Pid.c
```diff
@@ -197,14 +198,14 @@ void menuPid(void)
 
             menuDrawItem(&pidItems.items[key_num], key_num);
             temperatureReDraw(curTool_index, &pidHeaterTarget[curTool_index], false);
-            break;
+          }
```
kisslorand, line 201: Good catch! Thank you!

PR #2593, opened 2022-08-31
Popup macros

kisslorand:
- macros with parameter descriptions for all popup types (dialog, reminder, splash)
- tiny code reduction


rondlh:
It would have been nice to also correct this typo in print.c:

      // confirm file selction

It should be "selection" of course.

kisslorand:
@rondlh 
Fixed in PR #2603. Thx!

PR #2601, opened 2022-09-14
Fix smart filament runout selection

### LINKED ISSUE #2600: filament runout sensor settings UI
Hi, on previous firmware versions I could use the UI to setup the filament runout sensor. There where three options: OFF, ON, SMART.
Now there is only the option ON or OFF. SMART seems to be missing.
Is this a feature and did I just miss this being announced or is this an error?

There is an updated way of setting up the sensor type in the config file but i used to use the UI to set it up.

kisslorand:
### Requirements

BTT or MKS TFT

### Description

- In "Settings -> Feature" was not possible to select SMART filament runout sensor, this PR fixes it.
- Bitwise macros tweak

### Benefits

In "Settings -> Feature" it is now possible to select SMART filament runout sensor

### Related Issues

Fixes #2600

### DIFF
--- a/TFT/src/User/Menu/FeatureSettings.c
+++ b/TFT/src/User/Menu/FeatureSettings.c
@@ -16,7 +16,7 @@ const LABEL itemToggleAuto[ITEM_TOGGLE_AUTO_NUM] =
 #define ITEM_TOGGLE_SMART_NUM 2
 const LABEL itemToggleSmart[ITEM_TOGGLE_SMART_NUM] =
 {
-  LABEL_ON,
+  LABEL_NORMAL,
   LABEL_SMART
 };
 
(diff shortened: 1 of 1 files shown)


digant73:
it was made by @guruathwal to fix a reported bug. Once in config.ini you select the basic runout sensor (smart or not smart), the TFT allows to switch ON/OFF that selected runout sensor

guruathwal:
It does not make any sense to undo a fix which is working ok. The sensor type is supposed to be selected in config files and the only option in the menu should be to turn it on or off.

kisslorand:
@digant73, @guruathwal 
Thanks for the warnings, issues fixed.

digant73:
#2600 simply asked for a clarification. You started a PR that was introducing a bug properly fixed by Gurmeet and then you converted it to something not requested/needed by anyone. The runout sensor type is an hardware setting as many others configurable (following a common approach) only in `config.ini`. It is set only one time and it never needs to be changed unless you upgrade the printer.
The remaining space in the FLASH could be used to make configurable other not HW types of settings requested in the past by some users such as the shim of the paper used in leveling menus (Z offset, manual leveling etc...). Those menus could be used frequently by some users according to environment temp and/or HW stresses and the tool (e.g. papers with different thickness) available for the leveling.

PR #2662, opened 2023-01-01
Safer alternatives for "strncpy"

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR is about "strncpy()" function.
There's an endless controversy about its usage, most folks advise against the use of it, it's not safe and it can lead to very hard to detect bugs. Its major flaw is that it doesn't null-terminate the destination string so it's the programmer's duty to always make sure the resulting string is always null-terminated. It's easier to say than do it.
Let's say it is used in some function/procedure/subroutine and at that point it is safe to use that function, no danger of having a string not being null-terminated. Being an open source, some other programmer might use that function/subroutine in a new 
feature he's developing and not checking every existing function/subroutine he's using along and use the said function with parameters that triggers the resulting string not being null-terminated. Another scenario could be the changing of a parameter/variable size due to some new functions/features that need that parameter being bigger in size, a parameter that is used somewhere along the code in that function/subroutine that uses "strncpy" and because of the enlarged size it will result in a non null-terminated string.
   I think it is obvious why a non null-terminated string is screaming for trouble so I won't stress on explaining it.
Even the best of us can slip it over, just one example would be the "statusScreen_setMsg()" function which when called from "parseACK()" than data from a variable which can have 512 bytes of data (dmaL2Cache) is copied to a variable (msgBody) with the maximum size of 70 bytes. No measures are taken that the eventual truncated result is not null-terminated. It's that easy to fall into the trap of the function "strncpy()".

This PR introduces 3 alternatives for "strncpy()":

__1. strxcpy__(char * destination, const char * source, size_t num):
- this one  always copy __num-1__ characters from __source__ to __destination__ regardless of null terminating character found in source
- destination __always__ ends with '\0'

__2. strwcpy__(char * destination, const char * source, size_t width):
- this one copies the contents of the __source__ to __destination__ but no more than __width-1__ characters
- if null terminating character is found in the __source__ than the rest in the __destination__ is padded with 0
- destination __always__ ends with '\0'

__3. strscpy__(char * destination, const char * source, size_t size):
 - this one copies the contents of the __source__ to __destination__ but no more than __size-1__ characters
 - if null terminating character is found in the __source__ than the copy stops there
 - destination always ends with '\0'

These 3 versions do not return anything, I saw it pointless to inherit the return a pointer to the __destination__ since it is something that is already known at the time of calling these functions.

### Benefits

- safer alternative for "strncpy()" by always resulting a null-terminated string
- lift the weight from the programmer's shoulder to always make sure to have a null-terminated string
- reduces compiled flash size


TFT/src/User/API/Vfs/vfs.c
```diff
@@ -272,19 +272,16 @@ bool getPrintTitle(char * buf, uint8_t len)
 {
   // example: "SD:/test/cap2.gcode" -> "SD:cap2.gcode"
 
-  char * strPtr = getPrintFilename();
-
-  // "+ 2": space for terminating null character and the flag for filename extension check
-  if (strlen(getFS()) + strlen(strPtr) + 2 > len)
+  // '\0': terminating null character
+  //  "+ 1": space for flag for filename extension check
+  if (snprintf(buf, len, "%s%s%c", getFS(), getPrintFilename(), '\0') + 1 > len)
```
digant73, line 277: current code = safe
proposed code = unsafe

kisslorand: In a very-very-very unrealistic specific scenario the extension wouldn't be hidden:
- most of the characters in the filename being unicode (more than 1 byte)
- the filename before extension less than (LCD_WIDTH - BYTE_WIDTH) characters (with the above characteristics)
- total number of bytes exceeds MAX_TITLE_LEN and there's a partial extension (ex.: .gcod)

I wouldn't call it "unsafe" but everyone is entitled to an opinion...
Anyway, even that scenario is covered now.

kisslorand: Outdated

TFT/src/User/Menu/Popup.c
```diff
@@ -124,27 +124,14 @@ void menuDialog(void)
   }
 }
 
-void popup_strcpy(uint8_t *dst, uint8_t *src, uint16_t size)
```
bigtreetech, line 127: I'm not sure if `memcpy` is safe when `const char *src` is `NULL`, so I will refactor it and submit commit for review later

bigtreetech:
[strscpy](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2662/commits/5bf70c8d023fcc9963a848cf2068c4b7a4192ce2) refer to the implementation of the linux kernel, and the parameters and return values are consistent with the general api

bigtreetech:
@kisslorand I see you mark it as draft, do you have any new ideas?

kisslorand:
I see problems with the new approach but Iam very busy at the moment. I marked it as draft to prevent merging it. Will come back with more details later today. 

Sent from Yahoo Mail on Android 
 
  On Fri, Feb 10, 2023 at 11:00, ***@***.***> wrote:   


@kisslorand I see you mark it as draft, do you have any new ideas?

bigtreetech:
Got it

kisslorand:
> @kisslorand I see you mark it as draft, do you have any new ideas?

I do not have any new ideas but observations on latest commit (https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2662/commits/5bf70c8d023fcc9963a848cf2068c4b7a4192ce2).
The new implementation of "strscpy" is the same as I did but with some more bells and whistles, additions that are not needed. The original "strncpy" doesn't check neither if *source* and _destination_ are NULL pointers or not, this is not needed in this project, all the cases where _strncpy_ is used originally uses initialized "strings". My implementation also handled just well if "size" parameter was zero. Current implementation increases flash size and doesn't fit this project in all the places it is uses it. For example it breaks the hiding of extensions of the G-code files in the file list. If extensions are selected to be hidden, they are still shown and if you select a file to be printed than a filename with a garbage at its tail is shown. Of course no print starts as that filename doesn't exist.


Another thing to mention is that the returning of the number of characters copied is not used anywhere in this project where _strncpy_ is used.

The new _strscpy_ is out of the scope of this PR so if you rather go with the new one than I would like to close this PR as this new _strscpy_ is not what this PR was meant for, a safe, more lightweight, drop-in alternative to _strncpy_ with some additional variants for future use-cases.

kisslorand:
@bigtreetech 
Reverted "use strscpy" commit (https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/pull/2662/commits/6781bfd58c98e82e0fd2cc74913da4c0a9616178), rebased and fixed a bug in NULL-terminating the string in `strscpy()'.

PR #2687, opened 2023-01-23
EXTENDED_CAPABILITIES_REPORT added to Readme

kisslorand:
Updated "Readme.md" with another general must have to be enabled option in Marlin.


radek8:
So can you add it to both config.ini files as well?

kisslorand:
Sure, thx for the tip.

radek8:
Shouldn't you edit the other one too?
BIGTREETECH-TouchScreenFirmware\Copy to SD Card root directory to update\config.ini

kisslorand:
As far as I remember it is done automatically by a Python script, it is copied to the "Copy to SD Card root directory to update" folder.

radek8:
Ok

PR #2701, opened 2023-02-05
Fix comment parsing speed regression

kisslorand:
### Requirements

BTT or MKS TFT

##
PR #2671 promised to "Optimize comment parsing in comment.c", instead it crowded the MCU with unnecessary instructions, all these during printing, where speed is crucial. The opposite of optimize was done, there is no benefit from it, only loss. The resulting code is not even smaller with one single byte.
Because of PR #2671 the first word in every comment from the G-code file is preprocessed even if it is not even a candidate for a known keyword. This takes time and it is unnecessary, not to mention again that this is during printing.

This PR brings back the systematic approach that was done before this "optimization" and also does what PR #2671 assumably wanted to do, reduces the footprint of "comment.c" on the FW's size.

##
An alternative for `strlwr()' and 'strupr()' is introduced. The introduced functions are lighter and are returning the pointer to the processed string. 
##

### Benefits

- faster overall processing of G-code comments

### Related Issues

 - luckily none reported yet

### DIFF
--- a/TFT/src/User/API/comment.c
+++ b/TFT/src/User/API/comment.c
@@ -5,7 +5,7 @@
 #define KEYWORD_INTIALS "lLrRtT"  // the initials of the searched keywords, add new initials for new keywords
 
 char gCodeCommentLine[COMMENT_MAX_CHAR] = {0};
-bool slicerTimePresence = false;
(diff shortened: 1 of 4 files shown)


kisslorand:
Bump

PR #2708, opened 2023-02-11
Completion of the missing key in the CZ and FR translation

radek8:
Completion of the missing key in the CZ translation.
Value translation.


digant73:
please do the same also on language_fr.h.
All the other languages are ok.

radek8:
OK

PR #2713, opened 2023-02-13
Leveling points bugfix

kisslorand:
### Requirements

BTT or MKS TFT

#### Bugs fixed:
 - if home offset in non-zero, the leveling points for the  "Leveling corner" and "Leveling" menus are determined wrong, they are asymmetrical and not respecting the leveling edge parameter (off by the values of the home offset)
 - if home offset is non-zero than the minimal edge level value is calculated wrong resulting skipped leveling points measurements
 -  just by entering the "L corner" menu the steppers would never auto-disable after the predefined inactivity timeout

#### Enhancements:
 - exiting from the "Leveling" menu the nozzle is raised to a predefined value (Config.ini: "level_z_raise") rather than leaving it very close to the bed

### Benefits

- usability of the aforementioned menus for those who have non-zero home offset
- safer nozzle height when exiting from "Leveling" menu
- smaller compiled flash size

### Related Issues

- none reported

### DIFF
--- a/TFT/src/User/API/LevelingControl.c
+++ b/TFT/src/User/API/LevelingControl.c
@@ -10,12 +10,18 @@ typedef struct XY_coord
 LEVELING_POINT probedPoint = LEVEL_NO_POINT;  // last probed point or LEVEL_NO_POINT in case of no new updates
 float probedZ = 0.0f;                         // last Z offset measured by probe
 
+int16_t setCoordValue(AXIS axis, ALIGN_POSITION align)
+{
+  return ((align == LEFT || align == BOTTOM) ? infoSettings.machine_size_min[axis] + infoSettings.level_edge
+                                             : infoSettings.machine_size_max[axis] - infoSettings.level_edge) - infoParameters.HomeOffset[axis];
+}
+
 void levelingGetPointCoords(LEVELING_POINT_COORDS coords)
 {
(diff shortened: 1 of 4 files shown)


TFT/src/User/API/LevelingControl.c
```diff
@@ -10,25 +10,31 @@ typedef struct XY_coord
...
+  int16_t x_right = setCoordValue(X_AXIS, RIGHT);
+  int16_t y_bottom = setCoordValue(Y_AXIS, LEFT);
+  int16_t y_top = setCoordValue(Y_AXIS, RIGHT);
 
   if (GET_BIT(infoSettings.inverted_axis, X_AXIS))
   { // swap left and right
-    int16_t temp = x_left;
-    x_left = x_right;
-    x_right = temp;
+    x_left = x_left ^ x_right;
```
bigtreetech, line 28: Have you disable "Allow edits and access to secrets by maintainers" ?
The fancy usage here is meaningless and only increases the difficulty of understanding. Please restore it

kisslorand: Sure, sorry for that.

rumkoff:
After start install firmware 18.04.2023 on BTT TFT35 E3 V3.0.1 i reed this mesage "illegal user app!"

PR #2714, opened 2023-02-13
Size reduction by function calls reduction

kisslorand:
### Requirements

BTT or MKS TFT

### Description

Almost 500 bytes of flash size reduction by removing unnecessary usage of `getParameter()` and `setParameter()` functions.

### Benefits

- lower compiled FW size

### Related Issues

- none


guruathwal:
Not a good idea to access the variable from one API directly from another API.

kisslorand:
This was a bad idea (for different reason than mentioned above), so closing this. Thx for the warning thou.

PR #2718, opened 2023-02-15
Minor cleanup

digant73:
Minor cleanup on top of last merged PRs:
* added/reviewed comments in parseAck.c
* in both interfaceCmd. and parseAck.c any check and set/get of a machine parameter through getParameter() or setParameter() function is provided on single line (compact but also good readability form)
* fixed indentation in some source files
* used common comment style across a whole file (e.g. lower case if the majority of comments were in lower case in the file).
* GUI fucntions (e.g. popup etc.) moved after code logic when possible (e.g. after status variables are set etc...) and separated by an empty line
* some cleanup on my_misc.c/h (function names, function order etc...)

**PR STATE:** Ready for merge


TFT/src/User/Menu/FeatureSettings.c
```diff
@@ -47,7 +47,7 @@ typedef enum
   SKEY_START_GCODE_ENABLED,
   SKEY_END_GCODE_ENABLED,
   SKEY_CANCEL_GCODE_ENABLED,
-  SKEY_RESET_SETTINGS,        // Keep reset always at the bottom of the settings menu list.
```
guruathwal, line 50: after looking at this PR, I find it hard to understand the logic for when to make the first letter of the sentence Capital or small. 🤣

TFT/src/User/API/menu.c
```diff
@@ -1255,55 +1254,63 @@ void loopBackEnd(void)
...
   // Speed & flow monitor
   loopSpeed();
 
+  // Buzzer handling
   #ifdef BUZZER_PIN
-    // Buzzer handling
     loopBuzzer();
   #endif
 
+  // Handle a print from (remote) onboard media, if any
```
guruathwal, line 1269: Here the first alphabet is changed to capital.

TFT/src/User/Menu/ABL.c
```diff
@@ -66,8 +67,7 @@ void ablStart(void)
 
     case BL_UBL:  // if Unified Bed Leveling
       storeCmd("G29 P1\n");
-      // Run this multiple times since it only fills some missing points, not all.
-      storeCmd("G29 P3\n");
+      storeCmd("G29 P3\n");  // run this multiple times since it only fills some missing points, not all
```
guruathwal, line 70: Here the fist alphabet is changed to small?

TFT/src/User/Menu/FeatureSettings.c
digant73: personally, I always prefer to provide inline comments in lower case and without "." at the end of a sentence. I Use first capital letter and "." when the there are more sentences (e.g. "this is first sentence. This is a second sentence"). I also try to use a common style in a whole file. So if the majorithy of comments were lower case, I use that, Most of the time the comments are lower case, so you see that more often

TFT/src/User/Menu/ABL.c
digant73: See my first comment. Just used same form in the whole file

TFT/src/User/API/menu.c
digant73: as the other existing comments, so I didn't change them. See my first comment

PR #2732, opened 2023-02-23
Added MKS TFT32L V3.0 and fixed horizontal mirror display.

stef-ladefense:
### Requirements

MKS TFT32L V3.0

### Description

For this display, the SS and GS bits must be identical. Not having another MKS 28 or 32 display, I added its management in addition to not break the existing one. For information, the MKS 32 series shares the firmware of the MKS 28 series. the firmware must be named "MKSTFT28" to be taken into account by the boot loader, so naming its folder TFT32 is counterproductive and is therefore named "TFT28 " like the firmware.

### Benefits

Added MKS TFT32L V3.0 and fixed horizontal mirror display.

### Related Issues

fixes horizontal mirror display.


thisiskeithb:
You'll want to also add this to both of the github workflows for this repo so code is compiled/tested & committed for this new TFT: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/tree/master/.github/workflows

Note: The firmware binaries in the "Copy to SD Card root directory to update" folder is automatically built & committed through the previously mentioned workflows, so you don't need to add them to PRs.

PR #2768, opened 2023-04-11
Bug fixes on top of last merged PRs

digant73:
**BUG FIXES:**
* Fixed instability bug introduced by #2662: despite the safety code claimed by #2662, the changes applied by that PR (use of strxcpy() function) made the code unsafe. strxcpy is not applicable at all in VFS API where the source buffer's size is always less that the destination buffer's size. strxcpy is not safe in general, meaning an out of buffer read occurs when the source buffer's size is less than the destination buffer's size (as in VFS API). Also strwcpy() and strscpy() were not working (freeze or wrong displayed files navigating on  sd card/usb stick). Proper functions named strncpy_pad()/strncpy_no_pad() are now used
* Fixed instability bug provided by setPrintTitle() function (in Printing menu): the hideExtension()/restoreExtension() functions made visible on VFS API by #2662 are not applicable on a standard string buffer. Those functions are specific to manage the data defined in the API (they have been hided also for that reason). Previous code is restored
* No new mesh applied by MeshEdit menu: with the changes applied by #2737 the meshSetValue() function is now setting the mesh for the original mesh instead of the mesh value provided by MeshTuner menu. It means that none of the meshes set by MeshTuner menu is applied at all. Furthermore, pressing on save button the original data grid was restored instead of setting the current data grid as the new original data grid
* Fixed unmatching function declaration/definition in .h/.c introduced by #2654: Fixed babystepSetValue() signatures
* Other bug fixes: e.g. wrong check on BedLeveling menu introduced by #2766

**NOTE:** #2766 removed the CaseLightControl API properly created (as any other API) to decouple the GUI from the core functions (e.g. the that API was used in interfaceCmd API). Now, a return to the past (3 years ago) has been applied

**PR STATE:** ready for merge

### DIFF
--- a/TFT/src/User/API/BabystepControl.c
+++ b/TFT/src/User/API/BabystepControl.c
@@ -7,7 +7,7 @@ static float babystep_value = BABYSTEP_DEFAULT_VALUE;
 #define BABYSTEP_CMD_SMW "G43.2 Z%.2f\n"
 
 // Set current babystep value
-void babystepSetValue(const float value)
+void babystepSetValue(float value)
 {
   babystep_value = value;
 }
--- a/TFT/src/User/API/HomeOffsetControl.c
+++ b/TFT/src/User/API/HomeOffsetControl.c
@@ -56,10 +56,11 @@ float homeOffsetGetValue(void)
 // Reset Z offset value to default value
 float homeOffsetResetValue(void)
 {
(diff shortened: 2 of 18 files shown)


rondlh:
Merged and tested OK, seems like an important fix!

PR #2783, opened 2023-04-24
MB Reset Fix

### LINKED ISSUE #2779: [BUG] Issuing a reset to Marlin causes issues for the TFT
If you issue a Marlin reset command ("KILL" externally or via terminal) then the TFT seems to get disconnected, also the set value like Flow, speed, fan speed etc. stay at the previously set values.

I added the following in parseACK.c so the TFT can survive the motherboard reset:

```
else if (ack_seen("start\n")) { // IRON, ADDED TO SURVIVE MOTHERBOARD RESTART
		infoHost.connected = false;
	    infoHost.wait = false;

```
Still the set values are not reset to their original values.

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR fixes the TFT not communicating with the motherboard after it has ben reset (software reset, external reset, watchdog reset, brown out reset)

### Related Issues

Fixes #2779

### Credit

All credit goes to @rondlh, he found the problem and the solution for it, I just expanded it for more possible situations.

### Notes

~Marked as draft as it is waiting for user feedback~

### DIFF
--- a/TFT/src/User/API/FanControl.c
+++ b/TFT/src/User/API/FanControl.c
@@ -13,8 +13,15 @@ static uint8_t needSetFanSpeed = 0;
 static bool ctrlFanQueryWait = false;
 static uint32_t nextCtrlFanTime = 0;
 
+void fanResetSpeed(void)
+{
+  needSetFanSpeed = 0;
+  memset(setFanSpeed, 0, sizeof(setFanSpeed));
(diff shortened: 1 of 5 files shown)


rondlh:
Tested, works well in all cases, this is an excellent improvement of the user friendliness of the TFT

PR #2798, opened 2023-05-26
Power loss recovery - bugfix

kisslorand:
### Requirements

BTT or MKS TFT

### Description

- PR #1959 introduced a bug in the power loss print restart procedure. Because of this bug there's a high chance of missed/skipped initial commands for a PLR print.
- PR #2419 introduced a bug, after a power failure during print if a new PLR backup file cannot be created than the print will restore from the beginning, not from the backed up position 
- PR #2596 introduced a bug, if PLR is enabled than start gcode defined in config.ini is not executed
- if X, Y, Z are relative coordinates than upon a PLR the E is also set to relative regardless if it was relative or absolute
- the number of cached commands for PLR can exceed the maximum number available in the cache leading to TFT freeze

This PR fixes these bugs.

### Benefits

PLR

### Related Issues

There were reports about PLR hiccups.

### DIFF
--- a/TFT/src/User/API/PowerFailed.c
+++ b/TFT/src/User/API/PowerFailed.c
@@ -97,38 +97,34 @@ bool powerFailedInitData(void)
     if (infoBreakPoint.pause)
       z_raised += infoSettings.pause_z_raise;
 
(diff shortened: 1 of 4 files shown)


kisslorand:
Bump

kisslorand:
Dear stale bot, this is a bugfix. Maybe some day it will find its way into the master branch...

kisslorand:
Closing it due to being obsolete.

kisslorand:
Closing it due to being obsolete.

PR #2799, opened 2023-05-29
Further print abort improvement

kisslorand:
### Requirements

BTT or MKS TFT

### Description

 - This PR is a follow-up of PR #2728. That PR came with a nice idea but implemented it poorly. In case of printing from TFT media or remote host not all possible blocking operations are released (just only one) so the print will not stop ASAP.
This PR fixes it, in case of a print abort all blocking operations are released so the abortion of the print happens without any hesitation.

 - FW size is shrunken.


kisslorand:
Bump

PR #2807, opened 2023-06-19
MPC method follow-up

kisslorand:
### Requirements

 - MKS or BTT TFT
 - Marlin with MPC enabled

### Description

PR #2791 implemented the choice of MPC algorithm method. Selecting the preferred MPC method makes the whole screen flicker. This PR fixes it.

There's also a minuscule, tiny, itsy-bitsy, 8 byte FW size reduction due to removing an `if` conditional by relocating a "busy processing" splash screen function call in the MPC module.

### Benefits

No flickering screen selecting the preferred MPC method.

### Related Issues

None reported, own finding.


kisslorand:
Bump!

PR #2812, opened 2023-07-01
Remove SERIAL_FLOAT_PRECISION Requirement

thisiskeithb:
### Description

Marlin will not report a temperature or position with more than two digits beyond the decimal place since it uses `_MIN(SERIAL_FLOAT_PRECISION, 2)` for temperature reporting and 2 will always be less than 4. Other values printed via serial are now hard-coded to use one or two digits.

### Benefits

Removes unnecessary Marlin config change.

### Related Issues

None. Discovered while working through other serial issues in Marlin.

### DIFF
--- a/Copy to SD Card root directory to update/config.ini
+++ b/Copy to SD Card root directory to update/config.ini
@@ -55,7 +55,6 @@
 #
 # Options to support dialog with host:
 #   EMERGENCY_PARSER (in Configuration_adv.h)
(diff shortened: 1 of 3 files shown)


thisiskeithb:
This may get fixed upstream, so I'm going to close this for now:
- https://github.com/MarlinFirmware/Marlin/issues/26171

PR #2824, opened 2023-08-20
Added Marlin ADVANCED_OK + static ADVANCED_OK support

### LINKED ISSUE #2761: [Q] Random pauses or hesitation when printing.
While printing once or twice a print, sometimes more, the machine will stop right in the middle of printing for a second, maybe even two. No errors or anything, it leaves an ugly blob, and it just continues on. It does not happen the same place on the file or on the machine. If I had to say anything, it might happen more often during circle moves, or curves. This does not happen while using the same screens for my CNC machines.

I have 6 printers  and 2 CNC's running TFT35 V3 E3 screens. I have noticed this issue for probably as long as I have used the screens, more than two years I would guess. Some screens are brand new and some are very old. I update marlin and the TFT several times a year. Some of my gcode is old and ran perfectly fine on older 1286 screens, and some of my files are new.

So I think this is either some sort of configuration issue, slicer setting, or a screen firmware bug. Any ideas?

Config, 
[config.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/11165582/config.zip)

### LINKED ISSUE #2806: [FR] (feature request title) help is needed
I created an issue - no answer
https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/2727

I tried to create a support ticket through the site - the system does not work

I wrote an email to
service001@biqu3d.com,
service004@biqu3d.com,
service005@biqu3d.com,
support01@bigtree-tech.com
no answer

the problem https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/2727 is still relevant

image

### LINKED ISSUE #2810: [CLOSED] (TFT70 V3.0 & SKR 3 EZ issue where only marlin mode works)
I bought them separate from each other, after I couldn't get my Ender 3 V2 screen to work with the SKR 3 EZ

using marlin bugfix 2.1 for STM32H723VG_btt & the newest firmware from github for the TFT70 (GD) variant
I have all 3 ribbon cables plugged in correctly and also checked the pinout for the 232 cable, now the touch screen mode shows its connected and displays temperatures from thermistors but I can't move axis or make any changes while in TFT mode, only if I switch to marlin mode will it work (make temperature changes & move axis)

I've asked quite a few people to help including the Marlin Discord, shared my configuration.h & h.adv as well as the config.ini for my screen. Others have said it all looks good, so what am I doing wrong?

### LINKED ISSUE #2816: [Q] (Manual moving speed to Z axis ?)
Hi !
What speed command is sent to the 3D printer's controller when the Z-axis is manually controlled? 
I have an MKS TFT35 v1.0 and it sends an F3000 speed command to my 3D printer, when I manually control the z-axis movement. My printer almost jumps at that speed.
If I burn "BIGTREETECH-TouchScreenFirmware" into my MKS TFT what is the movement speed of the Z-axis with manual movement

**specify the device variant related to your question.**
MKS TFT35 v1.0

**Additional context**
Add any other context or screenshots about the question here.

### LINKED ISSUE #2819: [FR] ADVANCED_OK support, base code implemented already
Marlin supports ADVANCED_OK, which returns info about the available buffers when acknowledging a command with "ok".
This is currently not used by the TFT, which sends 1 command and waits for the "ok" response before sending the next command.
I suspect that this is the reason for some random pauses and hesitations when printing #2761

Marlin support BUFFER_MONITORING, which can give you information about the buffer underruns, a buffer underrun occurs when Marlin is ready to process the next command, but no command is available yet. I found **thousands** of buffer underruns per hour. These underruns are likely to be of very short duration, but clearly that is not what you want. Some buffer underruns are normal and unavoidable because the TFT is holding back data during the heating phases, this accounts for about 100 underruns in my case.

I made a rudimentary support of ADVANCED_OK which works as follows:

1. When you have ADVANCED_OK enabled in Marlin, then Marlin responds to commands like this "ok P31 B15". The B-parameter indicates the amount of available command buffers.
In parseACK.c the "ok" acknowledgement is processed. I added a check for the "B" parameter. The parameter value is stored in the HOST struct (defined in main.h). 

```
#define BUFSIZE 4
// "ok" response handling
if (ack_starts_with("ok"))  {
  infoHost.wait = false;
  infoHost.AvailableBuffers = MIN(infoHost.AvailableBuffers + 1, BUFSIZE);
  // if regular "ok\n" response or ADVANCED_OK response (Marlin) (e.g. "ok N10 P15 B3\n")
  if (ack_cache[ack_index] == '\n')
    goto parse_end;  // there's nothing else to check for  

    if (ack_continue_seen("P")) 
      if (ack_continue_seen("B")) { // ADVANCED_OK response
        infoHost.AvailableBuffers = MAX(ack_value(), 1);
        goto parse_end;  // there's nothing else to check for
      }
}
```

AvailableBuffers is initialized to 4 (Marlin's default) when the printer connection is established. (parseACK.c)

```
  infoHost.connected = true;
  infoHost.AvailableBuffers = BUFSIZE; // Marlin's default
  requestCommandInfo.inJson = false;

```

2. In interfaceCmd.c I changed the 2 sections with the sendCmd. For each send I deduct the buffer counter, and pause if no more buffers are available.

```
if ((infoHost.AvailableBuffers > 0) || !infoHost.connected)
  if (sendCmd(false, avoid_terminal) == true)  // if the command was sent
  {        
    infoHost.AvailableBuffers--;
    if (infoHost.AvailableBuffers > 0) {
         infoHost.wait = false;
         Delay_ms(2); // Allow a little bit of time for Marlin to process the commands
      }
     else
       infoHost.wait = infoHost.connected;
  }
```

Now there are no more additional buffer underruns during the printing process and at least basic operations seem to work as before. (Move menu works, terminal works, adjusting values during printing works).

I think the current TouchScreen FW can even be improved significantly without using ADVANCED_OK by just assuming that at least 4 command buffers are available (can be made configurable), and keep track of the response count ("ok"). So you can send 4 commands, but then need to wait for an "ok". If an "ok" comes then another command can be send. This should solve the pauses and hesitations.

Update: I did some tests to see if this can work without having ADVANCED_OK enabled. It's possible, but a bit complicated by the fact the command and response count could be undefined at startup. Some commands could be send that are never acknowledged by Marlin.

UPDATE: Source Code files attached (based on the (currently latest) June 2023 release), just overwrite the 3 attached files. It will work both with ADVANCED_OK enabled and disabled in Marlin. If it's disabled then the default amount of 4 buffers is assumed. Make sure to increase RX_BUFFER_SIZE (Configuration_Adv.h) to at least 256 bytes, but 512 is recommended.
Updated Download: #2824

digant73:
**IMPROVEMENTS:**
- Added `advanced_ok` setting on `config.ini` file to allow `Marlin ADVANCED_OK` and `static (TFT based) ADVANCED_OK` support: Based on original code shared by @rondlh on #2819. Code reviewed, fixed and improved

  If enabled:
  - if "ADVANCED_OK" feature is enabled in Configuration_adv.h in Marlin firmware, the TFT will use
    the available G-code TX slots indication provided by the mainboard to schedule the transmission
    of multiple G-codes, if any, for a maximum of the given indication.
  - if "ADVANCED_OK" feature is disabled in Configuration_adv.h in Marlin firmware, the TFT will
    support the transmission of G-codes according to the configured `tx_slots` setting.

  If disabled, the TFT will provide the standard transmission logic based on one G-code per time.

- Added `tx_slots` setting on `config.ini` file: Maximum number of G-code TX slots used by the TFT for the communication with the printer used when static ADVANCED_OK is applied
- Added `Monitoring` menu: Available in `Notification` menu (pressing on `Info` button) only if `DEBUG_MONITORING` option is enabled in `Configuration.h`. It allows to monitor/show resources usage (e.g. buffered gcodes, pending gcodes, available TX slots etc...) useful for troubleshooting (see attached pictures)
- Minor code optimization on `os_timer.c`: As reported in FR #2832, useless invokation of updatePrintTime() function was performed while it could be limited just to 1 per second (as needed during a printing)
- Minor improvement in `_GUI_DispLenString()` function (`GUI` API) when displaying a truncated message
- Minor code cleanup/optimization

**NOTES:**
- Backward compatibility with current gcode transmission logic based on one command per time is supported/allowed simply by setting `advanced_ok` to 0 (disabled)
- ADVANCED_OK feature made configurable also at runtime on `Feature` menu
- Good print quality results for static ADVANCED_OK seem to be reached even with `tx_slots` set to 2 (no frequent idle state on mainboard's RX buffer

**BUGFIXES:**
- Bugfix on print abort loop: Function Serial_Available provided by last merged PR #2827 was not enough to always properly handle a print abort loop. It was possible to flood M108 causing buffer overruns on Marlin side and related errors displayed on TFT as popup error messages
- No TFT connection established at TFT boot when a SMART runout sensor is enabled: With a SMART runout sensor enabled, the TFT was not able to connect at boot (it was necessary to use the disconnect button on `Connection` menu to properly connect the TFT to mainboard). The TFT sent a `M114 E` gcode for which mainboard didn't provide any reply. Bug is more on mainboard than on TFT but it is fixed on TFT side simply delaying the transmission of M114 E after the TFT properly connected to mainboard

resolves #2819
resolves #2806 (zombie)
resolves #2810 (zombie)
resolves #2816 (zombie)

**PR STATE:** Ready for merge

Notifications_1

other_7


rondlh:
Here a small tweak for the notification list (Gui.c)
If you like it please feel free to add it to any PR you like.

It checks if the truncate indicator "…" is actually needed, it's not needed if:
1. the last character is already displayed.
2. the only missing character fits into the available space reserved for the truncate indicator.

**Let me show what it does by giving some examples.**
"Date:2023-09-18 18:21:34" (The complete notification message fits into the available space)
"Date:2023-09-18 18:21:34…" (OLD Listview menu displays this, message is complete but truncate indicator is added)
"Date:2023-09-18 18:21:34 " (NEW, message is complete, no need for truncate indicator)

"Date:2023-09-18 18:21:34A"  (The complete notification message just fits into the available space)
"Date:2023-09-18 18:21:34…"  (OLD Listview menu displays this, truncates message)
"Date:2023-09-18 18:21:34A" (NEW, use the truncate space for the last character)

"Date:2023-09-18 18:21:34AB" (The complete notification message doesn't fit)
"Date:2023-09-18 18:21:34…" (OLD Listview menu displays this, CORRECT)
"Date:2023-09-18 18:21:34…" (No change)

The 2 changed lines have comments.


```
const uint8_t* _GUI_DispLenString(int16_t x, int16_t y, const uint8_t *p, uint16_t pixelWidth, bool truncate)
{
  if (p == NULL) return NULL;

  CHAR_INFO info;
  uint16_t curPixelWidth = 0;

  if (truncate) pixelWidth -= BYTE_HEIGHT;

  while (curPixelWidth < pixelWidth && *p)
  {
    getCharacterInfo(p, &info);
    if (curPixelWidth + info.pixelWidth > pixelWidth)
    {
      if ((truncate) && *(p + 1)) // check if there is at least 1 more character
      {
        if (*(p + 2) || info.pixelWidth > BYTE_HEIGHT) // truncate indicator if 2 more characters OR not enough space for current character in reserved space
          getCharacterInfo((uint8_t *)"…", &info);
        GUI_DispOne(x, y, &info);
      }

      return p;
    }
    GUI_DispOne(x, y, &info);
    x += info.pixelWidth;
    curPixelWidth += info.pixelWidth;
    p += info.bytes;
  }
  return p;
}
```

rondlh:
I noticed that you made the whole DMA_CIRCULAR_BUFFER volatile, but I believe that it's only required for the members that are actually changed in the ISR, so wIndex only.
It's not efficient for `cacheSize` and `cache` for example, they could even be defined constant to help the compiler.
 Any reason to do this?
```
typedef volatile struct  // precautionally declared as volatile due to access from interrupt handler and main thread
{
  char *cache;
  uint16_t wIndex;  // writing index
  uint16_t rIndex;  // reading index
  uint16_t flag;    // custom flag (for custom usage by the application)
  uint16_t cacheSize;
} DMA_CIRCULAR_BUFFER;
```

digant73:
> I noticed that you made the whole DMA_CIRCULAR_BUFFER volatile, but I believe that it's only required for the members that are actually changed in the ISR, so wIndex only. It's not efficient for `cacheSize` and `cache` for example, they could even be defined constant to help the compiler. Any reason to do this?
> 
> ```
> typedef volatile struct  // precautionally declared as volatile due to access from interrupt handler and main thread
> {
>   char *cache;
>   uint16_t wIndex;  // writing index
>   uint16_t rIndex;  // reading index
>   uint16_t flag;    // custom flag (for custom usage by the application)
>   uint16_t cacheSize;
> } DMA_CIRCULAR_BUFFER;
> ```

yes, volatile could be applied only to wIndex. I would provide this change in the PR related to #2835. `const` cannot be applied to `cache` or `cacheSize` (warning and compile error)

digant73:
> if (*(p + 2) || info.pixelWidth > BYTE_HEIGHT) // truncate indicator if 2 more characters OR not enough space for current character in reserved space
>           getCharacterInfo((uint8_t *)"…", &info);
>         GUI_DispOne(x, y, &info);

shouldn't it be:

```
if (*(p + 2) || info.pixelWidth > BYTE_HEIGHT) // truncate indicator if 2 more characters OR not enough space for current character in reserved space
{
  getCharacterInfo((uint8_t *)"…", &info);
  GUI_DispOne(x, y, &info);
}
```
?

rondlh:
> Ok, minor improvement added

THANKS! I know it's a minor thing, but my standard startup messages (date and time, ip address, ETA etc) have this issue, so it annoyed me a bit.

rondlh:
@digant73 
I found a potential issue... it could be me, but it seems that the BTT FW only sends "M105\n" once, and when it doesn't get a response then you are out of luck. (I can see exactly what arrives at the host)
I have the latest Marlin (about 3 weeks old) + #2824 up-to-date.

Perhaps it's the Marlin bug you mentioned?
I simulate this by resetting the motherboard, and then 2 seconds later I reset the TFT.
Buffered gcodes and pending gcodes = 1. tx_slots = 0.
When I again reset the TFT, then they connect fine, which makes me think that it's not an Marlin issue, but a TFT issue.

digant73:
> @digant73 I found a potential issue... it could be me, but it seems that the BTT FW only sends "M105\n" once, and when it doesn't get a response then you are out of luck. (I can see exactly what arrives at the host) I have the latest Marlin (about 3 weeks old) + #2824 up-to-date.
> 
> Perhaps it's the Marlin bug you mentioned? I simulate this by resetting the motherboard, and then 2 seconds later I reset the TFT. Buffered gcodes and pending gcodes = 1. tx_slots = 0. The TFT Disconnect solves the issue like you mentioned, because another "M105" is sent. When I reset the TFT again, then they connect fine, which makes me think that it's not an Marlin issue, but a TFT issue.
> 
> UPDATE: When going from the current master to #2824, then I always need to use "reset.txt" to get things going. So I update all icons/fonts/ini etc, then get a black screen and get stuck. Using reset.txt gets things going again.

strange. Try to disable ADVANCED_OK in Marlin and verify if the issue is still present

rondlh:
Confirmed, fixed, if not connected then M105 keep coming, well done!
I guess the issues with the SMART runout sensor were also related to this issue:
"Fixed no TFT connection issue with SMART runout sensor enabled".

Let me check if my ESP3D issue (early ESP3D message stops the connection) is fixed now too.

PR #2825, opened 2023-08-20
Corner Leveling Marlin Fix

### LINKED ISSUE #2815: [BUG] Probing corners using UBL "L Corner"
### Description

Probing corners using UBL "L Corner" 

### Steps to reproduce

1. Update TFT35 Firmware with bin.
2. Go to L corners via Movement -> Bed Level -> UBL -> L Corner.
3. Select any corner
4. Bed gets probed
5. No result showed on screen

**Expected behavior**
Bed height should be displayed

**Actual behavior**
Nothing is changing on the screen

### Hardware Variant
BTT TFT35 E3 V3.0
[BIGTREE_TFT35_V3.0_E3.27.x.bin](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/blob/master/Copy%20to%20SD%20Card%20root%20directory%20to%20update/BIGTREE_TFT35_V3.0_E3.27.x.bin) Marlin bugfix-2.1.x

### Additional Information
Config of printer https://pastebin.com/YdqK0FMG
Working video https://youtu.be/lQHiDFrEKaw

### LINKED ISSUE #2849: [BUG] Non-functional corner leveling using a probe
After four months I updated Marlin, and found I had the same problem as shown here:
https://github.com/MarlinFirmware/Marlin/issues/26171

The SERIAL_FLOAT_PRECISION option no longer works in Marlin.
The TFT display is no longer able to analyze the values from the leveling probe due to the format of the value sent by the Marlin.
When leveling and straightening corners with a probe, the measured values are not displayed.
Would someone look into this and fix the problem?
Thank you

### Hardware Variant
TFT 35 E3 V3.0
current TFT Firmware 
current version Marlin 2.1 bugfix

### LINKED ISSUE #2855: [FR] Home button in the corner leveling menu
I'm missing the Home button in the corner leveling menu.
If I move the screws, the center of the bed also moves, and I would like to execute the G28 command again.
If the M48 button were moved elsewhere, there would be a free position.
The M48 button is not important when leveling corners. 
The M48 button is already in the BL-Touch menu.

What do you think about replacing the M48 button with a Home button?

kisslorand:
### Requirements

BTT or MKS TFT

### Description

Some Marlin updates broke the corner leveling.
PR https://github.com/MarlinFirmware/Marlin/pull/24390 makes the TFT to freeze when corner levels are probed.
PR https://github.com/MarlinFirmware/Marlin/pull/25996 breaks the positioning of the probe to the right positions if home offset is used.

This PR fixes these issues making it backwards compatible. There's a FW size reduction by removing unnecessary code and optimizing the corners probe flow. 
Also there's some cosmetic addition too:
 - if corners are not probed than "---" is shown in the value section of each corner represented on the TFT
 - "M48" is replaced with a button for "Home Z" function
 - in manual leveling menu the button for disarming the X and Y steppers is removed (it had no practical usage there), the need for a submenu became redundant so "Distance" and "Shim" buttons are visible now right in the menu itself

### Benefits

Corner leveling usable with newest Marlin bugfix (as of 2023.VIII.21).

Fixes #2815
Fixes #2849
Fixes #2855

### Related Issues

#2815, #2849, #2855

### PR Status

Ready to merge but I keep my hopes low, posted here mostly for a reference in [my repository](https://github.com/kisslorand/BTT-TFT-FW).

### Notes

Leechers are welcome to include it in their own private build and share it through Thingiverse or whatever else.


radek8:
@kisslorand 
I tested your fix.
Unfortunately, it only works partially.
The front left and front right corners are not tested.
Only error information appears (Z out of range) and then only the right and left rear corners are tested.

kisslorand:
Do you use "home offset" in Marlin?

radek8:
I do not use "home offset"

radek8:
https://youtu.be/TKDgbg9kZME

radek8:
The problem is that in the configuration I have set the minimum value of the Y axis to a negative value.
min Y = - 35

If I set the corner offset to 45mm, the probing is done in the front, but in the wrong place (10mm from the edge of the bed).
This is also wrong. Negative values are the problem.

kisslorand:
Can you please provide the response of the printer to the "M115" command?
I am interested in the response that looks like:
"area:{full:{min:{x:0.000,y:0.000,z:0.000},max:{x:305.000,y:305.000,z:390.000}},work:{min:{x:0.000,y:0.000,z:0.000},max:{x:305.000,y:305.000,z:390.000}}}"

kisslorand:
Thank you. Until I finish working on fixing this please try the following:
Replace `else if (ack_seen("full:"))` in "parseACK.c" with `else if (ack_seen("work:"))`.
It should work on your case although it's not a universal fix suitable for all the possible printer configurations.

radek8:
The edit to the parseACK.c file you suggested solved my problem.
Thank you

radek8:
Now I see that the original value was "work:"
You changed the value to pomto PR and that broke the feature for people with a configuration like mine.
Why did you change it when it worked?
Revert this change to undo and the problem will be gone :-)

kisslorand:
I changed it because the value of "work" is inconsistent.
Make this test:
1. boot your printer
2. set a home offset with `M206' (for ex `M206 X13 Y14`)
3. send `M115' and remember the values for "work" area
4. home printer (`G28`)
5. send `M115` again and check the values for "work" area

This is just one test to show you what I am talking about, there are other scenarios when "work" area is inconsistent. It is always dependent if you homed or not the printer since the boot of the motherboard. Before homing "work" area is always 0 for minimum and the maximum is the bed size. After homing the minimum and maximum area corresponds to the soft endstops if they are enabled, otherwise remains the same.
There is no command to check if the printer was homed or not before. There are other problems too, I am working on make it suitable for all printer setups.
<br>

> You changed the value to pomto PR and that broke the feature for people with a configuration like mine.

Latest Marlin broke my setup if using "work" because I use home offset.

radek8:
> 
> BTW, I saw you use
> 
> ```
> #define X_MIN_POS -12
> #define Y_MIN_POS -35
> ```
> 
> According to Marlin it represents how much further the axis can travel after triggering the endstop sensor. Does your X and Y axis can travel further after triggering the endstop sensor?


No, activation of the limit switch is also the physical end of travel.
After activating the switch, it must return to the beginning of the bed to the zero point.

kisslorand:
> After activating the switch, it must return to the beginning of the bed to the zero point.

That's what the "home offset" is for.

If you issue a `G0 X117.5 Y117.5` command does your nozzle go exactly above the center of your bed?

radek8:
I don't need a home offset. There are problems with him.

Yes. My bed is defined as 230x230.
If I give the command G0 X115 Y115, the active nozzle will be set to the center of the bed.

radek8:
A negative value of X will allow the second nozzle to reach the entire bed, i.e. the left edge of the bed. (the second nozle one is placed approx. 12mm to the right of the first one)

A negative Y value will allow me to probe the entire bed. (the probe is cce 34mm behind the nozzles)

kisslorand:
OK, understood. Thanks for the explanations.

kisslorand:
@radek
I finished redoing this PR to accommodate more printer config scenarios while playing along with latest Marlin too. Please test it with your configuration.
Also please note that "config.ini" is changed, new data is required for this PR to work.
Thank you in advance.

radek8:
Thank you.. 
When I'm at the printer, I'll try your repair.

radek8:
Latest edits tested.
It works as expected.
Thank you

radek8:
Link also #2815
It's the same problem

kisslorand:
Thank you, I was just searching for it to be included.

radek8:
I have a question.
I'm missing the Home button in the corner leveling menu.
If I move the screws, the center of the bed also moves, and I would like to execute the G28 command again.
If the M48 button were moved elsewhere, there would be a free position.
The M48 button is not important when leveling corners. I think the M48 button is already in the BL-Touch menu.
Would you consider this change?

kisslorand:
I am contemplating for a long time about the need for that button there. I still have to figure out how to make it with double function, sometimes only `G28 Z` is enough but sometimes all axis need to be homed if the bed moves while user adjusts the bed's screws (some might have very low IDLE time for steppers disable).
Please make a feature request so I can refer to it. I'm glad I am not the only one who sees homing button better suited there and sees M48 pointless in that menu.

I am still not sure about `G28 Z` or rather `G28` would be a better choice there.

kisslorand:
PR updated with the implementation of the FR #2855.

radek8:
@bigtreetech 
Please merge. This is a bug fix.
Thank you

PR #2827, opened 2023-08-23
Fixed random pause/hesitation when printing

### LINKED ISSUE #2761: [Q] Random pauses or hesitation when printing.

digant73:
**BUGFIXES:**
- Fixed #2761: Based on deep analysis and testing shared by @rondlh on #2761
- Improved partial message retrieval handling: `Serial_Get` finction in `SerialConnection` API improved to prevent the possibility to read multiple times the same partial message on next function invokation

fixes #2761 

**PR STATE:** Ready for merge

### DIFF
--- a/TFT/src/User/API/Printing.c
+++ b/TFT/src/User/API/Printing.c
@@ -105,7 +105,7 @@ void loopBreakToCondition(CONDITION_CALLBACK condCallback)
   // from that command. Than another "M108" will be sent to unlock a next possible blocking command.
   // This way enough "M108" will be sent to unlock all possible blocking command(s) (ongoing or enqueued) but not too much and
   // not too fast one after another to overload/overrun the serial communication
-  TASK_LOOP_WHILE(condCallback(), if (infoHost.rx_ok[SERIAL_PORT] == true) sendEmergencyCmd("M108\n"))
+  TASK_LOOP_WHILE(condCallback(), if (Serial_Available(SERIAL_PORT) != 0) sendEmergencyCmd("M108\n"));
 
   // remove any enqueued command that could come from a supplementary serial port or TFT media
   // (if printing from remote host or TFT media) during the loop above
--- a/TFT/src/User/API/SerialConnection.c
+++ b/TFT/src/User/API/SerialConnection.c
@@ -112,7 +112,7 @@ void Serial_Forward(SERIAL_PORT_INDEX portIndex, const char * msg)
     Serial_Put(SERIAL_DEBUG_PORT, msg);
   #endif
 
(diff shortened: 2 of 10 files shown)


rondlh:
How to get this PR merged asap?

digant73:
It seems to be ok. Other features like advanced ok or any other improvement can be provided in a separate PR.

rondlh:
> It seems to be enough to solve the issue (just keeping in this PR the minimal code).`Serial_Get` (reader) and `USART_IRQHandler` (writer) are totally decoupled now. #2761 is open since April. Other features like advanced ok or any other improvement can be provided in a separate PR.

@digant73 Are you responding to me? If so then you seem to have misunderstood my question...

rondlh:
> I have also to work and travel during the day. If you are afraid about the possible issue due to no use of `volatile` for wIndex I just made an edit in my last comment in #2761. It will be added.

I mean this... many PRs have NOT been merged for a long time, perhaps BTT is not paying much attention, so the question is how do we get your recent work merged asap? I'm not worried about the volatile variable, it has been off during all of my testing.

digant73:
> > I have also to work and travel during the day. If you are afraid about the possible issue due to no use of `volatile` for wIndex I just made an edit in my last comment in #2761. It will be added.
> 
> I mean this... many PRs have NOT been merged for a long time, perhaps BTT is not paying much attention, so the question is how do we get your recent work merged asap? I'm not worried about the volatile variable, it has been off during all of my testing.

I can only suppose that some PRs are not merged for reasons that the author perfectly know. After that said, as far as I see BTT is OoO since more than 10 days. I already asked BTT to consider to merge many PRs if it can help.

V1EngineeringInc:
Niiiice!!

rondlh:
Merged 🥇  Great job, thanks for your support!

Issue #2832, opened 2023-09-07
[BUG] Buggy time handling code (doesn't consider timer rollover)


rondlh:
While looking through the TFT source code I noticed some issues, the time handling code does not handle timer rollovers correctly. This happens in many places in the code (just search for "OS_GetTimeMs"), luckily rollovers are not frequent (it takes 50 days), but this issue will probably cause a hiccup of the TFT, perhaps even a lockup.

**Here an example (common.c):**

```
// Check time elapsed against the time specified in milliseconds for displaying/updating info on screen
// Use this for timed screen updates in menu loops only
bool nextScreenUpdate(uint32_t duration)
{
  static uint32_t lastTime = 0;
  uint32_t curTime = OS_GetTimeMs();

  if (curTime > (lastTime + duration)) // BUG!!!
  {
    lastTime = curTime;
    return true;
  }
  else
  {
    return false;
  }
}
```

The problem here is that "lastTime + duration" does rollover before curTime does, causing the `if` statement guard to become true immediately, which will happen again and again until curTime also rolls over.

**The correct code would be this:**

`if (curTime - lastTime > duration)`

**This code is rollover safe.**

Another example (menu.c)
```
  reminder.time = OS_GetTimeMs() + STATUS_BAR_REFRESH_TIME; // BUG!!!

  if (menuType != MENU_TYPE_FULLSCREEN) drawReminderMsg();
}

void loopReminderManage(void)
{
  if (reminder.status == SYS_STATUS_IDLE || OS_GetTimeMs() < reminder.time) return; // BUG!!!
```
**This should be:**

```
  reminder.time = OS_GetTimeMs();

  if (menuType != MENU_TYPE_FULLSCREEN) drawReminderMsg();
}

void loopReminderManage(void)
{
  if (reminder.status == SYS_STATUS_IDLE || OS_GetTimeMs() - reminder.time < STATUS_BAR_REFRESH_TIME) return;
```

So this makes me think that timer rollovers were not considered, but then I came across this piece of code in timer.c

```
void TIM7_IRQHandler(void)
{
  if ((TIM7->SR & 0x01) != 0)
  { // update interrupt flag
    TIM7->SR &= (uint16_t)~(1<<0);  // clear interrupt flag

    os_counter++;

    updatePrintTime(os_counter);

    loopTouchScreen();

    if (os_counter == (uint32_t)(~0))
    {
      os_counter = 0;
    }
  }
}

```
[...]

PR #2833, opened 2023-09-10
Relax OS_Time IRQHandler

kisslorand:
### Requirements

BTT or MKS TFT.

### Description

This PR is an attempt to ease up the OS timer IRQ handler. Major credit goes to @rondlh who brought it to our attention here: https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/issues/2832. More than that @rondlh made several suggestions to make the IRQ handler lighter on the MCU, suggestions that are included in this PR.
The extra changes are:
 - remove modulo operation which is expensive  and replace it with bare simple arithmetic ones
 - ~remove the usage of any conditional and ternary from one of the functions called in that interrupt handler and replaced it with bare arithmetic operations~
 - moved the interrupt flag reset to the end of the operations to avoid any unpredictable behaviour of the IRQ handler

### Benefits

In theory it speeds up the TFT.

### DIFF
--- a/TFT/src/User/API/UI/TouchProcess.c
+++ b/TFT/src/User/API/UI/TouchProcess.c
@@ -26,8 +26,10 @@
 #define F TSC_Para[5]
 #define K TSC_Para[6]
 
+#define TOUCH_DEBOUNCE_MS 20  // 20ms
+
 int32_t TSC_Para[7];
-static volatile bool touchScreenIsPress = false;
+static volatile uint8_t touchCountdown = TOUCH_DEBOUNCE_MS;
 bool touchSound = true;
 
 void TS_Get_Coordinates(uint16_t *x, uint16_t *y)
@@ -142,30 +144,17 @@ uint16_t Key_value(uint8_t total_rect, const GUI_RECT *menuRect)
   return IDLE_TOUCH;
 }
 
-void loopTouchScreen(void)  // Handle in interrupt
+void checkTouchScreen(void)  // Handle in interrupt
 {
-  static uint8_t touch;
-  if (!XPT2046_Read_Pen())
-  {
-    if (touch >= 20)  // 20ms
-    {
-      touchScreenIsPress = true;
-    }
-    else
-    {
-      touch++;
-    }
-  }
+  if (XPT2046_Read_Pen() == LOW)
+    touchCountdown -= !!touchCountdown;
   else
-  {
-    touchScreenIsPress = false;
-    touch = 0;
-  }
+    touchCountdown = TOUCH_DEBOUNCE_MS;
 }
 
-uint8_t isPress(void)
+bool isPress(void)
 {
-  return touchScreenIsPress;
+  return (touchCountdown == 0);
 }
 
 void (*TSC_ReDrawIcon)(uint8_t position, uint8_t is_press) = NULL;
@@ -177,7 +166,7 @@ uint16_t KEY_GetValue(uint8_t total_rect, const GUI_RECT* menuRect)
 
   uint16_t key_return = IDLE_TOUCH;
 
-  if (touchScreenIsPress)
+  if (isPress())
   {
     if (firstPress)
     {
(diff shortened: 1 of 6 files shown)


rondlh:
Brilliant improvement for updatePrintTime, but why did you make void loopTouchScreen 2 to 3 times slower than before? Sabotage?

kisslorand:
Sorry, my sense of humor might be very different from yours, I really cannot decide it was meant as a joke or you are asking for real.

rondlh:
> Sorry, my sense of humor might be very different from yours, I really cannot decide it was meant as a joke or you are asking for real.

I'm asking for real, your implementation is 2-3 times slower than before, the sabotage part is obviously a joke.

kisslorand:
Can you pleased share the benchmark you used to measure the performance decrease?
All I see is performance increase there.

rondlh:
> Can you pleased share the benchmark you used to measure the performance decrease?
> All I see is performance increase there.

Show me... I theoretically can see your implementation is much slower

kisslorand:
Show what?
```
  touch = !XPT2046_Read_Pen() * (touch + !touchScreenIsPress);
  touchScreenIsPress = !(TOUCH_DEBOUNCE_MS - touch);
````
Are we talking about these 2 lines?

**Comment by rondlh** [comment]
Yes, I'm talking about loopTouchScreen.
You say you see performance increase, what did you see/find? What test? Any numbers?

**Comment by kisslorand** [comment]
How did you came to the conclusion it is 2-3 slower?

**Comment by rondlh** [comment]
> How did you came to the conclusion it is 2-3 slower?

Like I said, theoretically I can see that. But you said:

> All I see is performance increase there.

So the question is... how did you see performance improvement?

**Comment by kisslorand** [comment]
I have the same question, how do you see performance decrease?

**Comment by rondlh** [comment]
There is "an issue" in your code, similar to the issue in "ADVANCED_OK", that stops it from being efficient.

**Comment by kisslorand** [comment]
You said there's no need for benchmarks, but you just intrigued me. I will make some benchmarks and will come back with the results. It might take an hour or so.

**Comment by rondlh** [comment]
OK, take it easy

**Comment by kisslorand** [comment]
I am already on it. ;)

**Comment by kisslorand** [comment]
I have the results. :)
CodeBattle

**Comment by rondlh** [comment]
How did you test?
Can you show me the code?

**Comment by kisslorand** [comment]
Actually the two codes compiled do perform the same, the compiler does its job quite well. The difference above could be a rounding issue as the timer is counting microseconds. The previous test were done on a Cortex-M4 MCU, I have the results on a Cortex-M3 MCU also:

CodeBattle1

I'm aware that it is harder to understand the code I wrote, I have no issue undoing it and keeping the original code since the compiler knows how to optimize it.

**Comment by kisslorand** [comment]
> May I ask what nationality you are? (I'm Dutch)

There's no secret, I am Hungarian. You could already tell that since I made many additions/changes to the Hungarian language pack.

**Comment by rondlh** [comment]
My tests show that your loopTouchScreen is only 25% slower than the original code (STM32F207), not the 2-3 times I theoretically expected. Only code readability has decreased by a factor of 2-3.

**Comment by kisslorand** [comment]
What speed increase did your test show for your proposed code?

**Comment by rondlh** [comment]
> What speed increase did your test show for your proposed code?

A few percent only

**Comment by kisslorand** [comment]
Opinions are always welcome.

**Comment by rondlh** [comment]
Where is the opinion in this? I only see facts.

> * loopTouchScreen has become about 25% slower
This is a fact, even your own tests show that at best loopTouchScreen has not become faster, only more unreadable. That your code is slower was obvious to me immediately when I saw it, although I overestimated the speed decrease that it causes. Look again at the code... it's obvious! If you still don't see it then I can show it to you.

> * loopTouchScreen is completely unreadable and unmaintainable
Fact, you also agree to this above.

> * most of the changes are just copied from
Very likely, or perhaps you came up with the exact same idea a day after @digant73 did

> * TIM7_IRQHandler is a bit faster (still not optimal)
Fact, like I showed you above, you need to count down, not up to be faster. 

> * Some unused code was removed
Fact, but if you deny it I'm willing to take your side.

**Comment by rondlh** [comment]
Can somebody please review this code change force-pushed by @kisslorand?
Please consider execution speed and code readability. In my view it's obvious, but I still did a benchmark which confirmed my suspicions? This is a routine handling the touch screen presses of the TFT and is executed 1000x a second.

ORIGINAL CODE:
```
static volatile bool touchScreenIsPress = false;
void loopTouchScreen(void)  // Handle in interrupt
{
  static uint8_t touch;
  if (!XPT2046_Read_Pen())
  {
    if (touch >= 20)  // 20ms
    {
      touchScreenIsPress = true;
	  lastTouchTime = OS_GetTimeMs(); // IRON, ADDED TO RECORD TOUCH TIME
    }
    else
    {
      if (OS_GetTimeMs() - lastTouchTime > 250) // IRON, PREVENT TOO FAST TOUCHING
      touch++;
    }
  }
  else
  {
    touchScreenIsPress = false;
    touch = 0;
  }
}
```

@kisslorand VERSION: (comments removed that hold the original version, I wonder why those were left in the code)
```
static volatile bool touchScreenIsPress = false;
void loopTouchScreen(void)  // Handle in interrupt
{
  static uint8_t touch;
  touch = !XPT2046_Read_Pen() * (touch + !touchScreenIsPress);
  touchScreenIsPress = !(TOUCH_DEBOUNCE_MS - touch);
}
```

**Comment by kisslorand** [comment]
> @kisslorand has seen the light

Actually what I've seen is the rookie mistake I made benchmarking the original function and mine. In the benchmark I didn't use the variables from the functions benchmarked, the compiler optimized the code and eliminated the unused variables.
A lesson for everybody, you are not benchmarking the code you wrote, you are benchmarking the code the compiler generates!

After the realization of my mistake I redid the tests, the results are the following:

Benchmark

Where "My new code" is:
```
void loopTouchScreen(void)  // Handle in interrupt
{
  if (XPT2046_Read_Pen() == LOW)
    touchCountdown -= !!touchCountdown;
  else
    touchCountdown = TOUCH_DEBOUNCE_MS;
}
```
"My old code" is:
```
void loopTouchScreen(void)  // Handle in interrupt
{
  static uint8_t touch;
  touch = !XPT2046_Read_Pen() * (touch + !touchScreenIsPress);
  touchScreenIsPress = !(TOUCH_DEBOUNCE_MS - touch);
}
```

"Ori code" is:
```
void loopTouchScreen(void)  // Handle in interrupt
{
    static uint8_t touch;
    if (!XPT2046_Read_Pen())
    {
      if (touch >= 20)  // 20ms
      {
        touchScreenIsPress1 = true;
      }
      else
      {
        touch++;
      }
    }
    else
    {
      touchScreenIsPress1 = false;
      touch = 0;
    }
}
```

"Dumb code" is same as "Ori code" but variables not being used, it can be observed how the optimization of compiler the kicks in.

So @rondlh was right, my old code was 20% slower than the original code (not 25% as he stated, but who knows, maybe my tests are still inaccurate).

I have yet to verify the up/down count speed theory, although it is about the comparison to 0 (zero) and not the direction of the count. The specialized literature mentions that it was a thing in the 1980s but modern compilers and MCUs do not exhibit this behaviour but better test it, right?

Note: negative criticism is good as long as it is done in a civilized manner. Github also have this rule: "criticize the idea, not the person".

**Comment by kisslorand** [comment]
Up/Down count and comparison to zero results:

Count

Upcount: count from 0 to X
Downcount: count from X + 10 down to 10
CompToZero: count from X down to 0

The difference is clearly from comparison to 0 (zero), not from the count direction. The results of @rondlh  for down count being more faster than down count is probably because the tested a down count to 0 (zero). It is the comparison to 0 (zero) that makes down count faster. Anyhow I can only see a 3% difference, not 15%, but in any case it is a free trick for speed.

Everyday we learn something...

**Comment by kisslorand** [comment]
> The DMA serial writing I have under test has significantly more impact.

It's not a contest.

**Comment by rondlh** [comment]
What do you aim to achieve by moving `TIMER_INTF(TIMER6) &= (uint16_t)~(1<<0); to the back of the ISR?
Generally that's not the right thing to do, but I'm curios to hear your explanation.

**Comment by kisslorand** [comment]
In many microcontroller architectures, including those that use the STM32 series with the STM32 HAL library, clearing the interrupt flag at the end of the ISR is the recommended and standard practice (check TIM3, the buzzer interrupt). This is to ensure that the ISR completes its main tasks before allowing the processor to respond to new interrupts, minimizing interrupt latency and potential issues with nested interrupts.
In our case in reality it is irrelevant where's the interrupt reset done in the ISR for TIM7. I guess I did it out of habit.

**Comment by rondlh** [comment]
#2824 now contains a performance benchmark that you can enable in Configuration.h (DEBUG_MONITORING). When enabled, in the notification menu you will find another button to bring you to the monitoring screen. The scan rate will show you how many times the loopProcess is executed. This way you can test the impact of your improvement easily.

**Comment by kisslorand** [comment]
Thanks for the info, I did that.
The speed increase is at best 2% compared to the original code. I even tested modulo with 1024 (`os_timer & 0x03ff == 0`) which gave occasionally at best 3% speed increase. It's amazing how practice beats theory. :)
I think this PR can be entirely ignored given the new data.
Yes, the TIM7 is 10% faster than it was originally but in the big scheme of the entire FW it gives at best 2% speed increase.

**Comment by rondlh** [comment]
So you mean that this PR increases the scan rate by about 2%?

**Comment by kisslorand** [comment]
> So you mean that this PR increases the scan rate by about 2%?

Yes.

**Comment by rondlh** [comment]
What concrete numbers do you get? Before and after scan rates?

**Comment by kisslorand** [comment]
The checking was done on a BTT TFT35 V3.0 with STM32F207

The scan rate on average was:
1. master branch -> 29400~29600
2. this PR -> 29900~30100

**Comment by rondlh** [comment]
>>    So you mean that this PR increases the scan rate by about 2%?

>Yes.

> The checking was done on a BTT TFT35 V3.0 with STM32F207
> 
> The scan rate on average was:
> 
>     1. master branch -> 29400~29600
> 
>     2. this PR -> 29900~30100

I think you are doing something very wrong, nothing you state here makes any sense.
The speed improvement is a solid 0.0% (BTT TFT35 V3.0)
2% is complete non-sense, anybody can see that when looking at the tiny source code changes.

**Comment by kisslorand** [comment]
> I think you are doing something very wrong, nothing you state here makes any sense. 

It's the opinion of the same person who has the strong belief that counting down is faster than counting up. :)
<br>

> 2% is complete non-sense, anybody can see that when looking at the tiny source code changes.

A real professional argument brought to the table! :)
<br>

I guess it's true what they say, a circus is not a circus without a clown.

File: TFT/src/User/API/UI/TouchProcess.c [hunk 1]
```diff
@@ -144,28 +146,15 @@ uint16_t Key_value(uint8_t total_rect, const GUI_RECT *menuRect)
...
-    {
-      touchScreenIsPress = true;
-    }
-    else
-    {
-      touch++;
-    }
-  }
+  if (XPT2046_Read_Pen() == LOW)
+    touchCountdown -= !!touchCountdown;
```
bigtreetech, line 150: This method is not acceptable as it depends on the number of times this API is called and is limited by the entire task cycle of the large loop.
It is easy for the filtering time to be too short, not performing the filtering function it should have, or the filtering time to be too long, resulting in slow touch.

PR #2834, opened 2023-09-10
Move menu cosmetics

kisslorand:
### Requirements

BTT or MKS TFT

### Description

This PR brings small cosmetics to the Move menu.

Y+ and Y- icons were confusing as the icon associated to Y+ clearly shows the bed moving to back and the icon associated to Y- clearly shows a bed mowing forward. As most of printers have the zero position when the bed is fully on the back, moving the bed to the back the Y coordinate decreases and moving the bed to forward the Y coordinate increases.
In this PR the icons representation now correspond to the actual movement of the bed.

The coordinate values on the top of the display got a small attention, they are moved to the same height as reminders and menu names are, when a reminder is displayed, it will remain visible until it naturally goes away, than the coordinate values are visible again.

Before:

Move_2

Move_3

After:

Move_0

Move_1

Move_6

### DIFF
--- a/TFT/src/User/API/menu.c
+++ b/TFT/src/User/API/menu.c
@@ -667,6 +667,11 @@ void setMenu(MENU_TYPE menu_type, LABEL * title, uint16_t rectCount, const GUI_R
   #endif
 }
 
+SYS_STATUS getReminderStatus(void)
+{
+  return reminder.status;
+}
+
 void drawReminderMsg(void)
 {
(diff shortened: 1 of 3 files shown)


rondlh:
Does this mean that everyone will have to change the inverted_axis value for Y in config.ini?

kisslorand:
> Does this mean that everyone will have to change the inverted_axis value for Y in config.ini?

No.

Issue #2836, opened 2023-09-18
[FR] General TFT Speed Improvements


rondlh:
I want to create a central space for general speed improvement discussions, hopefully resulting in a PR.

To start I want to raise some topics that perhaps can be applied to all applicable areas of the TFT.
I can support the proposals with benchmarks to see if, and how much improvement they bring.
Of course the most frequently visited code is most interesting in this case (loopProcess!).
Let me start with raising some topics, please add yours below...

1. Calling a function, and then in the function check if the call is actually needed is quite inefficient, especially if you have to pass a lot of parameters to the function. It is much faster to first check if the call is actually needed, before calling the function. Applying this idea to Serial_Get can increase the loopProcess scan rate by about 20%.

2. 8, 16 bits variables vs 32 bit variables. All here supported MCUs are 32 bit, and generally using 32 bits variables is recommended even if an 8 bit variable is enough, this is because most of the assembler instructions are 32 bit, so 8 and 16 bit registers need to be "emulated", which means that actually a 32 bit variable is used, that simulates the behavior of a 8 or 16 variable. This of course causes overhead.
A quick test on a prototype version of `Serial_Get` showed that converting the seven 16-bit variables to 32 bit increased the speed by 9%. Not bad considering that it's such an easy task to replace a few variables declarations.

3. @kisslorand showed that the % operator (modulo) is quite an expensive operator. This is because this operator needs an integer division which is computational intensive. I'm aware of 2 solutions:
**3A.** @kisslorand showed that in some cases a counter can be used, which is faster
**3B.** If a value of 2 to the power of X can be used (2, 4, 8, 32...), then the module operator is not expensive anymore. In example 3A we needed to count to 1000, but if this could be converted to 1024 then the problem would be gone. This is useful for buffer memories!
**3C.** Also counting down to 0 if faster than the other way around, I found a speed difference of 15% when only considering the counting part. This is the case because there are specific assembler instruction for comparing to 0, but not for comparing to 1000. So for example:

`for (int a = 0; i < 1000; i++) { do your task 1000x times }` // counting up is slower than

`for (int a = 1000; i > 0; i--) { do your task 1000x times }` // counting down to 0

So for example if you want to make a string uppercase, then do it from the back to the front, not from the front to the back.

4. I found many cases in the code that look like this
`if (!x) { do A } else { do B }`
Faster, and easier to understand is: (UPDATE: This is incorrect, this could be faster or slower, see below why)
[...]

PR #2840, opened 2023-09-22
Added TX Interrupt/DMA based serial writing

### LINKED ISSUE #2835: [FR] Improved serial data processing (smarter reading, faster writing)
**1. Serial reading** 
The current FW uses DMA for reading data from the serial ports, which is done in a very efficient way, the process is started and will run automatically in the background. Interrupts are generated when data is available, this happens when the serial line goes idle. 
This is overall a very efficient approach, but has a small drawback. The received data only gets processed once the serial line goes idle, under most circumstances this is fine because the received messages are short and fragmented ("ok\n"). But if a long continues message is received, like the response to M43 (pins information), then a TFT buffer overrun can occur and a part of the received message is lost.
This issue can be solved by not waiting for the serial idle interrupt, but immediately starting to process the data that is being received. Based on the improvements in #2824 (improvements in Serial_Get) it is quite obvious how to implement this. I will provide some sample code soon.

**The advantages:**
* the hardware dependent interrupt code is not needed anymore
* buffer overruns become less likely, commands like M43 will work (better)

**2. Serial writing**
Practically the TFT only needs to receive a little bit of data ("ok\n" messages), but needs to write a lot of data to the motherboard (the gcode commands, about 40 characters per command). 
The current serial write implementation is very slow, the software actually needs to wait after each byte until it's physically send over the serial line. This slows down the TFT response time, especially for lower baud rates and when the workload is high anyway. A buffered DMA based would significantly help to increase the TFT responsiveness.

Currently I have a DMA based serial write solution under test, which works in this way:
* The message from Serial_Put is stored in a buffer memory and the program can continue immediately. If not enough buffer memory is available then the Serial_Put has to wait until enough space becomes available.
* Once a complete message is in the buffer the DMA process is started and the Serial transfer will occur in the background, which will trigger an interrupt when done, then more data can be transferred is needed.

**The advantages:**
 * overall improved speed and responsiveness of the TFT
 * high scan rates (Serial_Get) even under heavy load
 * buffer overruns become less likely, commands like M43 will work

A disadvantage is that this code is hardware dependent (DMA setup).
I will post the STM32F2_4 sample code here soon for review and discussion.

**I have some questions:**
1. Is it acceptable to provide this solution for one hardware platform only?
2. **Who can help test/implement a STM32F10x implementation?**
BTT TFT24 V1.1
BTT TFT28 V1.0
BTT TFT35 1.0/1.1/1.2/2.0
BTT GD TFT24 V1.1
BTT GD TFT35 V2.0
MKS TFT28 V3.0/4.0
MKS TFT28 NEW GENIUS
MKS TFT32 V1.3/1.4
MKS TFT32L V3

3. **Who can help test/implement a gd32f20x implementation?**
BTT GD TFT35 V3.0
BTT GD TFT35 E3 V3.0
BTT_GD TFT35 B1 V3.0
BTT GD TFT43 V3.0
BTT GD TFT43 V3.0
BTT GD TFT50 V3.0

### LINKED ISSUE #2858: [BUG] Error in parseACK.c when using a Delta Printer
### Description

I haven't used my printer in about 8 months now. Trying to build this now and I'm getting the following error for a TFT43_V3. Seems its a issue with the if blocks in parseACK.c, specifically for when `DELTA_PROBE_TYPE != 0`. Not too familiar with the coding, but that's the spot where the first error is poping up at.

```
TFT/src/User/API/parseACK.c: In function 'parseACK':
TFT/src/User/API/parseACK.c:939:9: error: expected '}' before 'else'
         else
         ^~~~
TFT/src/User/API/parseACK.c:369:10: warning: variable 'avoid_terminal' set but not used [-Wunused-but-set-variable]
     bool avoid_terminal = false;
          ^~~~~~~~~~~~~~
TFT/src/User/API/parseACK.c:949:5: error: 'else' without a previous 'if'
     else if (ack_starts_with("M200"))
     ^~~~
Compiling .pio/build/BIGTREE_TFT43_V3_0/src/src/User/Fatfs/ffsystem.o
TFT/src/User/API/parseACK.c:1398:10: error: 'avoid_terminal' undeclared (first use in this function); did you mean 'menuTerminal'?
     if (!avoid_terminal && MENU_IS(menuTerminal))
          ^~~~~~~~~~~~~~
          menuTerminal
TFT/src/User/API/parseACK.c:1398:10: note: each undeclared identifier is reported only once for each function it appears in
TFT/src/User/API/parseACK.c: At top level:
TFT/src/User/API/parseACK.c:1422:1: error: expected identifier or '(' before '}' token
 }
 ^
*** [.pio/build/BIGTREE_TFT43_V3_0/src/src/User/API/parseACK.o] Error 1
```

### Steps to reproduce

1. Try to build when using a Delta printer
2. Fails to build with the error messages above

**Expected behavior**

**Actual behavior**

### Hardware Variant

TFT43_V3

### Additional Information

https://pastebin.com/uYdfFTWi

digant73:
**IMPROVEMENTS:**
- Added TX Interrupt/DMA based serial writing based on excellent work and discussion shared by **@rondln** in #2835: TX serial tasks (e.g. transmission of gcodes to mainboard or ACK messages forwarding to ESP3D) is now decoupled from main thread and handled by dedicated HW (interrupt or DMA mode selectable at compile time). It allowed to drastically improve responsiveness of TFT maintaining a perfect load towards mainboard and even supplementary serial ports (e.g. ESP3D).
Attached pictures show a comparison with current implementation. The current implementation struggles to reach and maintain the target TX slots (only 1 pending gcode towards the mainboard instead of the 31 available) having a very low scan rate (number of times the main loopProcess() function is invoked per second) while the proposed implementation maintains the target TX slots (a constant 31 pending gcodes towards the mainboard) with a huge difference on scan rate
- The implemented TX module also supports a blocking transmission in case the message to be transmitted is longer than the TX cache size or there is not enough free space on the TX cache. Chunks of the message will be sent once space on the TX buffer becomes available.
By the way, TX cache size is set to 256 bytes for mainboard and supplementary serial ports. It is bigger more than enough to buffer any gcodes (max 100 bytes) or even a long ACK message from M43 output (even 200 bytes) avoiding any need of blocking transmission
- Improved RX serial reading: the TFT is now able to manage contiguous (not separated by IDLE on serial line) and very long ACK messages (e.g. M43 output) from mainboard without any loss as discussed in #2835. It allowed a big increase in reading responsiveness
- Added priorities to back end and front end tasks to speed up more critical tasks (e.g. printing) as suggested by **@rondln** in #2836
- Extended baudrates range for serial ports. Values range for param `serial_port` in `config.ini` has been extended providing other standard baudrates such as 230400 and 921600 (e.g. supported by ESP3D)
- Minor cleanup/bug fixes

**NOTES:**
- TX mode (DMA or Interrupt mode) is selectable at compile time at the beginning of `Serial.h`. By default, TX DMA mode is configured to be used. Both DMA and Interrupt modes provide similar performance
- TX and RX improvements are companion improvements of ADVANCED OK feature recently merged. Putting all together, they allow to improve print quality and to increase serial connection speed (for mainboard and supplementary devices) if not limited by EMI or by other limits of the connected devices. Basically the TFT is no more the main bottleneck in the chain

**BUGFIXES:**
- Fixed #2858

resolves #2835
fixes #2858

**PR STATE:** ready for merge

**Current solution:**

other_3

**Proposed solution:**

other_3


rondlh:
Well done, thanks!

I got the BTT TFT35V3.0.1(GD32F207 @ 120MHz), it's about 35% faster than the BTT TFTV3.0 (STM32F207 @ 120MHz).
The MKS TFT35V1.0 (STM32F407 @ 168MHz) is only about 20% faster than the BTT TFT35V3.0.
Tested with the scan rate benchmark.

rondlh:
I noticed that you already started to deal with the priorities in loopProcess and it's friends.

I think that a lot of tasks in `loopBackEnd` are actually frontend tasks, like . Almost everything under the "dividing line" seems to be front end:
1. loopCheckHeater, loopFan, loopSpeed are just for screen updates and requesting info updates.
2. loopBuzzer can be removed with the event-triggering approach: [buzzer.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/12722260/buzzer.zip)
3. The encoder needs a bit more attention, be careful with that one. Perhaps it should even be above the priority dividing line.
4. Mode switching, loopScreenShot, loopCheckBackPress, LCD_CheckDimming, LED_CheckEvent and rrfStatusQuery are also front end.

Usually front ends are given a **constant guaranteed amount of attention** by guaranteeing a number of updates per second.
Anything above twice the screen refresh rate feels response. I'm not sure what refresh rates the TFTs use, I assume 60Hz or lower. So 120 updates/s is responsive, but we can be generous and make it 3x the refresh rate (200/s, 5ms intervals).

digant73:
> I noticed that you already started to deal with the priorities in loopProcess and it's friends.
> 
> I think that a lot of tasks in `loopBackEnd` are actually frontend tasks, like . Almost everything under the "dividing line" seems to be front end:
> 
> 1. loopCheckHeater, loopFan, loopSpeed are just for screen updates and requesting info updates.
> 2. loopBuzzer can be removed with the event-triggering approach: [buzzer.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/12722260/buzzer.zip)
> 3. The encoder needs a bit more attention, be careful with that one. Perhaps it should even be above the priority dividing line.
> 4. Mode switching, loopScreenShot, loopCheckBackPress, LCD_CheckDimming, LED_CheckEvent and rrfStatusQuery are also front end.
> 
> Usually front ends are given a **constant guaranteed amount of attention** by guaranteeing a number of updates per second. Anything above twice the screen refresh rate feels response. I'm not sure what refresh rates the TFTs use, I assume 60Hz or lower. So 120 updates/s is responsive, but we can be generous and make it 3x the refresh rate (200/s, 5ms intervals).

Some features are split in a back end task and front end task. I agree with you to provide different priorities to some back end task. Currently I would avoid to make (rework) too much other things. I would simply integrate the improvements on buzzer.

rondlh:
[User.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/12742601/User.zip)

rondlh:
[User.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/12742657/User.zip)

rondlh:
I would recommend to make the changes of this PR as narrow as possible to reduce the risks. Possibly the writing mode should be user configurable so there is no risk of running into issues because there is a safe fallback. This would impact FW size a bit. I believe the smallest available flash size is 224KB, currently we are around 193KB, so we have some headroom for a useful feature like this.

---
Did you try the buzzer implementation? If so, don't forget to disable loopBuzzer in loopFrontEnd.
I checked if there are more processes that can use event triggering. There are quite a few, like notifications and such could be launched by the notification arriving. The problem is how to end the notification, for that loopProcess is still needed. This works well for the buzzer because it's interrupt based, so when the sound is done, you are still in control to make decisions. So it seems the easiest solution is just to give lower priority to these tasks.

---
My ESP3D 2.1.x is a bit modified (for the M43 response and general faster response times):
1. Increased RX buffer from 512 to 8KB: Config.h #define SERIAL_RX_BUFFER_SIZE 512 * 16 
2. syncwebserver.cpp reduced response time from 2000ms to 200ms.
```
//pickup the list
while ((millis() - timeout < 200) && !done) { // IRON, was 2000, delays response
```

digant73:
> I would recommend to make the changes of this PR as narrow as possible to reduce the risks. Possibly the writing mode should be user configurable so there is no risk of running into issues because there is a safe fallback. This would impact FW size a bit. I believe the smallest available flash size is 224KB, currently we are around 193KB, so we have some headroom for a useful feature like this.
> 
Do you mean to make it configurable at runtime (configured in `config.ini` and/or in `Feature` menu? Currently I make it configurable only at compile time.

> Did you try the buzzer implementation? If so, don't forget to disable loopBuzzer in loopFrontEnd. I checked if there are more processes that can use event triggering. There are quite a few, like notifications and such could be launched by the notification arriving. The problem is how to end the notification, for that loopProcess is still needed. This works well for the buzzer because it's interrupt based, so when the sound is done, you are still in control to make decisions. So it seems the easiest solution is just to give lower priority to these tasks.
> 
Not yet tried your buzzer implementation. I would possibly provide it in the PR related to other improvements.

> My ESP3D 2.1.x is a bit modified (for the M43 response and general faster response times):
> 
> 1. Increased RX buffer from 512 to 8KB: Config.h #define SERIAL_RX_BUFFER_SIZE 512 * 16
> 2. syncwebserver.cpp reduced response time from 2000ms to 200ms.
> 
> ```
> //pickup the list
> while ((millis() - timeout < 200) && !done) { // IRON, was 2000, delays response
> ```

My ESP3D is now perfectly working. It is able to receive the entire M43 output with no lost at all. This night I will make some testing reducing the RX cache size to 2048 instead of 4096.

rondlh:
> Do you mean to make it configurable at runtime (configured in `config.ini` and/or in `Feature` menu? Currently I make it configurable only at compile time. I would avoid to make it configurable at runtime (it means to check the mode at every transmission request).

Agree, that not very efficient, but it's a safe way to get started. And don't forget, it's only 1 check per message, not much overhead.

> > Did you try the buzzer implementation? If so, don't forget to disable loopBuzzer in loopFrontEnd. I checked if there are more processes that can use event triggering. There are quite a few, like notifications and such could be launched by the notification arriving. The problem is how to end the notification, for that loopProcess is still needed. This works well for the buzzer because it's interrupt based, so when the sound is done, you are still in control to make decisions. So it seems the easiest solution is just to give lower priority to these tasks.
> 
> Not yet tried your buzzer implementation. I would possibly provide it in the PR related to other improvements.

Please let me do it... This could be my first PR... our friend made a competing PR already based on my code, he's just a troll.
 
> > My ESP3D 2.1.x is a bit modified (for the M43 response and general faster response times):
> > 
> > 1. Increased RX buffer from 512 to 8KB: Config.h #define SERIAL_RX_BUFFER_SIZE 512 * 16
> > 2. syncwebserver.cpp reduced response time from 2000ms to 200ms.
> > 
> > ```
> > //pickup the list
> > while ((millis() - timeout < 200) && !done) { // IRON, was 2000, delays response
> > ```
> 
My ESP3D is now perfectly working. It is able to receive the entire M43 output (more than 14K) with no lost at all. This night I will make some testing reducing the RX cache size to 2048 instead of 4096.

Great, that's good to see. Working on ESP3D integration could be something nice to work on in the future. I get the time from the ESP3D, and give a estimated ETA when you start a print. And of course 921600 baud support.

digant73:
> > Do you mean to make it configurable at runtime (configured in `config.ini` and/or in `Feature` menu? Currently I make it configurable only at compile time. I would avoid to make it configurable at runtime (it means to check the mode at every transmission request).
> 
> Agree, that not very efficient, but it's a safe way to get started. And don't forget, it's only 1 check per message, not much overhead.
> 
> > > Did you try the buzzer implementation? If so, don't forget to disable loopBuzzer in loopFrontEnd. I checked if there are more processes that can use event triggering. There are quite a few, like notifications and such could be launched by the notification arriving. The problem is how to end the notification, for that loopProcess is still needed. This works well for the buzzer because it's interrupt based, so when the sound is done, you are still in control to make decisions. So it seems the easiest solution is just to give lower priority to these tasks.
> > 
> > 
> > Not yet tried your buzzer implementation. I would possibly provide it in the PR related to other improvements.
> 
> Please let me do it... This could be my first PR... our friend made a competing PR already based on my code, he's just a troll.
> 
sure

> > > My ESP3D 2.1.x is a bit modified (for the M43 response and general faster response times):
> > > 
> > > 1. Increased RX buffer from 512 to 8KB: Config.h #define SERIAL_RX_BUFFER_SIZE 512 * 16
> > > 2. syncwebserver.cpp reduced response time from 2000ms to 200ms.
> > > 
> > > ```
> > > //pickup the list
> > > while ((millis() - timeout < 200) && !done) { // IRON, was 2000, delays response
> > > ```
> 
> My ESP3D is now perfectly working. It is able to receive the entire M43 output (more than 14K) with no lost at all. This night I will make some testing reducing the RX cache size to 2048 instead of 4096.
> 
> Great, that's good to see. Working on ESP3D integration could be something nice to work on in the future. I get the time from the ESP3D, and give a estimated ETA when you start a print. And of course TFT 921600 baud support would be nice to have.

Yes, I will probably add also the support to 921600 baud (in the past I also asked the author of ESP3D to provide 1000000 baud)

rondlh:
Update, this one holds your corrects (4x) and an additional function.
[Serial.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/12755752/Serial.zip)

rondlh:
[stm32f10x.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/12767527/stm32f10x.zip)

rondlh:
[F1 Serial.zip](https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/files/12770398/F1.Serial.zip)

rondlh:
In loopBackEnd, it's probably safer to move the LCD_ENCODER_SUPPORT section above the priority divider.
You can try to use the encoder under high stress to confirm.

These are all frontEnd processes in my view:
```
  // Temperature monitor
  loopCheckHeater();

  // Fan speed monitor
  loopFan();

  // Speed & flow monitor
  loopSpeed();

  // Buzzer handling
  #ifdef BUZZER_PIN
    loopBuzzer();
  #endif
```

digant73:
> 1. In loopBackEnd, it's probably safer to move the LCD_ENCODER_SUPPORT section above the priority divider.
>    You can try to use the encoder under high stress to confirm.

I tested also under stress and it works

> 2. These are all frontEnd processes in my view:
> 
> ```
>   // Temperature monitor
>   loopCheckHeater();
> 
>   // Fan speed monitor
>   loopFan();
> 
>   // Speed & flow monitor
>   loopSpeed();
> 
>   // Buzzer handling
>   #ifdef BUZZER_PIN
>     loopBuzzer();
>   #endif
> ```
> 
> 3. I don't think there is a need for the function: `Serial_GetReadingIndexRX`, it's only used in 1 location in the code, but defined in 3 locations (for different hardware platforms).

No, but it is a function related to Serial.c/h as Serial_GetWritingIndexRX so I preferred to move in Serial.h.

> 4. In Serial_Get
> 
> ```
> rIndex = 0;
> while (rIndex <= flag)
> {
>   *(buf++) = cache[rIndex++];
> }
> ```
> 
> I believe this is the same as:
> 
> ```
> while (rIndex <= flag)
> {
>   *(buf++) = *(cache++);
> }
> ```
It doesn't work. I causes an infinite loop (rIndex is never updated)
> 
> 5. You mention the line below in `Serial_NewDataAvailable` and `Serial_Get`
> 
> `// NOTE: used 32 bit variables for performance reasons (in particular for data copy)`
> 
> But in Serial_**NewDataAvailable** there is no data copy, and in **Serial_Get** the data copy is done in 8-bit cycles.
> 
yes cut&paste of comment. I will change it

> In general I found that using and passing 32-bit variables is faster, that is probably because in the background 32-bit variables are always used, and `bools` and 8/16-bit variables are emulated, which takes extra steps. The official recommendation is to not blindly use the smallest datatype possible, but use 32-bit variables.

I made some testing using 32bit in functions but there is no difference.
> 
> 6. If the default is DMA mode, then perhaps the comment should be worded differently... "Comment out this line to..."
>    // uncomment this line to use TX DMA based serial writing. Otherwise, TX interrupt based serial writing will be used
>    #define TX_DMA_WRITE

Yes it is better. I will change it

rondlh:
All clear, one more thing...

The new baudrates do not seem to be in `Configuration.h` yet

digant73:
> All clear, one more thing...
> 
> The new baudrates do not seem to be in `Configuration.h` yet

yes, right thanks. updated also Configuration.h. Please, don't forget to test GD version

rondlh:
**Confirmed:** GD32F2 and STM32F4 **working fine**, both modes tested, well done!
Hardware: BTT TFT35V3.0.1 / MKS TFT35 V1.0

digant73:
> **Confirmed:** GD32F2 and STM32F4 **working fine**, both modes tested, well done! Hardware: BTT TFT35V3.0.1 / MKS TFT35 V1.0

thanks for the testing. PR ready for merge

ETE-Design:
My ESP3D is now perfectly working. It is able to receive the entire M43 output (more than 14K) with no lost at all. This night I will make some testing reducing the RX cache size to 2048 instead of 4096.

Great, that's good to see. Working on ESP3D integration could be something nice to work on in the future. I get the time from the ESP3D, and give a estimated ETA when you start a print. And of course TFT 921600 baud support would be nice to have.

@rondlh - I remember that MKS offered that feature for both their TFT and LCD so it was possible to to add Wireless Network direct from TFT / LCD, but only worked on their Hardware, cause some of their board offered ESP8266 as an addon for their Board and and also for their TFT
Would be a really great feature to add to this TFT also :-)

F14V:
Amazing improvement! I love things like this.

One note, it seems the config file in the "Copy to SD" directory was not updated. Is it supposed to be updated differently?

digant73:
@F14V Only BTT can update that folder. In the meanwhile you (everyone) have to compile the fw

F14V:
> @F14V Only BTT can update that folder. In the meanwhile you (everyone) have to compile the fw
Oh, good to know, sorry if it was stated somewhere and I missed it.

Indeed, I had to compile since the automatic build didn't run yet. But it's working great! (didn't see any pauses yet, they were occurring in the past)

digant73:
Yes, the automatic build populate that folder (config files, binary files). That task is executed by BTT only

PR #2903, opened 2024-03-02
Added new stale.yml

digant73:
**BUGFIXES:**
- Updated `stale.yml` according to the new supported `stale action` defined in https://github.com/actions/stale/tree/main?tab=readme-ov-file:
  Current `stale.yml` is deprecated and no more supported by github as reported in https://github.com/probot/stale. As result, both the inactive issues and PRs are currently no more marked as stale and deleted after the configured time (e.g. 60 and 7 days).
  The new stale action is scheduled to be run by crontab each day at 1:30 AM. Both the issues and PRs will be marked as stale after 60 days of inactivity and deleted after 7 days if not updated

**PR STATE:** Ready for merge


thisiskeithb:
The StaleBot fix & cleanup changes should really be split apart into separate PRs since they are unrelated.

digant73:
> The StaleBot fix & cleanup changes should really be split apart into separate PRs since they are unrelated.

it was just a minor cleanup. However, just moving the cleanup on a separate PR

PR #2918, opened 2024-03-16
✅ Update Actions versions

thisiskeithb:
### Description

- Update Actions to latest versions because *most* are using Node.js 20 now due to Node.js 16 being deprecated. See [GitHub Blog - GitHub Actions: Transitioning from Node 16 to Node 20](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) for more information.

- Update to Python 3.9

- `lock-threads` Action replaced `lock-threads-app` "Application", so `lock.yml` is no longer needed. It doesn't look like the LockBot ever ran here though, so this will be a new feature 😄 

- Limit `stale` Action to this repo (`bigtreetech/BIGTREETECH-TouchScreenFirmware`) like what is done for other Actions.


digant73:
do you know why stale is not applied also to PR? I mean, all the current PRs are older than 60 days but they were not yet marked as stale.

thisiskeithb:
> do you know why stale is not applied also to PR? I mean, all the current PRs are older than 60 days but they were not yet marked as stale.

@digant73: They're not stale. kisslorand continues to keep them updated. They need at least 60 days of zero activity to be marked as stale & then no comments/changes after another 7 days for them to be closed.

image

PR #2920, opened 2024-03-24
code reduction & cleanup

digant73:
Once the code is pretty stable since some time now, scope of this PR is an overall cleanup and code reduction (no code is changed).

**IMPROVEMENTS:**
- Code reduction: with the exception of 3rd party libs (e.g. ST, GD libraries etc.), `static` and `static inline` qualifiers has been applied for local-scope variables and functions (best practice) allowing a reduction of about 800 bytes on flash usage.
- Cleanup: with the exception of 3rd party libs, an overall cleanup has been applied. No code was changed, only some functions and declarations/definitions moved or ordered from .h to .c when possible.
- Some bugfixes on last integrated `GD32F30xx` API (used by recent MKS TFT28 variants):
  - UART ports set to 5: 6 UART ports have been previously defined but only 5 could be available/initialized (6th port `_USART6` was present but not initialized in `uart.c`).
  - RAM size set to 96KB: RAM size was previously set to 64KB in `gd32f30xxC_0x7000_iap.ld` file.

**PR STATE:** Ready for merge


rondlh:
Changed 281 files, amazing!!! Well done like always.

**Tiny issue:**
in User\Menu\ScreenSettings.c, line 19

```
static const char * const labelMarlinType[ITEM_MARLIN_TYPE_NUM] =
{
  // item value text(only for custom value)
  "128x64",
  "20x4"
};
```

This is only needed if the emulator is enabled, so probably best to guard it (#ifdef HAS_EMULATOR) or push it down to the HAS_EMULATOR section.

digant73:
yes, I will apply the filter on the next PR (I will complete the cleanup)

PR #2981, opened 2025-06-15
Fix display not showing Machine Type

kabroxiko:
### Description

As of https://github.com/MarlinFirmware/Marlin/commit/3fd175af8ea56db48ea41d7f3d55e967f0df8cd8, Display stop showing MACHINE_TYPE (printer name) on the main screen, in the Status section, and on the Information screen.
I tested the old & new M115 output with this patch and it seems to work correctly.

### Benefits

Back to normal display shows MACHINE_TYPE when ready

### Related Issues

Issue not reported yet

### DIFF
--- a/TFT/src/User/API/Mainboard_AckHandler.c
+++ b/TFT/src/User/API/Mainboard_AckHandler.c
@@ -1283,7 +1283,7 @@ void parseAck(void)
       {
         string = &ack_cache[ack_index];
         string_start = ack_index;
-        string_end = string_start;
+        string_end = ack_index - sizeof("MACHINE_TYPE:");;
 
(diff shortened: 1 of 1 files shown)


kabroxiko:
but it is broken again with https://github.com/bigtreetech/BIGTREETECH-TouchScreenFirmware/commit/343afa99406f35e97621469a982551bbe040b6db

kabroxiko:
I'm built a component for moonraker to connect a serial tft (https://github.com/moonraker-components/serial-tft-adapter). Normally M115 will return "FIRMWARE_NAME:Klipper {{ software_version }} SOURCE_CODE_URL:https://github.com/Klipper3d/klipper PROTOCOL_VERSION:1.0 MACHINE_TYPE:{{ printername }}....". I forked your project to show the SSID an the IP when the firmware is klipper. But that's another story.

Issue #2990, opened 2025-09-15
[BUG] Long press on title bar freezes the TFT


rondlh:
### Description

Long pressing the title bar freezes the TFT. Affects all hardware but perhaps in different ways. Confirmed on BTT TFT3.0 V3.0, BTT TFT35 V3.01 and MKS TFT35 V1.0.

### Steps to reproduce

1. Just long press (> 500ms) the title bar.

**Expected behavior**
Nothing should happen or perhaps the notifications menu is opened.

**Actual behavior**
The TFT freezes, it doesn't respond to any input anymore. A reset is required to recover from this state.

### Hardware Variant
BTT TFT35 V3.0, BTT TFT35 V3.01, MKS TFT35 V1.0

### TFT Firmware Version & Main Board Firmware details
Currently latest firmware, July 17 2025.

### Additional Information
I know the cause of the freeze bug and will post a Pull Request soon.
The cause is a hard fault of the MCU because of reading from an invalid address.
This might affect different hardware in different ways.

PR #2991, opened 2025-09-15
Long press title bar freeze bug fix

rondlh:
Long press on the title bar freezes the TFT.

### Requirements
None, affects all TFTs, but perhaps in different ways.

### Description

The core issue is a read from an invalid address when the title bar is long pressed because the key that is pressed (tempKey) is used for as index: ` getCurMenuItems()->items[tempKey].label.index`. The title bar is not a menu item which causes a read from a undefined address which typically causes hard fault of the MCU. I can confirm that this freezes the BTT TFT35 V3.0, BTT TFT35 V3.01 and MKS TFT35 V1.0.

### Benefits

TFT doesn't freeze anymore when the title bar is long pressed.

### Related Issues

#2990

### DIFF
--- a/TFT/src/User/API/menu.c
+++ b/TFT/src/User/API/menu.c
@@ -1229,7 +1229,7 @@ void loopCheckBackPress(void)
 
     TS_Sound = true;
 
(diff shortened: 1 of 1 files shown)


rondlh:
Code is final and ready for release.

digant73:
this PR could be merged

Issue #2992, opened 2025-09-15
[BUG] Code contains a 24 timer rollover bugs


rondlh:
### Description
Looking at the code I found it contains a large number of timer rollover causes by incorrect timer value addition/subtraction and comparison. 
Often the assumption is made that if you get a timestamp A and later compare it with the current time B, that B is bigger than A, this assumption is false under certain conditions because time 32 bit milliseconds timer will rollover from 0xFFFFFFFF to 0. By not just directly comparing the timestamps but checking the difference between the timestamps while using 32bit unsigned math, this issue can be avoided.

### Steps to reproduce

It's difficult to reproduce these bugs in normal use because it only happens after a long time. By adjusting the timer in the code I can do some test which shows that at rollover the TFT will behave erratically, but I never found to TFT to freeze. It depends on what the TFT is doing at the time.

**Expected behavior**
TFT behaves correctly even after a long time

**Actual behavior**
TFT can behave erratically after a long time

### Hardware Variant
All, this is not a hardware issue, this is a software issue.

### Additional Information
I have a PR ready to solve these bugs

PR #2993, opened 2025-09-15
24 rollover bugfixes using ELAPSE and PENDING macros

rondlh:
Solves 24 timer rollover bug using ELAPSED and PENDING macros.

### Requirements
None

### Description
This PR solves 24 timer rollover bugs I found in the code. These bugs can make the TFT behave erratically, but only after a long time. The cause is the 32 bit timer register rollover from 0xFFFFFFFF to 0. When using incorrect timer math the results can be incorrect.
To solve the bugs I use a strategy bored by Marlin, which uses macros to prevent issues. 

PENDING(when) with "when" a time in the future.
ELAPSED(starttime, interval) with starttime a time in the past, and the interval the time to check for if it has elapsed since the starttime.

### Benefits
1. No erratic TFT behavior after a long on time.
2. Set a good example for users who review the code base.
3. Improved code quality.

### Related Issues

#2992

### DIFF
--- a/TFT/src/User/my_misc.h
+++ b/TFT/src/User/my_misc.h
@@ -43,15 +43,15 @@ extern "C" {
 // flip all bits
 #define FLIP_BITS(num) ~num
 
+// timer rollover safe functions
+#define PENDING(when)                ((int32_t)((when) - OS_GetTimeMs()) > 0)
+#define ELAPSED(starttime, interval) !PENDING((starttime) + (interval))
+
 // time conversion
 #define SEC_TO_MS(t)  (t * 1000)           // seconds to milliseconds
 #define MS_TO_SEC(t)  (t / 1000)           // milliseconds to seconds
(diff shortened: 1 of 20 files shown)


TFT/src/User/my_misc.h
```diff
@@ -49,6 +49,10 @@ extern "C" {
 #define MIN_TO_SEC(t) (t * 60)             // minute to seconds
 #define MIN_TO_MS(t)  (SEC_TO_MS(t) * 60)  // minute to milliseconds
 
+// timer rollover safe functions
```
digant73: I would move this macros just above `// time conversion`

rondlh: > I would move this macros just above `// time conversion`

Done!

rondlh: Code is final and ready for release.

digant73:
this PR could be merged.
@rondlh I would only rename the macros to:
- PENDING_TIME
- ELAPSED_TIME

PR #2998, opened 2025-09-21
Fix notification title text visible behind toast message

rondlh:
Fix notification title text partly visible behind toast message. Title position was hard coded (x=10), use START_X variable instead.

### Description

The notification menu title text is still partly visible behind a toast message. This is because the title text position is hard coded at x=10, while other graphics items use a framework based on calculated variables (START_X, START_Y etc.).
The toast icon starts at START_X, which is 12 for a  480x320 display. So 2 pixels of the title text will still be visible behind the toast message. 

### Benefits

Fixes graphical corruption/misalignment, improves consistency of graphical layout independent of the TFT resolution, no hardcoded positions, but calculated position like other graphical items on the screen.

### Related Issues

None

### DIFF
--- a/TFT/src/User/API/menu.c
+++ b/TFT/src/User/API/menu.c
@@ -828,7 +828,7 @@ void menuDrawTitle(void)
 
   // draw title
   uint16_t start_y = (TITLE_END_Y - BYTE_HEIGHT) / 2;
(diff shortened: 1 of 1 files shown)


digant73:
this PR could be merged

Issue #3001, opened 2025-09-22
[BUG] USART6 pins are incorrect for MKS TFT35 V1.0


rondlh:
### Description
The defined USART6 pins are incorrect for MKS TFT35 V1.0.
They default to:
USART6_TX_PIN PG14
USART6_RX_PIN PG9
According to Hal\stem32f2_f4xx\uart.c

This pinout is not correct for the 100 pin package that is used for this TFT, perhaps it's not correct for any of the TFTs as they typically do not use packages with more than 100 pins, so they don't have Pins Gxx.

### Steps to reproduce

1. Connect to the USART6 port on the MKS TFT35 V1.0
2. Try to communicate

**Expected behavior**
Working serial communication

**Actual behavior**
Serial communication doesn't work

### Hardware Variant
MKS TFT35 V1.0

### TFT Firmware Version & Main Board Firmware details
July 17 2025 V0.28

### Additional Information
I will release a PR soon to fix this issue

PR #3002, opened 2025-09-22
Add USART6 pins to MKS TFT35 V1.0 variant so it doesn't fall back to the incorrect defaults

rondlh:
Add USART6 pins to MKS TFT35 V1.0 variant so it doesn't fall back to the incorrect defaults

### Requirements

MKS TFT35 V1.0 Touch screen

### Description

Add USART6 pins to MKS TFT35 V1.0 variant so it doesn't fall back to the incorrect defaults. The TFT uses the STM32F407VET6 MCU in a 100 pin package. (https://github.com/makerbase-mks/MKS-TFT-Hardware/blob/master/MKS%20TFT35/MKS%20TFT35%20V1.0_003/MKS%20TFT35%20V1.0_003%20SCH.pdf).
So the correct USART6 pins are:
USART6_TX_PIN PC6
USART6_RX_PIN PC7

Confirmed on the actual hardware.

### Benefits
Working USART6 serial communication

### Related Issues

#3001


digant73:
this PR could be merged

PR #3006, opened 2025-09-23
Updated icon files

NOVAXIM:
### Description

I improved the centering in one icon files and color in second icon file.

### Benefits

The image centering in icon is the same as in others, and the color provide a better perception.


rondlh (approved):
It's only a tiny update, and thus only a tiny improvement, but an improvement either way.
Currently the users will not be triggered to update the icons when they flash a new firmware. 
I think that's a good decision for this update. I have replaced my icons with these versions and I will keep it that way.

